PR #113, opened 2021-12-31
Optimize get_range() and add more benchmarks

cberner:


cberner:
@casey this should fix your slow query performance with multimap tables

PR #173, opened 2022-02-14
Improve error returned when database does not exist

### LINKED ISSUE #172: Consider removing `Error::DoesNotExist` in favor of `Error::Io`
It's possible to check if an `io::Error` was caused by a file being missing by checking if `error.kind() == io::ErrorKind:: NotFound`.\

I it might be preferable to remove `redb::Error::DoesNotExist`, and return an I/O error with kind `NotFound` if the database file doesn't exist. If I didn't look at the different variants of `redb::Error`, I might have missed `DoesNotExist`, and assumed that I should check the `Io` error variant to see if the database file was missing.

cberner:
Fixes #172

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -10,10 +10,11 @@ use memmap2::MmapRaw;
 use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
 use std::fs::{File, OpenOptions};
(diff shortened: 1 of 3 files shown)


casey:
Nice!

PR #178, opened 2022-02-19
Include page size in db stats

casey:
It strikes me know that this might actually not belong on DbStats, since it doesn't change during the life of the database, so it might be better as a method on `Database`. What do you think?

Calling `DbStats::new` with a bunch of usizes that are easy to mix up seemed error-prone, so I changed initialization to use the constructor directly, which is harder to mess up.


cberner:
I think putting it in the stats is fine. If it ever gets unwieldy, then ya splitting into two different methods could make sense

PR #179, opened 2022-02-19
Use pub(crate)

casey:
I've never seen `pub(in crate)` before, so I could definitely be wrong about this, but I think it's equivalent to `pub(crate)`.

This was done with `sed -i 's/pub(in crate)/pub(crate)/g' **/*.rs`.


cberner:
Thanks! I've been meaning to do that, but forgot about it. Ya, they're the same. I just didn't know you could drop the `in` at first.

Can you run rustfmt and force push?

casey:
Done!

cberner:
Merged. Thanks!

PR #193, opened 2022-02-25
Increase database size in resize_db test

casey:
This test crashes with an out-of-space error on my machine, so I bumped the database size, which fixes it.


cberner:
Closing, since this commit was in the other PR that I merged

PR #194, opened 2022-02-25
Rename transactions

casey:
I wanted to propose this name change, and it was easy enough that I thought I would just open a PR. The `Database` in `DatabaseTransaction` feel redundant, since it's already part of `redb`. I think in normal speech people usually say "read transaction" and "write transaction", so those seemed like good names.


casey:
Also, since `redb` is so short, if users want more explicit names, they can always use `redb::ReadTransaction` instead of importing it.

cberner:
:+1: I like this much better!

PR #195, opened 2022-02-25
Pass table definitions by value

casey:
This PR changes functions that take table definitions to take them by value, so you don't have to add a `&`, and implements `Copy` and `Clone` for table definitions. We have to implement copy and clone ourselves, because the derive macro can't figure out that it can copy and clone the value even when the type arguments aren't copy and clone, because we only use pass them to phantom values.


cberner:
Oh ya, good idea, thanks!

PR #199, opened 2022-02-27
Use result alias

casey:
This is a common pattern where a library will define a `Result` alias that defaults to `()` and the crates error type, so you don't have to write out `Error` everywhere.


cberner:
Nice! I didn't realize type parameters could have default values either, that's cool

PR #200, opened 2022-02-27
Use F_BARRIERFSYNC on macos

casey:
I'm not sure how you wanted to expose this to users, i.e. if it should be the default on macos because msync is awful, or by introducing a different commit type, but wanted to get a draft PR up. This code worked with the simple repo case, and I'll try it with the full ord database soon.

### DIFF
--- a/src/tree_store/page_store/mmap.rs
+++ b/src/tree_store/page_store/mmap.rs
@@ -1,23 +1,38 @@
 use crate::Error;
+use memmap2::MmapRaw;
+use std::fs::File;
 use std::ops::Range;
+use std::os::unix::io::AsRawFd;
 use std::slice;
 
 pub(crate) struct Mmap {
+    file: File,
     inner: memmap2::MmapRaw,
 }
 
 impl Mmap {
-    pub(crate) fn new(mmap: memmap2::MmapRaw) -> Self {
-        Self { inner: mmap }
+    pub(crate) fn new(file: File) -> Result<Self, Error> {
+        Ok(Self {
+            inner: MmapRaw::map_raw(&file)?,
+            file,
+        })
     }
 
     pub(crate) fn len(&self) -> usize {
         self.inner.len()
     }
 
+    #[cfg(target_os = "macos")]
+    pub(crate) fn flush(&self) -> Result<(), Error> {
+        let fd = self.file.as_raw_fd();
+        let code = unsafe { libc::fcntl(fd, libc::F_BARRIERFSYNC) };
+        assert_eq!(code, 0);
+        Ok(())
+    }
+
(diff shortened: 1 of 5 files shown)


cberner:
This looks fine for now. I'm not entirely sure this is safe, but random threads on stackoverflow suggest it's ok to mix mmap & fsync on modern OSs, and the fact that it takes 8ms on your laptop is about the right amount of time for flushing to disk. The correct fix is probably to switch all the writing to use `write()` and then use `fsync` on Linux too, but that's a way bigger change.

src/tree_store/page_store/mmap.rs
```diff
@@ -1,23 +1,38 @@
...
 
     pub(crate) fn len(&self) -> usize {
         self.inner.len()
     }
 
+    #[cfg(target_os = "macos")]
     pub(crate) fn flush(&self) -> Result<(), Error> {
-        self.inner.flush()?;
+        let fd = self.file.as_raw_fd();
+        let code = unsafe { libc::fcntl(fd, libc::F_BARRIERFSYNC) };
```
cberner, line 28: Can you add a `// TODO` noting that this may not be safe to mix with writes to mmap'ed memory, and to investigate switching to `write()`?

casey:
I added the todo, and proper error handling, although this adds a dependency on the errno crate. The code to get a human-readable string is actually [pretty annoying](https://github.com/lambda-fairy/rust-errno/blob/main/src/unix.rs#L26) but I can definitely remove the dependency if you'd prefer.

If we merge this as-is, it does have the problem that on mac data isn't guaranteed to be persisted after a transaction. I'm not relying on this, but I can imagine someone might at some point in the future.

src/tree_store/page_store/mmap.rs
```diff
@@ -1,23 +1,43 @@
...
     }
 
+    #[cfg(target_os = "macos")]
     pub(crate) fn flush(&self) -> Result {
-        self.inner.flush()?;
+        // TODO: It may be unsafe to mix F_BARRIERFSYNC with writes to the mmap.
+        //       Investigate switching to `write()`
+        let code = unsafe { libc::fcntl(self.file.as_raw_fd(), libc::F_BARRIERFSYNC) };
+        if code == -1 {
+            return Err(io::Error::new(io::ErrorKind::Other, errno()).into());
```
cberner, line 33: Would `Error::last_os_err()` do the same thing? I'd prefer to avoid the dependency on the `errno` crate, if possible

casey: Sweet that looks like it'll do. Fixed it.

cberner:
Merged, thanks! What a crazy bug to track down :)

PR #203, opened 2022-02-28
Add three levels of durability

casey:
The existing tests pass, but I didn't add any tests for the new durability level.

Why kinds of tests would be appropriate?

### DIFF
--- a/src/tree_store/page_store/mmap.rs
+++ b/src/tree_store/page_store/mmap.rs
@@ -26,20 +26,24 @@ impl Mmap {
         self.inner.len()
     }
 
-    #[cfg(target_os = "macos")]
     pub(crate) fn flush(&self) -> Result {
+        self.inner.flush()?;
+        Ok(())
+    }
+
+    #[cfg(not(target_os = "macos"))]
+    pub(crate) fn eventual_flush(&self) -> Result {
+        self.immediate_flush()
+    }
+
+    #[cfg(target_os = "macos")]
+    pub(crate) fn eventual_flush(&self) -> Result {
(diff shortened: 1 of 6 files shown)


casey:
I'm kind of sad that I removed `commit_helper_inner`, it would have been more fun to add a third layer, `commit_helper_inner_nested`.

src/tree_store/page_store/mmap.rs
```diff
@@ -26,8 +26,18 @@ impl Mmap {
         self.inner.len()
     }
 
-    #[cfg(target_os = "macos")]
     pub(crate) fn flush(&self) -> Result {
+        self.inner.flush()?;
```
cberner, line 30: On macos, this needs to call F_FULLFSYNC, doesn't it?

cberner (commented):
I can't think of a good way to add a test for this. The ideal test would be to do a bunch of eventually durable commits, and then unplug power from the hard disk. Not exactly easy to implement though =P

src/tree_store/page_store/mmap.rs
casey: Ah, right, fixed.

casey:
Why don't we duplicate a basic test that does a bunch of commits and change it to use eventually durable commits. It won't really test durability guarantees, but it will at least exercise the codepath, and it sounds like more testing infrastructure is needed to really exercise durability guarantees.

cberner:
Ya, that seems good for now. I'd suggest `integration_tests::persistence` that tests that data is persisted across closing & re-opening the database

casey:
Added a test for eventual persistence, this should be ready to go.

cberner:
Awesome, merged!

PR #206, opened 2022-03-01
Check formatting on CI

casey:
I also added a clippy run, but it had a bunch of scary complaints, so I gave up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,9 @@ jobs:
       - name: Run tests
         run: make build test
 
+      - name: Clippy
+        run: cargo clippy --all
+
       - name: Format
         run: cargo fmt --all -- --check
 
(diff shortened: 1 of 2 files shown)


casey:
I got clippy to run without complaining by not running it on all targets. Adding --all-targets makes it complain, but that's just tests and or benchmarks.

cberner:
I just merged a fix for all the Clippy errors. Want to rebase? Also, can you pin the rust version on the CI? I'd like to avoid it breaking unexpectedly when new versions are released

casey:
Just rebased and fixed some warnings.

Regarding pinning rust to the latest version, I've found that unless you're enforcing a MSRV, it's on balance better to use stable. New versions of rust have often substantial improvements to compilation speed, and it's nice to get those without having to remember to upgrade. Breakages are infrequent enough that they're not particularly annoying. Your call of course!

cberner:
That makes sense. Ok lets leave it on stable for now, and I'll pin it if it becomes annoying

casey:
Fixed the last of the warnings.

cberner:
Merged. Thanks!

PR #211, opened 2022-03-04
Remove memmap2 dependency

cberner:


src/tree_store/page_store/mmap.rs
```diff
@@ -58,9 +82,9 @@ impl Mmap {
     }
 
     pub(crate) fn get_memory(&self, range: Range<usize>) -> &[u8] {
```
casey, line 84: I wonder if `get_memory` should also be unsafe, since it can't be called with a range that overlaps with a current borrow from `get_memory_mut`.

cberner: Ah yes, that does seem better

PR #213, opened 2022-03-05
Add EditorConfig file

casey:
I have my editor configured to use two-space tabs by default, and this makes it do the right thing in the redb repo.


cberner:
Nice! I hadn't heard of this standard before. Cool that it's cross platform

PR #249, opened 2022-04-09
Optimize inserting into trees containing large values

### LINKED ISSUE #204: lmdb faster with large values
Ord is slowing down quite a bit after indexing around 150k blocks. This is different from the performance cliff we observed before, in that it still makes forward progress, it's just 5x slower than before. We're not sure this is due to redb, and we're planning to create a branch of ord that uses lmdb so we can do an apples to apples comparison, but in the meantime we benchmarked redb against lmdb with larger keys, and found that redb is 7x to 5x slower (7x on Liam's laptop, 5x on my laptop.)

I think this slowdown might not be captured in the existing benchmarks, so I wanted to open an issue for it. The code is in https://github.com/casey/redb-slowdown. Running on my machine:

redb: 5.57s user 15.25s system 69% cpu 29.899 total
lmdb: 0.42s user 3.50s system 65% cpu 5.950 total

I tried running the same benchmark with redb using non-durable commits, and with fsync instead of F_BARRIERFSYNC, which made it slightly faster, but didn't erase the gap completely.

cberner:
Fixes #204

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,10 @@ name = "lmdb_benchmark"
 harness = false
 
 [[bench]]
+name = "large_values_benchmark"
+harness = false
+
+[[bench]]
 name = "syscall_benchmark"
 harness = false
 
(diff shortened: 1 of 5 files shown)


casey:
Very nice! I'm surprised how small the diff is!

PR #351, opened 2022-09-09
Change write strategy names

### LINKED ISSUE #342: WriteStrategy name
I'm trying to figure out how to reproduce #337, and had the thought that changing the `WriteStrategy` name to be based on the algorithm might be a good idea. There are potentially a lot of different write strategies, and if they make different trade offs, `CommitLatency` and `Throughput` might wind up being not specific enough. Also, `CommitLatency` and `Throughput` are a little opaque.

Another thought is that `CommitStrategy` might be better than `WriteStrategy`, because it's used on commit, not on write. That is to say, all writes are done the same, it's the algorithm used to commit those writes that's different. And then 

I was thinking:

```
enum CommitStrategy {
    OnePhaseWithChecksum,
    TwoPhase,
}
```

What do you think? Happy to open a PR if you like the idea.

This is also partially based on the Rust standard library collections `std::collections::{BTreeMap, HashMap}`. I think these are better names than C++'s `map`, and `unordered_map`, since they tell you exactly how they're implemented, which lets you reason better about which one to use.

casey:
Closes #342. Let me know what you think of the comments. I wasn't exactly sure about the right terminology to use.

### DIFF
--- a/benches/large_values_benchmark.rs
+++ b/benches/large_values_benchmark.rs
@@ -123,7 +123,7 @@ fn main() {
         let tmpfile: NamedTempFile = NamedTempFile::new_in(current_dir().unwrap()).unwrap();
         let db = unsafe {
             redb::Database::builder()
-                .set_write_strategy(WriteStrategy::CommitLatency)
+                .set_write_strategy(WriteStrategy::Checksum)
                 .create(tmpfile.path(), 10 * 4096 * 1024 * 1024)
                 .unwrap()
         };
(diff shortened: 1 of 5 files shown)


casey:
Added some security notes. Honestly wanted to put something to the effect of "Other databases have the same security issues or worse, they just haven't thought about them enough or haven't documented them."

casey:
I also took the liberty to replace `use_checksums` with `WriteStrategy`, and implement `WriteStrategy::default()`. I think this improves readability a little bit, since someone browsing the codebase doesn't have to figure out where `use_checksums` is set, or which is the default `WriteStrategy`.

casey:
Also, noticed that rustfmt produced a lot of diffs. Not sure what's going on there.

cberner (approved):
Looks good to me! I'll figure out what's up with rustfmt. Maybe I have a global config set

PR #358, opened 2022-09-11
Make default region size consistent on 32bit & 64bit platforms

### LINKED ISSUE #332: MAX_USABLE_REGION_SPACE and target_pointer_width database portability
I forgot what the region size was, so I was looking at the MAX_USABLE_REGION_SPACE constant.

I noticed that it's different on 32 and 64 bit platforms, which made me wonder: What happens when a redb database is created in a 32 bit platform, and then opened on a 64 bit platform, and vice versa?

I think the ideal scenario might be for each redb database to record whether it's a 32 bit database or a 64 bit database. Then:

- 64 bit databases are created on 64 bit systems by default
- 32 bit databases are created on 32 bit systems by default
- 64 bit systems can open 32 and 64 bit databases
- 32 bit systems can open 32 bit databases, but get an error when trying to open 64 bit databases

Other options:

- Always create 32 bit databases, for portability, and require specifically requesting a 64 bit database
- Storing instead the max region size, instead of the bit target

Also, I may have entirely misunderstood the `MAX_USABLE_REGION_SPACE` constant, since it seems possible that regions might still be allocated as 4GiB on all systems, and only that keys are limited to 128MiB on 32 bit systems.

cberner:
Fixes #332

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -164,6 +164,7 @@ impl Database {
     ///
     /// The file referenced by `path` must not be concurrently modified by any other process
     pub unsafe fn create(path: impl AsRef<Path>, db_size: usize) -> Result<Database> {
+        let db_size = db_size as u64;
         let file = if path.as_ref().exists() && File::open(path.as_ref())?.metadata()?.len() > 0 {
             let existing_size = get_db_size(path.as_ref())?;
             if existing_size != db_size {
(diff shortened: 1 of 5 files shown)


casey:
Nice!

I didn't do a full review, but I noticed a few `x as usize`. This will silently truncate on 32 bit platforms where `x` is >= 2**32. I'm not sure if this is actually an issue here, but I like to avoid `as` in code like this, and use `.try_into()` and handle the error if it might fail, and `.try_into().unwrap()` if I know it will never fail. (Since the panic is better than silently truncating.)

cberner:
Ah yes, good idea! I filed a ticket to remove that from the whole code base: https://github.com/cberner/redb/issues/362

Issue #300, opened 2022-07-27
Windows support?


repi:
Would it be possible to add support for Windows for this crate also? 

It looks like currently the issue is that it uses raw unix file handles for the memory mapping with [`std::os::unix::io::AsRawFd`](https://crates.io/crates/memmap2), so may need to switch to a platform-independent version like `memmap2` crate for the memory mapping operations.

We've been interested in replacing our lmdb usage (and likely sled also) with redb and this is currently the main blocker.

cberner:
Ya, it'll probably take me a bit, since I'm unfamiliar with Windows APIs and am currently focused on stabilizing the file format.

I want to keep the dependencies of this crate to a minimum so am reluctant to add `memmap2`, but if you'd like to contribute a PR with Windows support that doesn't add a new dependency, I'd be happy to take it

xsxz01:
I also want to support Windows

PR #369, opened 2022-09-21
Add initial Windows support

Jake-Shadle:
1. Windows requires that file mappings to files on disk have a length > 0. This means that several tests immediately fail since they start with an empty file.
2. The Unix implementation truncates files to both decrease and increase the size of the mapped memory and then remaps and asserts that the base pointer is the same so that files can be resized while also being read from concurrently. While it's _possible_ that increasing the file size on Windows can be accomplished with [some "lightly" documented internal kernel functions](https://stackoverflow.com/a/55068177), _decreasing_ the size of the file is (AFAIK) impossible without unmapping the file first.

This provides initial support for Windows that passes _almost_ all tests, I've disabled `regression14 + 15` tests since they both allocate huge files and Windows since it doesn't support dynamic resize.

The only substantial change, other than just adding the mmap support for windows itself, was to change `TransactionalMemory::new` a bit so that it sizes the file appropriately _before_ opening the Mmap so that both Unix and Windows can be supported transparently. Since resizing is not supported the `dynamic_growth` toggle is always turned off on Windows with no way to enable it via the public API. I didn't go the more extreme route of removing the resize method from Mmap on Windows, since as I said above, it _could_ be possible to enable at least dynamic _increasing_ growth if wanted, but not be able to decrease it, but if that was wanted that could be moved to a separate PR as a follow up. Also allowing dynamic increasing growth would allow regression14 + 15 to be turned back on since they wouldn't hit disk limits.

There were also a few minor fixups to align the errors expected by tests when running on Windows.

Resolves: #300

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -188,7 +188,7 @@ impl Database {
             db_size,
             None,
             None,
-            true,
+            cfg!(unix),
             Some(WriteStrategy::default()),
         )
     }
@@ -202,7 +202,7 @@ impl Database {
         if File::open(path.as_ref())?.metadata()?.len() > 0 {
             let existing_size = get_db_size(path.as_ref())?;
             let file = OpenOptions::new().read(true).write(true).open(path)?;
-            Database::new(file, existing_size, None, None, true, None)
+            Database::new(file, existing_size, None, None, cfg!(unix), None)
         } else {
             Err(Error::Io(io::Error::from(ErrorKind::InvalidData)))
         }
@@ -503,7 +503,7 @@ impl Builder {
         Self {
             page_size: None,
             region_size: None,
-            dynamic_growth: true,
+            dynamic_growth: cfg!(unix),
             write_strategy: WriteStrategy::default(),
         }
     }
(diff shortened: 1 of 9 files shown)


cberner:
Cool, thanks for working on this!

1. Should be easy to fix. Take a look at `TransactionalMemory::new` I think if you just refractor it a bit you can pass the initial size to `Mmap` instead of needing to call resize()
2. This one is annoying, I think you can assert that `dynamic_growth` is false, and make that the default on Windows. That should disable the growing and shrinking logic

Jake-Shadle:
Ahh ok, I initially didn't want to touch anything outside of the mmap stuff itself to avoid making the PR bigger than it already is, I'll take a look at those suggestions tomorrow.

Jake-Shadle:
I'm not sure what the mmap leak test is for, but at least on the M1 I was running tests on it took ~20ms to do the mmap lock + construction which mean the 100000 iterations took a while... I also saw a test failure in the `reload_mmap` test, but I don't think that was because of this PR.

cberner (approved):
Awesome, thanks!

cberner:
The mmap leak test is to verify that it calls munmap. I had a bug at one point where it didn't, and then the fuzzer found that it ran out of address space after creating a few thousand Database objects and closing them :)

PR #370, opened 2022-09-22
Prevent parallel writes

casey:
It looks like the `live_write_transaction` mutex is being locked, asserted to be `None`, then dropped and re-acquired. I think this introduces a race condition where two transactions could grab the mutex, assert it's `None`, and then proceed.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -434,12 +434,13 @@ impl Database {
     /// Returns a [`WriteTransaction`] which may be used to read/write to the database. Only a single
     /// write may be in progress at a time
     pub fn begin_write(&self) -> Result<WriteTransaction> {
-        assert!(self.live_write_transaction.lock().unwrap().is_none());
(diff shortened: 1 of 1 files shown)


casey:
It looks like CI is failing, probably because of a new lint on latest stable.

Screen Shot 2022-09-22 at 11 41 05 AM

cberner:
Oops ya, nice find! Can you rebase on master so the tests work? I have branch protection on, so the tests have to pass before merging. 

> It looks like CI is failing, probably because of a new lint on latest stable.

This is why I had it pinned to a specific version =)

cberner (approved):

casey:
Done!

> This is why I had it pinned to a specific version =)

It's a little annoying when things break, but they usually break infrequently and an easily fixable way. I think it's very slightly more annoying to have to remember to update to the latest version. Plus when you forget to update, you're usually using a slower compiler.

PR #375, opened 2022-09-22
Add write lock

casey:
I replaced the live write transaction `Mutex` with an `RwLock` so that writes can be initiated in parallel on different threads, but will block each other. I'm not sure if this is right, since there are a lot of subtleties with things like this.

Also, I was thinking that `WriteTransaction::new` should be safe, and should only take a reference to the database, and should grab the `RwLock` and allocate a new transaction ID, instead of spreading the safety concerns between `Database::begin_write` and `WriteTransaction::new`.

Depends on #370.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -434,12 +434,13 @@ impl Database {
     /// Returns a [`WriteTransaction`] which may be used to read/write to the database. Only a single
     /// write may be in progress at a time
     pub fn begin_write(&self) -> Result<WriteTransaction> {
-        assert!(self.live_write_transaction.lock().unwrap().is_none());
+        let mut guard = self.live_write_transaction.lock().unwrap();
+        assert!(guard.is_none());
         let id = self.next_transaction_id.fetch_add(1, Ordering::AcqRel);
-        *self.live_write_transaction.lock().unwrap() = Some(id);
-        // Safety: We just asserted there was no previous write in progress
+        *guard = Some(id);
         #[cfg(feature = "logging")]
         info!("Beginning write transaction id={}", id);
+        // Safety: We just asserted there was no previous write in progress
         unsafe { WriteTransaction::new(self, id) }
     }
 
--- a/src/db.rs
+++ b/src/db.rs
@@ -15,7 +15,6 @@ use std::ops::RangeFull;
 use std::path::Path;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Mutex;
-use std::sync::RwLock;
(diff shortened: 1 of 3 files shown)


src/db.rs
```diff
@@ -149,7 +150,7 @@ pub struct Database {
     mem: TransactionalMemory,
     next_transaction_id: AtomicTransactionId,
     live_read_transactions: Mutex<BTreeSet<TransactionId>>,
-    live_write_transaction: Mutex<Option<TransactionId>>,
+    live_write_transaction: RwLock<Option<TransactionId>>,
```
cberner, line 153: I don't think this needs to change to an `RwLock`. You never call `.read()` and `.lock()` blocks just like `.write()`. Am I missing something?

src/db.rs
```diff
@@ -434,13 +429,14 @@ impl Database {
     /// Returns a [`WriteTransaction`] which may be used to read/write to the database. Only a single
     /// write may be in progress at a time
     pub fn begin_write(&self) -> Result<WriteTransaction> {
```
cberner, line 431: Can you update the docs to say that it will block if a write is in progress?

cberner (changes requested):
This seems like a good change, but I had a few comments. Also, can you add a test that tests the blocking with two threads?

src/db.rs
casey: You're totally right. I think maybe I had a prior revision which required a RwLock for some reason.

src/db.rs
casey: Done.

casey:
> This seems like a good change, but I had a few comments. Also, can you add a test that tests the blocking with two threads?

Comments addressed. Also, check out the "Make `begin_write` safe" commit. I think it's a slight improvement. At the expense of making two db fields `pub(crate)`, `WriteTransaction::new` is now safe, and the safety concerns are all in that function, instead of being spread across two functions.

casey:
Except for the tests! Where should I put them?

casey:
Added the test at the end of `basic_tests.rs`, let me know if there's a better place for it. This test consistently panics on master, but passes on the PR branch.

cberner:
Yep, basic_tests or integration_tests is where I put most things. I should sort them better at some point :)

cberner:
Looks like there's a compilation error. Also can you rebase to resolve the conflicts with your other PR?

casey:
Rebased and resolved.

cberner (approved):

PR #385, opened 2022-09-27
Use madvise(RANDOM)

cberner:
This improves performance by ~2x for databases that do not fit in memory

### DIFF
--- a/src/tree_store/page_store/mmap/unix.rs
+++ b/src/tree_store/page_store/mmap/unix.rs
@@ -49,10 +49,18 @@ impl MmapInner {
         if mmap == libc::MAP_FAILED {
             Err(io::Error::last_os_error().into())
         } else {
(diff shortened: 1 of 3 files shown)


casey:
👀

cberner:
@casey do you have a good benchmark for ord? This helps my synthetic benchmark that tries to be like ord's workload, but I ran ord with it and the transaction times are so high variance that I'm not sure whether it helps or not

casey:
Currently not. We want to write a good benchmark, but we haven't gotten around to it. It's a bit tricky, because range fragmentation happens slowly over time, and we've been vaguely worried that if we don't have the right pattern of transactions, we won't create a representative benchmark. We haven't been doing a ton of optimization recently, but that will probably change when we make a push to get users to run ord instances, and we'll definitely write a benchmark then.

PR #394, opened 2022-10-10
Implement Iterator for RangeIter

### LINKED ISSUE #285: Iterate over all the elements in table. How?
Hi,

Thank you for making `redb`. I am really enjoying working with it, but I hit a wall when wanting to list all the elements in a table.
So, is there a way to list all the elements in a table? I know about [`range()`](https://docs.rs/redb/latest/redb/struct.ReadOnlyTable.html#method.range) method but it doesn't seem to implement `IntoIterator`.

cberner:
Fixes #285

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -265,9 +265,9 @@ impl Database {
         }
 
         // Iterate over all other tables
-        let mut iter: BtreeRangeIter<str, InternalTableDefinition> =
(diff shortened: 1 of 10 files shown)


casey:
Very nice!

Clippy was very helpful in suggesting I start using using the new iterator implementation:

Screen Shot 2022-10-25 at 10 23 14 AM

PR #410, opened 2022-10-24
clear up bounds in `src`

notJoon:
## Description
clear up bounds in `src`.

I thought it would be better to organize it to increase readability. So I organized it. If possible, I'll clean up the other codes like `src/tree_store` and `page_store`

I'm not sure if it's ok. Please tell me if I did anything wrong. I am ready to edit at any time. And also, thanks for your time.

## TODO

- [x] clear tree_store
- [x] clear page_store

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -70,13 +70,21 @@ impl AtomicSavepointId {
 /// Defines the name and types of a table
 ///
 /// A [`TableDefinition`] should be opened for use by calling [`ReadTransaction::open_table`] or [`WriteTransaction::open_table`]
-pub struct TableDefinition<'a, K: RedbKey + ?Sized, V: RedbValue + ?Sized> {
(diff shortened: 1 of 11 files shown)


cberner (changes requested):
I prefer the inline trait bounds. Too many years of Java programming, and so it looks more natural to me than the `where` form :)

I'm happy to take the documentation improvements though, if you want to just submit those

notJoon:
Thanks. I'll try to participate from time to time.

cberner:
Sorry if my last comment wasn't clear, but I meant that I like the current form of type parameters. i.e. `pub struct TableDefinition<'a, K: RedbKey + ?Sized, V: RedbValue + ?Sized>` and not the `where` form that you changed it to.

notJoon:
Oh, I didn't understand it well. Anyway, thank you for your feedback.

PR #434, opened 2022-11-17
Convert makefile to justfile

casey:
I converted the makefile to a justfile, which is much cleaner, should have much better error messages on syntax errors, supports passing arguments to recipes, etc. Don't feel bad closing this if you're not interested, since this wasn't much work.

The main downside is that you have to install just, but that can be done with `cargo install just`.


casey:
Just noticed that the makefile is used on CI. This is pretty easy to fix with https://github.com/extractions/setup-just.

cberner:
Awesome, thanks! just syntax is much nicer :)

Issue #282, opened 2022-06-11
Benchmarks should not test using the temporary directory


ecton:
I recently learned about `tmpfs` when doing my own benchmarks for a database I'm working on. On Linux, `/tmp` may actually be mounted as a special filesystem that is implemented by using virtual memory for all written files. This means that all sync operations will essentially be instant, since there is no actual persistence being performed.

This leads to much different performance characteristics. For example, here's the large_values_benchmark on my machine before and after changing the location of the benchmarks:

```
+------------------------+--------+--------+---------+
|                        | redb   | lmdb   | sled    |
+====================================================+
| bulk load (2MB values) | 4605ms | 5376ms | 33013ms |
+------------------------+--------+--------+---------+

+------------------------+---------+---------+----------+
|                        | redb    | lmdb    | sled     |
+=======================================================+
| bulk load (2MB values) | 15607ms | 29809ms | 132410ms |
+------------------------+---------+---------+----------+
```

It's quite a large performance difference, and while I didn't try updating and testing the remaining benchmarks, it seems like it highlights redb's impressive performance even more. Great work!

cberner:
Thanks for pointing this out! Yes, it seems like Fedora and some other distros mount /tmp as a tmpfs by default. I'll move the temp files to the current directory

cberner:
@ecton off-topic, but since you have experience with database I'd love your feedback on #309 , if you have time! I'm trying to stabilize the redb file format in preparation for a 1.0 release.

PR #435, opened 2022-11-17
Ignore benchmark temporary files

casey:


cberner:
I think the thing to do here is fix the benchmark so it doesn't leak files. I guess it's not possible in every case, but a ctrl-C signal handler would solve most of them I think. How did you end up with leaked files?

casey:
How about using `tempfile::tempfile()` and `tempfile::tempdir()` to create tempfiles in `$TMPDIR`, which wouldn't show up in git and would get cleaned up eventually when the system clears the tmpdir?

cberner:
It used to do that, but I changed it because on some distros /tmp is a `tmpfs`: https://github.com/cberner/redb/issues/282

Maybe it could detect that it was a tmpfs and require the user to configure a different directory via an environment variable

casey:
In that case, how about using `./tmp`, adding it to gitignore, and not bothering to clean it up? That seems easier than having to add a ctrl-c handler to all tests that create temporary files.

cberner:
I moved it into .benchmark and also added the ctrl-c handler. Turned out to be pretty easy :)

PR #440, opened 2022-11-20
Refactor RedbValue usage to allow more flexible lifetimes

### LINKED ISSUE #431: Tuple type lifetime issue
I think there's an issue with tuple type lifetimes.

Check out the code in #430. I added a new test which tries to insert a `&str` borrowed from a `String` into a table:

```rust
#[test]
fn tuple_type_lifetime() {
    let tmpfile: NamedTempFile = NamedTempFile::new().unwrap();
    let db = unsafe { Database::create(tmpfile.path(), 1024 * 1024).unwrap() };

    let table_def: TableDefinition<(&str, u8), (u16, u32)> = TableDefinition::new("table");

    let write_txn = db.begin_write().unwrap();
    {
        let mut table = write_txn.open_table(table_def).unwrap();
        table
            .insert(&(String::from("hello").as_str(), 5), &(0, 123))
            .unwrap();
    }
    write_txn.commit().unwrap();

    let read_txn = db.begin_read().unwrap();
    let table = read_txn.open_table(table_def).unwrap();
    assert_eq!(table.get(&("hello", 5)).unwrap().unwrap(), (0, 123));
}
```

It fails with:

```
error[E0716]: temporary value dropped while borrowed
   --> tests/basic_tests.rs:132:23
    |
132 |             .insert(&(String::from("hello").as_str(), 5), &(0, 123))
    |                       ^^^^^^^^^^^^^^^^^^^^^ creates a temporary which is freed while still in use
133 |             .unwrap();
    |                      - temporary value is freed at the end of this statement
134 |     }
    |     - borrow might be used here, when `table` is dropped and runs the `Drop` code for type `Table`
    |
    = note: consider using a `let` binding to create a longer lived value
```

It looks like the `&str` is getting some lifetime that isn't compatible with the lifetime of the temporary, and thus can't take a str borrowed from the temporary.

cberner:
Fixes #431 
Fixes #430

### DIFF
--- a/benches/common.rs
+++ b/benches/common.rs
@@ -8,7 +8,7 @@ use std::fs::File;
 use std::path::Path;
 
 #[allow(dead_code)]
-const X: TableDefinition<[u8], [u8]> = TableDefinition::new("x");
+const X: TableDefinition<&[u8], &[u8]> = TableDefinition::new("x");
 
 pub trait BenchDatabase {
     type W<'db>: BenchWriteTransaction
@@ -117,7 +117,7 @@ impl<'db> BenchReadTransaction for RedbBenchReadTransaction<'db> {
 }
 
 pub struct RedbBenchReader<'txn> {
-    table: redb::ReadOnlyTable<'txn, [u8], [u8]>,
+    table: redb::ReadOnlyTable<'txn, &'static [u8], &'static [u8]>,
 }
 
 impl<'txn> BenchReader for RedbBenchReader<'txn> {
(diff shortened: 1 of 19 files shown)


cberner:
@casey can you give this branch a try? I think it fixes the lifetime issue you found. It has one annoying side effect though which is that `Table<&[u8], &[u8]>` now only accepts `&[u8]` as an argument to `insert()` and `get()` and types like `&[u8; 5]` and `&Vec<u8>` no longer work. Any ideas for making that work?

PR #442, opened 2022-11-22
Don't zero out freed pages

### LINKED ISSUE #386: Slow to open large database
We recently restarted ord, and the database was pretty slow to open, on the order of a minute or two. I'm not sure it was the culprit, but I think this was the first time we used a `WriteStrategy::Checksum` database on mainnet. Under what circumstances will `redb` check all the checksums for a database using `WriteStrategy::Checksum`? Is it every time the database opens, or only after a crash?

casey:
I'm not sure if there's more to it than this, but it seemed simple so I thought I'd open a PR. Fixes #386.


cberner:
Looks right to me!

PR #443, opened 2022-11-22
Always use 4k pages

### LINKED ISSUE #427: Base page size bounds
- [x] benchmark different base page sizes on SSD & HDD to see if this is something that users should really be allowed to configure
- [x] place sane limits on the range of base page size that is allowed

cberner:
Fixes #427


casey:
Nice, this is sweet. Fewer degrees of freedom FTW!

PR #444, opened 2022-11-22
Add iter method for iterating over all values in table

casey:
This adds `ReadableTable::iter` and `ReadableMultimapTable::iter` that both just call `self.range(..)`. I was hoping to implement `IntoIter` for both, so that they could be used in for loops, but couldn't because `range` can fail. An `iter` function seems relatively common, c.f. `Vec::iter` and `BTreeMap::iter`. I do `range(..)` a bunch, so this is nice to have. Also, `range(..)` can't infer the type, so you always have to use `range(MINIMUM_VALUE..)` to get it to work.


cberner:
Merged. Thanks!

PR #454, opened 2022-12-03
Update README.md to reflect no size limit on Databases

loren:
Size was removed in https://github.com/cberner/redb/commit/0fadb8fa1acb3d00cf407d0df81794767d404e24


cberner:
Merged. Thanks!

PR #481, opened 2023-01-04
Document enum Error.

darnuria:
Hello, wanted to help a little and found that `Error` could be documented a little bit! :)


cberner:
Merged. Thanks!

darnuria:
You're welcome will try to do some and maybe some easy good bug. Had some usage experience with Berkeley Database in rust at work and got feed up. :D

PR #482, opened 2023-01-05
Document Builder and TableDefinition.

darnuria:
Hello did some documentation on `Builder` + `TableDefinition`.

Document the defaults, looked a good idea to me, it's one of the "good" part of Berkeley db that I found useful to have default explicitly documented in dev-doc. :)

If it's wasn't necessary or need changes tell me! also i am not the best at spelling. :(

Just a suggestion, I may be wrong: `write_strategy` and `page_size` may be defined without their option and `write_strategy` could be directly set to `Checksum`. Maybe I missed some `public` code path in API that allow other choices, still toying around for now with the database. :)

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -475,6 +475,7 @@ pub enum WriteStrategy {
     TwoPhase,
 }
 
+/// Configuration builder of a redb [Database].
 pub struct Builder {
     page_size: Option<usize>,
     region_size: Option<usize>,
@@ -485,6 +486,13 @@ pub struct Builder {
 }
 
 impl Builder {
+    /// Construct a new [Builder] with sensible defaults.
+    ///
+    /// ## Defaults
+    ///
+    /// - `read_cache_size_bytes`: 1Gib = 1024 * 1024 * 1024 bytes
+    /// - `write_cache_size_bytes`: 100MiB Kib = 100 * 1024 * 1024 bytes
+    /// - `write_strategy`: [WriteStrategy::Checksum]
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


darnuria:
Whops just forgot to add some last minute fix in PR thanks CI. :)

src/db.rs
```diff
@@ -485,6 +492,13 @@ pub struct Builder {
 }
 
 impl Builder {
+    /// Construct a new [Builder] with sensible defaults.
+    ///
+    /// ## Defaults
+    ///
+    /// - `page_size`: 102400 Kib = 100 * 1024 * 1024 bytes.
+    /// - `write_cache_size_bytes`: 1048576 Kib = 1024 * 1024 * 1024 bytes
```
cberner, line 500: I think you meant `read_cache_size`, that's the one that defaults to 1GiB

src/db.rs
```diff
@@ -485,6 +492,13 @@ pub struct Builder {
 }
 
 impl Builder {
+    /// Construct a new [Builder] with sensible defaults.
+    ///
+    /// ## Defaults
+    ///
+    /// - `page_size`: 102400 Kib = 100 * 1024 * 1024 bytes.
```
cberner, line 499: I think you meant write_cache_size that's what defaults to 100MiB, not the page size

cberner (changes requested):

src/db.rs
darnuria: Oh that's right I just confused both while rewriting something and mixed it.

Fixed it and took your suggestion to write it as Mio/Gib.

cberner:
Thanks! I included most of these in #485

PR #484, opened 2023-01-06
Fix compilation of code in readme.md.

darnuria:
Since 57e4d20607e34933f1a80ac06eab9ba61f515af6 create is safe.
and fixed compilation error. :)


darnuria:
Feel free to double check the code I just pasted it in a new cargo project and make it compile with using 
```
[dependencies]
redb = "0.11.0"
```

darnuria:
Thanks for review and time, have a good day! :tada:

PR #490, opened 2023-01-15
Option support and alignment future proofing

### LINKED ISSUE #486: Option as RedbValue
I was thinking about combing two tables which map keys to u64s. However, the second table doesn't have an entry for each ID, so the value type would need to be `(u64, Option<u64>)`, but `Option<u64>` doesn't have an impl for `RedbValue`.

This isn't at all pressing, since I can just leave them separate, and I was combining them based on the assumption that having fewer tables with multiple values in each table would be faster, and I haven't actually benchmarked that. However I assume it'll be a common use-case.

cberner:
Fixes #486


casey:
Dope!

PR #508, opened 2023-01-26
WASI support

### LINKED ISSUE #507: WASI support?
Hi! I was looking to see if redb could compile to `wasm32-wasi` and came across a handful of issues (with mmap and file locking):
```
unresolved import `crate::tree_store::page_store::file_lock::LockedFile`
failed to resolve: use of undeclared type `MmapInner`
cannot find type `MmapInner` in this scope
```

I was wondering if you would be willing to support WASI as a compilation target in redb. I might be able to PR to fix the above compilation issues and test the following functionality:
1. mmap (which it looks like you may have grabbed from the unmaintained `memmap` crate?) can probably be fixed by just copying the unix implementation (WASI has a libc that should be the same as unix)
2. file locking may be exposable from unix like mmap; but I am not sure. Some file locking APIs are not available in WASI libc

GregoryConrad:
Fixes #507

### DIFF
--- a/src/tree_store/page_store/file_lock/mod.rs
+++ b/src/tree_store/page_store/file_lock/mod.rs
@@ -1,6 +1,6 @@
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "wasi"))]
 mod unix;
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "wasi"))]
 pub(super) use unix::LockedFile;
 
 #[cfg(windows)]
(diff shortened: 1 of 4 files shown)


GregoryConrad:
Based on https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/mman/mman.c, it looks as though redb in WASI would only work in read-only, if at all. Closing this PR as I don't think it makes any sense at the moment.

PR #510, opened 2023-01-28
Android compatible

morn-0:
Because libc lacks the Android platform api, it cannot be used on Android. Modify the compilation condition to be compatible with Android platform.


cberner:
Merged. Thanks!

PR #516, opened 2023-02-05
Fix crash when accessing the last 66 pages of a region

cberner:
Region headers are 66 pages long, and regions contain 2^32 bytes of data. The extra 66 header pages cause the address_range calculation to overflow a u32


cberner:
This fixes a data corruption issue introduced by https://github.com/cberner/redb/pull/476

PR #536, opened 2023-03-24
Remove mmap backend

### LINKED ISSUE #458: Audit and document all unsafe usage

cberner:
Fixes #458 

I decided that the maintenance burden of supporting the mmap backend is too high for too small of a performance improvement. This will make redb provably safe and trivial to port to other platforms. Proving that the mmap backend was safe seemed infeasible, and this greatly simplifies the codebase.


Kerollmops:
Hey @cberner 👋

I was wondering if the removal of the mmap backend means that the values are always allocated when retrieved? Which means that keeping too many values for too long can result in OOM?

cberner:
No, it should not. There's a cache which holds pages (redb's Page abstraction, not OS pages) and they're evicted when the cache is full. The size is configurable

Kerollmops:
Thank you very much for the answer, @cberner. I have a couple of other questions:
- What if I keep the [`AccessGuard`](https://docs.rs/redb/latest/redb/struct.AccessGuard.html) for a long time?
- Will the pages be kept for as long as I keep the guard, and therefore, it can blow up the memory?
- Or will the page be retrieved lazily when I call `AccessGuard::value`?
- And if it's the latter, I could keep the `&[u8]` it returns and keep **them** for a long time. What happens?

I am just thinking about this because I use LMDB and [extensively use memory mapping behavior in arroy](https://blog.kerollmops.com/multithreading-and-memory-mapping-refining-ann-performance-with-arroy) by keeping pointers to the vectors to process and read in parallel from the database.

cberner:
Yes, the pages will be kept for the lifetime of the `AccessGuard`. If the values are small, you could copy them out of the `AccessGuard` and retain them for a long-time. Alternately, you can keep the key around, and retrieve the value from the `Table` on-demand, which will read from the cache if the value is cached and read from disk if it has been evicted

Kerollmops:
> Alternately, you can keep the key around, and retrieve the value from the `Table` on-demand, which will read from the cache if the value is cached and read from disk if it has been evicted

Thank you, @cberner; it seems to be the best alternative. My last question is: Is there a way to read a non-committed transaction in parallel? Like reading the current transaction with multiple read transactions so that I can spawn multiple threads with each transaction and read the table in each (making sure no writes are performed while these read transactions live).

```rust
table.write(wtxn, "abc", "long embedding")?;
table.write(wtxn, "def", "another long embedding")?;

let rtxn_gen = wtxn.lock_for_read()?;
let rtxn0 = rtxn_gen.new_read_txn()?;
let rtxn1 = rtxn_gen.new_read_txn()?;

// not allowed while rtxn_gen lives
// table.write(wtxn, "def", "another long embedding")?;

assert_eq!(table.read(rtxn0, "abc")?, "long embedding");
assert_eq!(table.read(rtxn1, "def")?, "another long embedding");

rtxn0.abort();
rtxn1.abort();
drop(rtxn_gen);

table.write(wtxn, "ghi", "another super long embedding")?;
```

cberner:
Read transactions can only read committed data, so you can't directly do that. The two options would be:
1) `Table`s implement `Send` so you can perform multithreaded reads of the `Table` during the write transaction. For example, see this https://github.com/cberner/redb/pull/848
2) commit  (you can use a `Durability::None` commit, if you're concerned about latency) and then read

PR #546, opened 2023-03-31
Raise test line coverage in src/table from 88% to 98%.

darnuria:
Hello, it's not much but runned grcov and found that some lines were not covered. Had some time for a little contribution so I tried it out.

<https://github.com/mozilla/grcov#usage>.

In the meantime I had to disable compilation (commented in cargo.toml) of benchmarking, if not Grcov went mad. maybe using a workspace and keeping bench behind a feature gate is worth I don't know.


cberner:
Merged. Thanks!

Issue #548, opened 2023-04-01
don't depend on `pyo3-build-config`


ozgrakkurt:
Hey!

Currently it is pulled in by default, it seems like it can be disabled

cberner:
I couldn't figure out how to fix the fuzzer compilation issues with disabling it. Feel free to send a PR if you figure it out though.

PR #549, opened 2023-04-01
make pyo3-build optional

ozgrakkurt:
closes https://github.com/cberner/redb/issues/548


cberner:
@ozgrakkurt seems like there are some CI failures. Can you take a look?

PR #552, opened 2023-04-08
Make pyo3-build-config dep optional

cberner:
Fixes https://github.com/cberner/redb/pull/549
Fixes https://github.com/cberner/redb/issues/548


cberner:
No idea why this doesn't work.

PR #577, opened 2023-05-11
Use GitHub Flavored Markdown table

briankung:
This converts the benchmark table into one that will render appropriately on GitHub. Also made a graph using Canva's Graph Maker, but it is a bit over the top and not automatable in case the benchmarks change, so I just included it here.

redb benchmarks


briankung:
Oh, and here's what the table will look like on GitHub

Screenshot 2023-05-10 at 9 20 53 PM

cberner:
Merged. Thanks!

PR #583, opened 2023-05-21
Add WASI support

### LINKED ISSUE #507: WASI support?

GregoryConrad:
Adds experimental WASI support. Requires nightly and file locking is currently a no-op, at least for the immediate future (needs to get implemented upstream).

Partially fixes #507

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,12 +22,12 @@ libc = "0.2.104"
 log = {version = "0.4.17", optional = true }
 pyo3 = {version = "0.18.0", features=["extension-module", "abi3-py37"], optional = true }
 
-# These are actually test dependencies
+# Common test/bench dependencies
 [dev-dependencies]
 rand = "0.8"
 tempfile = "3.5.0"
 
-# These are actually benchmarking dependencies
+# Just benchmarking dependencies
 [target.'cfg(not(target_os = "wasi"))'.dev-dependencies]
 ctrlc = "3.2.3"
 fastrand = "1.9.0"
@@ -47,8 +47,6 @@ python = ["pyo3"]
 logging = ["log"]
 # Enable cache hit metrics
 cache_metrics = []
-# Enable experimental wasi support (requires nightly and has no file locking!)
-wasi = []
 
 [profile.bench]
 debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,17 +22,19 @@ libc = "0.2.104"
 log = {version = "0.4.17", optional = true }
 pyo3 = {version = "0.18.0", features=["extension-module", "abi3-py37"], optional = true }
 
-[dev-dependencies]
+[target.'cfg(not(target_os = "wasi"))'.dev-dependencies]
 ctrlc = "3.2.3"
 fastrand = "1.9.0"
-rand = "0.8"
 lmdb-rkv = "0.14.0"
 sanakirja = "1.3.3"
-tempfile = "3.5.0"
 sled = "0.34.7"
 rocksdb = "0.21.0"
 libc = "0.2.99"
 comfy-table = "6.1.0"
+
+[dev-dependencies]
+rand = "0.8"
+tempfile = "3.5.0"
 
 [target.'cfg(target_os = "linux")'.dev-dependencies]
 io-uring = "0.5.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ python = ["pyo3"]
 logging = ["log"]
 # Enable cache hit metrics
 cache_metrics = []
-# Whether to enable experimental wasi support (requires nightly and has no file locking!)
+# Enable experimental wasi support (requires nightly and has no file locking!)
 wasi = []
 
 [profile.bench]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,8 @@ python = ["pyo3"]
 logging = ["log"]
 # Enable cache hit metrics
 cache_metrics = []
+# Whether to enable experimental wasi support (requires nightly and has no file locking!)
+wasi = []
 
 [profile.bench]
 debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,12 @@ libc = "0.2.104"
 log = {version = "0.4.17", optional = true }
 pyo3 = {version = "0.18.0", features=["extension-module", "abi3-py37"], optional = true }
 
+# These are actually test dependencies
+[dev-dependencies]
+rand = "0.8"
+tempfile = "3.5.0"
+
+# These are actually benchmarking dependencies
 [target.'cfg(not(target_os = "wasi"))'.dev-dependencies]
 ctrlc = "3.2.3"
 fastrand = "1.9.0"
@@ -31,10 +37,6 @@ sled = "0.34.7"
 rocksdb = "0.21.0"
 libc = "0.2.99"
 comfy-table = "6.1.0"
-
-[dev-dependencies]
-rand = "0.8"
-tempfile = "3.5.0"
 
 [target.'cfg(target_os = "linux")'.dev-dependencies]
 io-uring = "0.5.1"
--- a/src/db.rs
+++ b/src/db.rs
@@ -710,11 +710,13 @@ impl std::fmt::Debug for Database {
 
 #[cfg(test)]
 mod test {
+    use tempfile::NamedTempFile;
+
     use crate::{Database, Durability, Error, ReadableTable, TableDefinition};
 
     #[test]
     fn small_pages() {
-        let tmpfile = crate::create_tempfile();
+        let tmpfile: NamedTempFile = NamedTempFile::new().unwrap();
 
         let db = Database::builder()
             .set_page_size(512)
@@ -731,7 +733,7 @@ mod test {
 
     #[test]
     fn small_pages2() {
-        let tmpfile = crate::create_tempfile();
+        let tmpfile: NamedTempFile = NamedTempFile::new().unwrap();
 
         let db = Database::builder()
             .set_page_size(512)
@@ -825,7 +827,7 @@ mod test {
 
     #[test]
     fn small_pages3() {
-        let tmpfile = crate::create_tempfile();
+        let tmpfile: NamedTempFile = NamedTempFile::new().unwrap();
 
         let db = Database::builder()
             .set_page_size(1024)
@@ -858,7 +860,7 @@ mod test {
 
     #[test]
     fn small_pages4() {
-        let tmpfile = crate::create_tempfile();
+        let tmpfile: NamedTempFile = NamedTempFile::new().unwrap();
 
         let db = Database::builder()
             .set_cache_size(1024 * 1024)
@@ -895,7 +897,7 @@ mod test {
 
     #[test]
     fn crash_regression1() {
-        let tmpfile = crate::create_tempfile();
+        let tmpfile: NamedTempFile = NamedTempFile::new().unwrap();
 
         let db = Database::builder()
             .set_cache_size(1024 * 1024)
@@ -918,7 +920,7 @@ mod test {
 
     #[test]
     fn dynamic_shrink() {
-        let tmpfile = crate::create_tempfile();
+        let tmpfile: NamedTempFile = NamedTempFile::new().unwrap();
         let table_definition: TableDefinition<u64, &[u8]> = TableDefinition::new("x");
         let big_value = vec![0u8; 1024];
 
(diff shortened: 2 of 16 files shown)


src/db.rs
```diff
@@ -650,7 +650,7 @@ impl Builder {
     /// Set the amount of memory (in bytes) used for caching data
     pub fn set_cache_size(&mut self, bytes: usize) -> &mut Self {
         // TODO: allow dynamic expansion of the read/write cache
-        self.read_cache_size_bytes = bytes * 9 / 10;
+        self.read_cache_size_bytes = bytes / 10 * 9;
```
GregoryConrad, line 653: This fixes a memory overflow with wasm since it is 32 bit

GregoryConrad:
@cberner This is currently blocked by:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Unsupported, message: "operation not supported on this platform" }', src/tree_store/page_store/page_manager.rs:451:30
```

Which is:
```rust
        let mut storage = PagedCachedFile::new(
            file.try_clone().unwrap(), // panics on the unwrap()
            page_size as u64,
            read_cache_size_bytes,
            write_cache_size_bytes,
        )?;
```

Any ideas on what can be done to bypass this `try_clone()`, as it is not supported on WASI?

cberner:
I'm pretty sure it can just be removed: https://github.com/cberner/redb/pull/584

I don't remember why I added that, but it looks like I refactored the code at some point and it's no longer necessary

src/db.rs
cberner: seems fine to me

GregoryConrad:
`wasmer` was being a bit problematic with the file permissions for some reason, so I switched to `wasmtime`, and it works!

```
cargo +nightly build --target wasm32-wasi --features=wasi && wasmtime run target/wasm32-wasi/debug/main.wasm --dir .
```

TODO:
- [ ] Scan through other `cfg`s in the repo to see if wasi needs to be stitched into places that currently only support unix
- [ ] Figure out how to run the test suite with wasi (`cargo-wasi` looks like it could be helpful at first glance)
- [ ] Add WASI testing/checks into CI

Cargo.toml
```diff
@@ -22,18 +22,20 @@ libc = "0.2.104"
 log = {version = "0.4.17", optional = true }
 pyo3 = {version = "0.18.0", features=["extension-module", "abi3-py37"], optional = true }
 
-[dev-dependencies]
+[target.'cfg(not(target_os = "wasi"))'.dev-dependencies]
```
GregoryConrad, line 25: I *really* hate this, but I don't think there is any other way to do benchmark specific dependencies (even with features). If you have any ideas though, I am all ears.

(This divides test specific deps and benchmark specific deps, since essentially none of the benchmark deps compile to WASI)

src/tree_store/page_store/file_lock/unix.rs
```diff
@@ -1,14 +1,33 @@
+// TODO once Rust's libc has flock implemented for WASI, this file needs to be revisited.
+// What needs to be changed is commented below.
+// See also: https://github.com/WebAssembly/wasi-filesystem/issues/2
+
+// Remove this line once wasi-libc has flock
+#![cfg_attr(feature = "wasi", allow(unused_imports))]
```
GregoryConrad, line 6: This silences some unused import warnings that will go away once wasi-libc has flock

GregoryConrad:
Currently running the wasi tests with:

```bash
cargo +nightly wasi test --features wasi -- --nocapture
```

[Not sure how I'll be able to allow fs access with the above approach since `--dir xyz` doesn't work](https://bytecodealliance.github.io/cargo-wasi/testing.html), but I'll cross that bridge when I get there.

For now, I'm getting 
```
test db::test::crash_regression1 ... thread 'main' panicked at 'no filesystem on wasm', library/std/src/sys/wasi/os.rs:228:5
```

Guessing this is because of `tempfile`. Might see if I can patch in some dummy implementation that will work for WASI (i.e., hardcodes a `tmp/` on WASI or something).

GregoryConrad:
I've started to get tests passing! 🥳🎉

To get useful test output, I've needed `cargo-wasi`; here's the command for that:
```bash
CARGO_TARGET_WASM32_WASI_RUNNER="wasmtime --dir=." cargo +nightly wasi test --features=wasi -- --nocapture
```
To run tests without any extra dependencies, you can use this (but note this is near impossible to debug in case a test fails):
```bash
CARGO_TARGET_WASM32_WASI_RUNNER="wasmtime --dir=." cargo +nightly test --target wasm32-wasi --features=wasi
```

So I recommend we go with the `cargo-wasi` option for CI.

src/lib.rs
```diff
@@ -36,3 +43,12 @@ mod transactions;
 mod tree_store;
 mod tuple_types;
 mod types;
+
+#[cfg(test)]
```
GregoryConrad, line 47: @cberner Can I get your opinion on this (and 3891da5 in general)?

`tempfile` used as it was causes all tests to fail in WASI. This little workaround is needed in order to prevent that.

Is this an appropriate place for this function, or should it be placed elsewhere (or should I use a different solution entirely)? Wasn't 100% sure what your preference was, but wanted to get your feedback before I went all in and converted all tests to use this.

GregoryConrad: I think I am instead just going to use `NamedTempFile::new_in("")` everywhere instead of `NamedTempFile::new()`. A bit easier that way, and tests still pass on other platforms. Only downside is CWD is used instead of whatever the system temp dir is, but I don't really think that matters too much for tests

tests/integration_tests.rs
```diff
@@ -122,9 +122,10 @@ fn immediate_persistence() {
     test_persistence(Durability::Immediate);
 }
 
+#[cfg(not(feature = "wasi"))] // TODO figure out why this fails on WASI
 #[test]
 fn free() {
```
GregoryConrad, line 127: This is the only test that is failing that shouldn't be failing. Looked thru and not entirely sure what is going on in this test so might need some help to look at this:
```
test free ... thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `275`', tests/integration_tests.rs:189:5
```
Line in question:
```rust
    assert_eq!(allocated_pages, txn.stats().unwrap().allocated_pages());
```
Wondering if this is something to do with page size or the fact that WASM is 32 bit? Not entirely sure though.

GregoryConrad: For reference, expected and actual are both `2` on my macOS computer. What could cause the actual value to be 275 in WASI?

GregoryConrad:
@cberner Left some comments above that I would appreciate some feedback on. Other than that, this PR is ready to go as far as I can tell

tests/multithreading_tests.rs
```diff
@@ -1,64 +1,66 @@
-use redb::{Database, ReadableTable, TableDefinition};
-use std::sync::Arc;
-use std::thread;
-use tempfile::NamedTempFile;
+#[cfg(not(target_os = "wasi"))]
+mod multithreading_test {
```
GregoryConrad, line 2: This file is the same as before. Just wrapped it in a `mod` to avoid compiling/running any of it on WASI

src/lib.rs
cberner: I'd rather use the helper function. The problem with cwd is that if the tests crash (or get ctrl-C) I think the temp files could end up leaking and polluting the working directory.

tests/integration_tests.rs
cberner: I bet this is a usize = 32bits bug. I've had a couple of those. I think the problem is line 162. Try changing that to `(i as u64).to_le_bytes()` so that it matches line 177

cberner: yep, 99% sure that's the issue. Try with this patch: https://github.com/cberner/redb/pull/588

cberner:
Overall this is looking much simpler than I had expected it to be!

src/lib.rs
GregoryConrad: Yea, I thought about & noticed that problem too. I might need to duplicate the helper function in a few places though (since tests/ isn’t apart of the crate), which is why I originally removed it after I added it.

A bit new to rust still, so I’ll just go ahead and duplicate the function in those several places it’s needed. If you know of a way to avoid the duplication though, I’m all ears

tests/integration_tests.rs
GregoryConrad: Thanks, was wondering if it had to do with it being 32 bit. I’ll try it out tomorrow

GregoryConrad: It's passing locally now, thanks!

GregoryConrad:
In addition to the tempfile test changes I'll make at some point later today, I wanted to mention some warnings you get while compiling with nightly:

```
warning: variable does not need to be mutable
   --> src/tree_store/page_store/page_manager.rs:824:13
    |
824 |         let mut secondary = state.header.secondary_slot_mut();
    |             ----^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: variable does not need to be mutable
   --> src/tree_store/page_store/page_manager.rs:880:13
    |
880 |         let mut secondary = state.header.secondary_slot_mut();
    |             ----^^^^^^^^^
    |             |
    |             help: remove this `mut`

   Compiling librocksdb-sys v0.11.0+8.1.1
warning: using `.borrow()` on a double reference, which returns `&<K as RedbValue>::SelfType<'_>` instead of borrowing the inner type
  --> src/table.rs:49:62
   |
49 |             .map(|x| x.map(|(key, _)| K::as_bytes(key.value().borrow()).as_ref().to_vec()));
   |                                                              ^^^^^^^^^
   |
   = note: `#[warn(suspicious_double_ref_op)]` on by default

warning: using `.borrow()` on a double reference, which returns `&<K as RedbValue>::SelfType<'_>` instead of borrowing the inner type
  --> src/table.rs:68:62
   |
68 |             .map(|x| x.map(|(key, _)| K::as_bytes(key.value().borrow()).as_ref().to_vec()));
   |                                                              ^^^^^^^^^

warning: using `.borrow()` on a double reference, which returns `&<K as RedbValue>::SelfType<'_>` instead of borrowing the inner type
   --> src/tree_store/btree.rs:270:55
    |
270 |             assert!(operation.safe_delete(entry?.key().borrow())?.is_some());
    |                                                       ^^^^^^^^^

warning: using `.borrow()` on a double reference, which returns `&<K as RedbValue>::SelfType<'_>` instead of borrowing the inner type
   --> src/tree_store/btree.rs:307:58
    |
307 |                 assert!(operation.safe_delete(entry.key().borrow())?.is_some());
    |                                                          ^^^^^^^^^
```

I'd be happy to fix these in a new PR since I imagine they will trickle down to stable eventually.

GregoryConrad:
@cberner Instead of adding that `create_tempfile` back in, how would you feel about tests using a `tmp/` folder within the project root that can be created & cleaned up via just? Even with the `create_tempfile` function, WASI would still pollute the project root whenever tests are run (and would probably need this `tmp/` anyways because of that). Further, with a local `tmp/`, you can also inspect failed tests' temp files in the future more easily.

cberner:
I don't think that's a good idea, because I want to support development workflows that do not use just. I do something similar in the benchmarks, where they create and try to cleanup their own tmpdir and it's kind of a mess. The only reason I do it there is that some OSs put /tmp in a tmpfs which messes up the benchmark timing.

What about reading the tmpdir from an environment variable? Would it be possible to create a directory in /tmp and then mount it via `wasmtime --dir=` in the justfile?

GregoryConrad:
Should be all set now. Non-wasi platforms will act the same way they did before, and `just test_wasi` will use the system's `$TMPDIR`.

GregoryConrad:
After this gets merged I can make 2 dummy commits to show the diffs that will be needed:
- Once wasi-libc gets `flock` (and after that change is propagated to Rust's libc)
- Once WASI no longer requires nightly

Just so those changes are less of a guessing game when the time comes. I'll link the diffs in #507

cberner (approved):

cberner:
Merged. Thanks!

PR #659, opened 2023-08-18
fix(insert sig lifetime): key and value should have different lifetimes

zjp-CN:
also add `generic_signature_lifetimes` test for it

---

For the latest version, the `write_key_generic` function won't compile:

```rust
error[E0597]: `buf` does not live long enough
  --> src/main.rs:27:31
   |
20 |         key: K::SelfType<'_>,
   |         --- has type `<K as RedbValue>::SelfType<'1>`
...
23 |         let buf = [1, 2, 3];
   |             --- binding `buf` declared here
...
27 |             table.insert(key, buf.as_slice()).unwrap();
   |             ------------------^^^^^^^^^^^^^^-
   |             |                 |
   |             |                 borrowed value does not live long enough
   |             argument requires that `buf` is borrowed for `'1`
...
30 |     }
   |     - `buf` dropped here while still borrowed
```

### DIFF
--- a/src/multimap_table.rs
+++ b/src/multimap_table.rs
@@ -585,15 +585,11 @@ impl<'db, 'txn, K: RedbKey + 'static, V: RedbKey + 'static> MultimapTable<'db, '
     /// Add the given value to the mapping of the key
     ///
     /// Returns `true` if the key-value pair was present
-    pub fn insert<'a>(
+    pub fn insert<'k, 'v>(
         &mut self,
-        key: impl Borrow<K::SelfType<'a>>,
-        value: impl Borrow<V::SelfType<'a>>,
-    ) -> Result<bool>
-    where
-        K: 'a,
-        V: 'a,
-    {
+        key: impl Borrow<K::SelfType<'k>>,
+        value: impl Borrow<V::SelfType<'v>>,
+    ) -> Result<bool> {
         let value_bytes = V::as_bytes(value.borrow());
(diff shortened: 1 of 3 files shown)


src/multimap_table.rs
```diff
@@ -585,15 +585,11 @@ impl<'db, 'txn, K: RedbKey + 'static, V: RedbKey + 'static> MultimapTable<'db, '
...
-        key: impl Borrow<K::SelfType<'a>>,
-        value: impl Borrow<V::SelfType<'a>>,
-    ) -> Result<bool>
-    where
-        K: 'a,
-        V: 'a,
-    {
+        key: impl Borrow<K::SelfType<'k>>,
+        value: impl Borrow<V::SelfType<'v>>,
+    ) -> Result<bool> {
```
cberner, line 592: Can you add the `where` clause back? I'm a little unclear on whether it's required for the GAT, but I think it's recommended

cberner (changes requested):
Looks good to me, except one minor thing

src/multimap_table.rs
zjp-CN: Adding it back is trivial. But it's redundant because there is `K: RedbKey + 'static, V: RedbKey + 'static` lifetime bound and `K: 'any, V: 'any` always holds.

cberner: Ah, yes I see. Thanks for pointing that out. Will merge it as is!

cberner (approved):

PR #665, opened 2023-08-27
Allow opening a database using a file handle instead of a path

adamreichold:
This enables using this library with e.g. [`cap-std`](https://docs.rs/cap-std) which avoids path-based access in favour of a capability-based approach.

I am not sure about the name, but `open` while probably most fitting is already taken for the path-based approach. Alternatives I considered were `from_file`, `access`, `mount` and `open_or_init`.

I added another commit to ensure that the build is Clippy-clean before submitting the PR. Please let me if know if you like me to drop that from the PR as it is admittedly unrelated. (Bumping `fastrand` resolves a few warnings about unnecessary mutability in the LMDB benchmark.)

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -12,7 +12,6 @@ use crate::{
 use crate::{ReadTransaction, Result, WriteTransaction};
 use std::fmt::{Display, Formatter};
 use std::fs::{File, OpenOptions};
-use std::io;
 use std::io::ErrorKind;
 use std::marker::PhantomData;
 use std::ops::RangeFull;
@@ -769,20 +768,19 @@ impl Builder {
 
     /// Opens an existing redb database.
     pub fn open(&self, path: impl AsRef<Path>) -> Result<Database, DatabaseError> {
-        if !path.as_ref().exists() {
-            Err(StorageError::Io(ErrorKind::NotFound.into()).into())
-        } else if File::open(path.as_ref())?.metadata()?.len() > 0 {
-            let file = OpenOptions::new().read(true).write(true).open(path)?;
-            Database::new(
-                file,
-                self.page_size,
-                None,
-                self.read_cache_size_bytes,
-                self.write_cache_size_bytes,
-            )
-        } else {
-            Err(StorageError::Io(io::Error::from(ErrorKind::InvalidData)).into())
+        let file = OpenOptions::new().read(true).write(true).open(path)?;
+
+        if file.metadata()?.len() == 0 {
+            return Err(StorageError::Io(ErrorKind::InvalidData.into()).into());
         }
+
+        Database::new(
+            file,
+            self.page_size,
+            None,
+            self.read_cache_size_bytes,
+            self.write_cache_size_bytes,
+        )
     }
 
     /// Open an existing or create a new database in the given `file`.
--- a/src/db.rs
+++ b/src/db.rs
@@ -808,7 +808,10 @@ impl std::fmt::Debug for Database {
 
 #[cfg(test)]
 mod test {
-    use crate::{Database, Durability, ReadableTable, TableDefinition};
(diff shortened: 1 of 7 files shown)


adamreichold:
Having read the implementation of `Builder::open`, I wonder if the following diff

```diff
diff --git a/src/db.rs b/src/db.rs
index 586bd98..57f10cb 100644
--- a/src/db.rs
+++ b/src/db.rs
@@ -12,7 +12,6 @@ use crate::{
 use crate::{ReadTransaction, Result, WriteTransaction};
 use std::fmt::{Display, Formatter};
 use std::fs::{File, OpenOptions};
-use std::io;
 use std::io::ErrorKind;
 use std::marker::PhantomData;
 use std::ops::RangeFull;
@@ -769,20 +768,19 @@ impl Builder {
 
     /// Opens an existing redb database.
     pub fn open(&self, path: impl AsRef<Path>) -> Result<Database, DatabaseError> {
-        if !path.as_ref().exists() {
-            Err(StorageError::Io(ErrorKind::NotFound.into()).into())
-        } else if File::open(path.as_ref())?.metadata()?.len() > 0 {
-            let file = OpenOptions::new().read(true).write(true).open(path)?;
-            Database::new(
-                file,
-                self.page_size,
-                None,
-                self.read_cache_size_bytes,
-                self.write_cache_size_bytes,
-            )
-        } else {
-            Err(StorageError::Io(io::Error::from(ErrorKind::InvalidData)).into())
+        let file = OpenOptions::new().read(true).write(true).open(path)?;
+
+        if file.metadata()?.len() == 0 {
+            return Err(StorageError::Io(ErrorKind::InvalidData.into()).into());
         }
+
+        Database::new(
+            file,
+            self.page_size,
+            None,
+            self.read_cache_size_bytes,
+            self.write_cache_size_bytes,
+        )
     }
 
     /// Open an existing or create a new database in the given `file`.
```

could be used to avoid the extra system calls to `stat` and `open`?

cberner:
ya, that diff seems like it would work. As long as there's a test for a missing and an empty file it should be safe to change it

src/db.rs
```diff
@@ -784,6 +784,19 @@ impl Builder {
             Err(StorageError::Io(io::Error::from(ErrorKind::InvalidData)).into())
         }
     }
+
+    /// Open an existing or create a new database in the given `file`.
+    ///
+    /// The file must be empty or contain a valid database.
+    pub fn build(&self, file: File) -> Result<Database, DatabaseError> {
```
cberner, line 791: I think I like `open_file` better. Maybe if I ever do a 2.0 release, I'll change `open` to `open_path`.

Also, I think this needs the same empty file test that `open` has

cberner (changes requested):

cberner:
Couple minor comments. Otherwise this seems good to me!

src/db.rs
adamreichold: > Also, I think this needs the same empty file test that open has

Since the implementation is based on that of `create`, I was under the impression that a newly created empty file would be initialized with a new database?

cberner: Ah right, I see. In that case lets name it `create_file` -- open to other suggestions on the name, but I think indicating that it's the same functionality as `create` is good.

adamreichold:
> ya, that diff seems like it would work. As long as there's a test for a missing and an empty file it should be safe to change it

I added test cases for these two error paths as a separate commit before the refactoring.

src/db.rs
adamreichold: Changed it to `create_file` and added a test case for the creation of new database in an empty (temporary) file.

cberner (approved):

cberner:
Merged. Thanks!

Issue #666, opened 2023-08-29
Struct serialization using bincode


ChHecker:
One of the most famous features of Rust is the trivial (de-)serialization of structs using `serde`. `bincode` can then use these structs in order to (de-)serialize structs to bytes.
Therefore, it should probably be possible to use this fact to implement `RedbValue` for a much bigger range of structs.

ChHecker:
So I tried working on this and a generic
```rust
impl<'a, T: serde::Deserialize + serde::Serialize> RedbValue for T {
     ...
}
```
conflicts with the implementation for `&[u8]` and probably `&str`. I guess a macro is the only solution as long as there is no specialization and negative trait bounds in Rust.

PR #667, opened 2023-08-29
Add serialization using serde and bincode as well as macros

ChHecker:
As described in #666, I added an automatic serialization using `serde` and `bincode`.

The simple option is the macro `be_value_serialize!` after activating the feature `serialize`. This, however, does not feature the same level of comfort as the derive macro in the feature `serialize-derive`. Derive macros require a second crate, though, so I create it. You would probably have to add it to crates.io yourself.


cberner:
Types should be fixed width if all values of that type can be serialized to a slice of the same width.

More broadly though, I'll need to think a bit about whether allowing people to derive this is a good idea. A couple concerns I have: changing the name of a struct that uses this macro will now be a breaking change, due to the `stringify` use. And adding or removing a field will be also

dpc:
So `width` should return `None`? For a convenience functionality like this I think it's the safest route. A type_name should be provided by an argument to macro? What are these type names for anyway?

I don't think redb should be treating bincode in any special way. `serialize` and `deserialize` from bincode should not be re-exported as some default, I think.

Cargo.toml
```diff
@@ -18,9 +18,15 @@ crate-type = ["cdylib", "rlib"]
 pyo3-build-config = "0.19.0"
 
 [dependencies]
+bincode = { version = "1.3.3", optional = true }
 libc = "0.2.104"
-log = {version = "0.4.17", optional = true }
-pyo3 = {version = "0.19.0", features=["extension-module", "abi3-py37"], optional = true }
+log = { version = "0.4.17", optional = true }
+pyo3 = { version = "0.19.0", features = [
```
dpc, line 24: ?

cberner:
> So `width` should return `None`? For a convenience functionality like this I think it's the safest route. A type_name should be provided by an argument to macro? What are these type names for anyway?
> 
> I don't think redb should be treating bincode in any special way. `serialize` and `deserialize` from bincode should not be re-exported as some default, I think.

The type names are for type safety. If you try to open a table using a different key or value type that that with which it was created, it will fail

dpc:
I see. Hmmm...

What if ... I lie? Change the type without changing the name?

Does it technically need to be `unsafe` as it must uphold a contract or otherwise risk undefined behavior?

E.g store a `u8` key with a value `0`, then change the type `to `NonZeroU8` yet keep the name the same?

ChHecker:
> So `width` should return `None`? For a convenience functionality like this I think it's the safest route. A type_name should be provided by an argument to macro? What are these type names for anyway?

`width` is currently set to `None` unless a macro argument is added. The type name by argument is a good idea. I'm new to procedural macros, but I can look into it.

> I don't think redb should be treating bincode in any special way. `serialize` and `deserialize` from bincode should not be re-exported as some default, I think.

I just added that because the procedural macro inserts the literal code, so without it, the user has to import bincode manually. I just thought it inconvenient, but I see your point.

dpc:
BTW. I have went with the same path in my code, and I wonder if the macros should be more generic, `serde` ones (`with = ...`, `serialize_with = `, `deserialize_with = `).


My code:

```rust
impl RedbValue for ItemValue {
    type SelfType<'a> = ItemValue;

    type AsBytes<'a> = Vec<u8>;

    fn fixed_width() -> Option<usize> {
        None
    }

    fn from_bytes<'a>(data: &'a [u8]) -> Self::SelfType<'a>
    where
        Self: 'a,
    {
        bincode::deserialize(data).expect("bincode deserialization error")
    }

    fn as_bytes<'a, 'b: 'a>(value: &'a Self::SelfType<'b>) -> Self::AsBytes<'a>
    where
        Self: 'a,
        Self: 'b,
    {
        bincode::serialize(value).expect("bincode serialization error")
    }

    fn type_name() -> redb::TypeName {
        redb::TypeName::new("item-data")
    }
}
```

could be just:

```
#[derive(RedbValue(serialize_with=bincode::serialize, deserialize_with=bincode::deserialize, name="item-value"))]
struct ItemValue { /* ... */ }
```

So far I have 2 structs using bincode, and two newtypes (that just delegate to inner-value), so I was planning to `macro_rules`-them as it's easy. But derive macro would be even better.

ChHecker:
So the arguments to derive macros the way you wrote them are apparently not possible. That's why I added those additional arguments like
```rust
#[derive(Clone, Debug, PartialEq, Deserialize, Serialize, RedbValue)]
#[fixed_width]
#[type_name("test-type")]
struct Test(usize);
```
I'd love if someone could prove me otherwise, as I'm worried if too many arguments would get crowded. Especially if I added `serialize_with` arguments as you mentioned.

dpc:
[`serde` does it similiarly so you're probably right](https://serde.rs/container-attrs.html).

Though like serde I'd do `#[redb(fixed_with)]`, etc. instead of "naked" versions.

dpc:
```rust
#[derive(Clone, Debug, PartialEq, Deserialize, Serialize, RedbValue)]
#[redb(fixed_width, type_name = "test-type", with="bincode")]
struct Test(usize);
```

Is the destiny. ;)

src/types.rs
```diff
@@ -383,3 +383,37 @@ be_impl!(i64);
...
+#[cfg(feature = "serialize")]
+#[macro_export]
+macro_rules! be_value_serialize {
+    ($t:ty) => {
+        impl RedbValue for $t {
+            type SelfType<'a> = Self;
+            type AsBytes<'a> = Vec<u8> where Self: 'a;
+
+            fn fixed_width() -> Option<usize> {
+                Some(std::mem::size_of::<$t>())
```
cberner, line 396: This doesn't seem right. Does `bincode::serialize` guarantee that the result is the same length as `size_of`?

cberner (changes requested):
So one concern I have with this is that users who use this macro may expect it to handle schema changes for them, and I'm guessing it doesn't.
The existing types in redb are safe against this because the tuple type encodes all the fields in the `TypeName`. Is there a way to do with these proc macros, so that all the types of the fields are included?

The case I have in mind is. Someone writes code like:
```
#[derive(Clone, Debug, PartialEq, Deserialize, Serialize, RedbValue)]
#[fixed_width]
#[type_name("test-type")]
struct Test(usize);
```

and then changes it to: `struct Test(isize)`. This should cause an error at runtime, when trying to read the previously stored data.

dpc:
> So one concern I have with this is that users who use this macro may expect it to handle schema changes for them, and I'm guessing it doesn't.

That is not something I would personally expect. Also not something I would necessarily waste bytes on.

One application I immediately see for redb is involving billions of tiny records, where every byte adds up.

cberner:
Ah, to clarify, I don't think it should support schema changes. But I do
think there should be an error if the user tries to change the schema

redb-derive/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+            fn as_bytes<'a, 'b: 'a>(value: &'a Self::SelfType<'b>) -> Self::AsBytes<'a>
+            where
+                Self: 'a,
+                Self: 'b,
+            {
+                redb::serialize(value).unwrap()
+            }
+
+            fn type_name() -> redb::TypeName {
+                redb::TypeName::new(#type_name)
```
cberner, line 48: I've never written a proc macro before. Is it possible to change this to include the types of all the fields in the struct? For example, the way I [did it for tuples](https://github.com/cberner/redb/blob/505e21915b681ac4e409b9b8e5b280ed5657d55e/src/tuple_types.rs#L75-L90). This would resolve my concern about users changing the type of a field and then getting corrupted data instead of an error. Something like `#type_name {#field_1_type, #field_2_type, #field_3_type...}`

dpc: Definitely possible.

BTW. This is stored only once per table, right? I guess my worries were needless if so.

cberner: Yes, once per table. And it's only checked once when the table is opened, so shouldn't add much overhead if it's long

dpc: In that case a must have.

cberner:
I left some comments. Also, this needs tests :)

src/types.rs
brxken128: Nope, and the v2 RCs of `bincode` don't even have support for determining what the size of the value would be after serialization.

cberner: I see. Well types that return `Some` from `fixed_width()` need to return a slice of the same length from `as_bytes()`

ChHecker:
I'm going to be honest, it seems like I vastly underestimated the implications of this PR. I'd love for these changes to somehow be implemented, but I neither have the time nor the required skill level. Therefore, I will close this PR for now.
Sorry if I wasted your time, I'm fairly new to contributing to open source code.

PR #668, opened 2023-08-30
Pin redb1 at 1.0.0

devrandom:
without pinning the version, Cargo chooses the latest semver compatible version, which is 1.1.0 - defeating the intention of the backwards-compatibility tests.


cberner:
Merged. Thanks!

PR #700, opened 2023-10-12
Add option to set a repair callback

### LINKED ISSUE #639: Provide a way to test if a database file needs repair before opening it
For large redb database files, the repair performed when opening a database that has not been shut down cleanly takes a long time (often more than 30 minutes for the ordinals index when running with `--index-sats`). We can't currently warn the user that a repair is necessary, so we often have users assuming `ord` hangs and then killing the process.

What would be the best way to see if `mem.needs_repair()` is true for a given database file before calling `Database::new`?

cberner:
This allows the user to abort the repair process and receive progress notifications

Fixes #639

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -352,7 +352,10 @@ impl Database {
             return Ok(true);
         }
 
-        Self::do_repair(&mut self.mem)?;
+        Self::do_repair(&mut self.mem, &|_| {}).map_err(|err| match err {
+            DatabaseError::Storage(storage_err) => storage_err,
+            _ => unreachable!(),
+        })?;
         self.mem.begin_writable()?;
 
         Ok(false)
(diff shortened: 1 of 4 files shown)


cberner:
@veryordinally / @raphjaph can you take a look at this API when you get a chance? Will merge it, if it looks like it covers your use case

casey:
Nice! This works for us. It looks like `Self::verify_primary_checksums(mem)` is called first, before the repair callback is invoked. Can this take a long time? If so, it would be nice to do a quick check if anything that takes a long time is going to happen, and call `repair_callback` first if so. Related, if `repair_callback` is called first, then having `progress` return `0` for the first call would avoid needing to check if we should skip printing the message, i.e., we could just do `if handle.progress() == 0.0 { print_message() }`.

cberner:
I think that's already implemented. It's just in the caller of `do_repair`. See here: https://github.com/cberner/redb/pull/700/files#diff-7d7d3a6ce5c2a3e8b39ce4f49b78a71365c8e06c0d9ddb78bc2dca68bfeb4a5bR683-R688

casey:
Ah, okay awesome. Looks good to me!

cberner:
I would recommend against `if handle.progress() == 0.0` though. The API doesn't guarantee that the first call will be with a progress value of `0.0` and I might find some reason to change it in the future. Although that does seem unlikely

PR #707, opened 2023-10-17
Add custom storage backends

DouglasDwyer:
I've been looking for a pure-Rust, embedded key value store that supports pluggable storage backends or a virtual file system interface (my use case involves storing data to the Javascript OPFS via WASM, which requires using Javascript file system APIs rather than `std::fs`). Unfortunately, there don't seem to be any available - but `redb` looks like a great project, and it was easy to add this feature to it. This PR abstracts the file system code with a `StorageBackend` trait, which the user can implement in order to customize how `redb` interacts with the file system.

Please let me know what you think, and if there are any other improvements that I should add!


cberner:
btw, if you run `just` in the source directory. It will run all the checks that the CI enforces

DouglasDwyer:
My apologies for the failed CI runs. They were a bit stricter than I expected, and I'm developing on Windows (also do not have `just` installed at present) so it's a bit difficult to check for issues on other targets lol
I appreciate you re-running them for me; sorry for any inconvenience.

DouglasDwyer:
Given the repeated issues, I switched to running the CI pipelines on my fork of `redb`. I managed to get all of them to pass, so the latest set of commits should hopefully pass in this repository too :)

src/db.rs
```diff
@@ -245,15 +273,22 @@ impl Database {
     /// * if the file does not exist, or is an empty file, a new database will be initialized in it
     /// * if the file is a valid redb database, it will be opened
     /// * otherwise this function will return an error
+    #[cfg(any(windows, unix, target_os = "wasi"))]
     pub fn create(path: impl AsRef<Path>) -> Result<Database, DatabaseError> {
         Self::builder().create(path)
     }
 
     /// Opens an existing redb database.
+    #[cfg(any(windows, unix, target_os = "wasi"))]
```
cberner, line 282: I'm especially unexcited about making the public API platform dependent. Why is this necessary?

src/db.rs
```diff
@@ -26,6 +31,29 @@ use crate::transactions::SAVEPOINT_TABLE;
 #[cfg(feature = "logging")]
 use log::{info, warn};
 
+/// Implements persistent storage for a database.
+pub trait StorageBackend: 'static + Debug + Send + Sync {
+    /// Whether the current storage backend is of zero length.
+    fn is_empty(&self) -> Result<bool, io::Error> {
+        Ok(self.len()? == 0)
+    }
```
cberner, line 39: I don't think this is used. Let's remove it?

cberner (changes requested):

src/db.rs
DouglasDwyer: Thank you for reviewing my changes. The reason is that the filesystem API is unavailable on certain targets (for instance `wasm32-unknown-unknown`) so without platform-gating these methods the crate fails to compile. This happens even on `master` currently, since an implementation for `LockedFile` is missing.

The two options that I see for working around this are:
- Making certain public methods only available on certain platforms (what I currently have)
- Creating a dummy `FileBackend` implementation for non-supported targets that just panics immediately at runtime

I initially went with the first option, but if you think the second one is preferable (or have a separate solution) please let me know and I will change it.

src/db.rs
```diff
@@ -10,11 +10,16 @@ use crate::{
     StorageError,
 };
 use crate::{ReadTransaction, Result, WriteTransaction};
-use std::fmt::{Display, Formatter};
+use std::fmt::{Debug, Display, Formatter};
+
+#[cfg(any(windows, unix, target_os = "wasi"))]
 use std::fs::{File, OpenOptions};
+use std::io;
+#[cfg(any(windows, unix, target_os = "wasi"))]
```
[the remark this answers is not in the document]
DouglasDwyer: This is directly related to the changes that make the public API platform dependent, please see my comment below for further explanation.

src/db.rs
DouglasDwyer: Indeed, it is not used - I added this because of the Clippy lint that forbids having a `len` method without `is_empty`. I can remove it and add an `#[allow(clippy::len_without_is_empty)]` if that would be better!

src/db.rs
cberner: Ah, I see. `wasm64-unknown-unknown` is Tier 3, so I'd consider any support of it experimental at best. I'd like to keep the codebase as free of `cfg` as possible, because it adds maintenance. So ya, lets remove these and if it's possible to cleanly implement option 2 for that wasm target, I'm fine having a separate backend for it.

Alternately, if there's a flag that indicates the filesystem API is available on a platform, then I'd be fine using that here. Basically, I want this to work on all platform where it's supported, without the need to add those platforms explicitly to this `cfg`

src/db.rs
cberner: Oh ya, let's add that `allow`. That seems like a silly Clippy rule :)

cberner (changes requested):
Some minor change requests. Otherwise this lgtm :)

src/lib.rs
```diff
@@ -66,7 +66,8 @@ pub use multimap_table::{
 };
 pub use table::{Drain, DrainFilter, Range, ReadOnlyTable, ReadableTable, Table};
 pub use transactions::{DatabaseStats, Durability, ReadTransaction, WriteTransaction};
-pub use tree_store::{AccessGuard, AccessGuardMut, Savepoint};
+pub use tree_store::file_backend::FileBackend;
```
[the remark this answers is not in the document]
DouglasDwyer: I would argue that exposing `FileBackend` allows others to extend this crate in useful ways. For example, imagine that I wanted to insert some middleware between the storage backend and `redb` - such as middleware that counts the number of times that reads and writes occur, or does some additional logging about them. I could do this by wrapping a `StorageBackend` in a new middleware object, but this is impossible if the underlying `StorageBackend` implementation is not exposed.

src/tree_store/page_store/file_backend/unix.rs
```diff
@@ -0,0 +1,110 @@
...
+pub struct FileBackend {
+    file: File,
+}
+
+impl FileBackend {
+    /// Creates a new backend which stores data to the given file.
+    // This is a no-op until we get flock in wasi-libc.
+    // Delete this function when we get flock.
+    #[cfg(target_os = "wasi")]
+    pub fn new(file: File) -> Result<Self> {
```
cberner, line 29: This should be `Result<Self, DatabaseError>`

src/lib.rs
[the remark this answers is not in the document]
cberner: Ok, that's reasonable. Let's include it. There's a `new` method that needs fixing though. I left a comment about that.

src/tree_store/page_store/memory_backend.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn write(&self) -> RwLockWriteGuard<'_, Vec<u8>> {
+        self.0.write().expect("Could not acquire write lock.")
+    }
+}
+
+impl From<Vec<u8>> for MemoryBackend {
+    fn from(value: Vec<u8>) -> Self {
+        Self(RwLock::new(value))
+    }
+}
```
cberner, line 46: And I think remove this too

cberner (changes requested):

cberner (approved):

C1710:
Sorry for asking this maybe very simple question:  
Does this Trait actually directly allow an OPFS-based implementation? After all, all types from `web-sys` are `!Send` and `!Sync`, which is incompatible with the requirements of `StorageBackend`. Am I missing something?

DouglasDwyer:
> Does this Trait actually directly allow an OPFS-based implementation? After all, all types from web-sys are !Send and !Sync, which is incompatible with the requirements of StorageBackend. Am I missing something?

That's a good question. The approach I have in mind (although I haven't gotten a chance to implement this yet) would look as follows:

- A background thread is started (using the `wasm_thread` crate) which opens a connection to the OPFS using `web-sys` types. The background thread has two channels (one for reading data and one for writing data) or something similar.
- The main thread (or other thread) creates a `Database` with a custom `StorageBackend`. The `StorageBackend` ferries all read/write requests over the channels to the worker thread. It then waits for the results, blocking with the `wasm_sync` synchronization primitives.
- The worker thread uses the OPFS API to perform the reads/writes, and sends the data back to the main thread over the channels.

This approach, while roundabout, works around the web's annoying restrictions; the `StorageBackend` would be usable from multiple threads and fully synchronous, like before.

PR #718, opened 2023-11-10
Implement RedbValue and RedbKey for char

casey:
Just noticed that key value impls for char were missing, so I added them.


cberner:
Merged. Thanks!

PR #723, opened 2023-11-24
New untyped stats API

### LINKED ISSUE #398: Include table stats in `DatabaseStats`
I'm auditing space usage in the `ord` index, and having information about individual tables would be super useful. In particular, the total size of each table in bytes, and how many keys it holds.

cberner:
Fixes #398

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,9 +62,13 @@ pub use error::{
     TransactionError,
 };
 pub use multimap_table::{
-    MultimapRange, MultimapTable, MultimapValue, ReadOnlyMultimapTable, ReadableMultimapTable,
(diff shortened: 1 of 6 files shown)


casey:
Nice, that's great. This [PR](https://github.com/ordinals/ord/pull/2747) uses the new API. The only issue I see is that database stats are only available on write transactions, whereas `open_untyped_table` is only available on write transactions, so in order to get all stats, you need to open a read transaction and a write transaction.

cberner:
ya, I was thinking about adding `open_untyped_table()` to write transactions too. But then I couldn't decide whether it should return a `ReadOnlyUntypedTable` or an `UntypedTable` that might in the future have mutation methods (`clear()` would be an option).

casey:
Seems reasonable to me. Thanks for adding this!

PR #730, opened 2023-12-08
docs: add Cargo.toml copy paste

kruserr:
This would make it 1 step smoother to use redb in new users cargo projects.


cberner:
Thanks but I think the dependency to add is obvious enough, and prefer keeping the docs short and simple

PR #732, opened 2023-12-21
Add lifetime-free range methods

### LINKED ISSUE #725: Add read API without lifetimes
It would be nice to have a read transaction API that uses reference counting instead of lifetimes. This would allow callers to return the values/iterators retrieved from a table

cberner:
Fixes #725

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -6,8 +6,7 @@ use crate::tree_store::{
 };
 use crate::types::{RedbKey, RedbValue};
 use crate::{
-    CompactionError, DatabaseError, Durability, ReadOnlyTable, ReadableTable, SavepointError,
-    StorageError,
+    CompactionError, DatabaseError, Durability, ReadOnlyTable, SavepointError, StorageError,
 };
 use crate::{ReadTransaction, Result, WriteTransaction};
 use std::fmt::{Debug, Display, Formatter};
--- a/src/multimap_table.rs
+++ b/src/multimap_table.rs
@@ -1296,6 +1296,35 @@ impl<'txn, K: RedbKey + 'static, V: RedbKey + 'static> ReadOnlyMultimapTable<'tx
             _lifetime: Default::default(),
         })
     }
+
+    /// This method is like [`ReadableMultimapTable::get()`], but the iterator is reference counted and keeps the transaction
+    /// alive until it is dropped.
+    pub fn get<'a>(&self, key: impl Borrow<K::SelfType<'a>>) -> Result<MultimapValue<'static, V>> {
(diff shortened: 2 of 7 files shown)


Frando:
Hi, this is great! In [iroh](https://github.com/n0-computer/iroh/) we have to return owned iterators to tables, and thus far resorted to the [ouroboros](https://crates.io/crates/ouroboros) crate, which allows to create self-referential structs through very unsafe means. We'd love to get rid of that, because ouroboros was plagued with unsoundness issues in the past.

However [one of our owned iterators](https://github.com/n0-computer/iroh/blob/main/iroh-sync/src/store/fs/ranges.rs#L204) contains a `Range` to table A, but returns values from table B (table A is an index table, and table B contains the actual values). To remove the need for ouroboros here, `redb` would have to provide an `ArcReadOnlyTable` in addition to the `ArcRange`. From reading the code, this seems feasable to add, in the same way as the `ArcRange` works. What do you think?

cberner:
I think I'll be able to just remove all the lifetimes from `ReadOnlyTable`. I haven't gotten to prototyping it yet, but one of my goals for 2.0 is to only have lifetimes on objects used for writing

jeromegn:
I've seen more functions and structs use the `owned` or `Owned` prefix/suffix. Maybe there's a future where there's no `Arc` usage in there and suddenly the naming makes less sense :)

> one of my goals for 2.0 is to only have lifetimes on objects used for writing

Is there a limitation requiring a lifetime for writes? I assume that means `WriteTransaction` and `Table` would still require a lifetime bound.

My current experiment is making a DBMS backed by `redb`, using `rhai` as its "query language". `rhai` requires static lifetimes for custom types. My temporary solution is to use a `once_cell::sync::Lazy` to get a `'static` redb database. I do have to open tables very often because they also have lifetimes. I can't keep them around as custom objects passed via `rhai`. Anyway, all this to say that "owned" types would help a lot for these kinds of things!

cberner:
> Is there a limitation requiring a lifetime for writes? I assume that means WriteTransaction and Table would still require a lifetime bound.

I think the lifetime can be removed from WriteTransaction, and maybe Table too. The `Range` returned from Table has to have a lifetime though, because it effectively borrows from `Table` which is mutable

cberner:
@Frando I think the latest master will allow you remove that. read-only tables are now returned with the `'static` lifetime

Frando:
@cberner I'm struggling to create an owned Iterator for a range. Do you have an example somewhere?

cberner:
Great! Ya, it's not released yet because I want to get a couple more breaking changes in before I release 2.0

PR #734, opened 2023-12-26
Use curly braces for `tuple_impl!` calls

casey:
Macros can be called with `()`, `{}`, and `[]`. Using `{}` is slightly more idiomatic for things that expand to items (impl blocks, struct definitions, etc), and has the benefit that `cargo fmt` doesn't try to format them.


cberner:
Ah nice!

PR #791, opened 2024-04-05
Make insert() docs more explicit

### LINKED ISSUE #787: Please clarify if `Table.insert()` does replace the attribute if it exists
Hi,

From reading the doc, it is unclear if [`Table.instert()`](https://docs.rs/redb/latest/redb/struct.Table.html#method.insert) does replace or not existing entries.

Thanks,
Alex

cberner:
Fixes #787


abique:
Thank you :+1:

PR #805, opened 2024-05-04
RFC: Add wrapper type to handle keys and values based on bincode serialization

adamreichold:
The first commit is just to make current Clippy build cleanly and can be drop if so desired.

Not sure if you want such additions as this could be considered "playing favourites" with a particular format and you might see requests to add many others. OTOH, Bincode is a pretty decent default for binary serialization formats of small messages.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,8 @@ crate-type = ["cdylib", "rlib"]
 pyo3-build-config = { version = "0.20.0", optional = true }
 
 [dependencies]
-log = {version = "0.4.17", optional = true }
-pyo3 = {version = "0.20.0", features=["extension-module", "abi3-py37"], optional = true }
+log = { version = "0.4.17", optional = true }
+pyo3 = { version = "0.20.0", features=["extension-module", "abi3-py37"], optional = true }
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.104"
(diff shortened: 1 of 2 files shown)


cberner:
Hmm, ya I'm not sure I want to add this. My philosophy has been to leave out extra features which users can easily implement on their own, but I'll give it some more thought. If you want to send that first Clippy commit has a separate PR I'd be happy to merge that

adamreichold:
> Hmm, ya I'm not sure I want to add this. My philosophy has been to leave out extra features which users can easily implement on their own, but I'll give it some more thought.

Maybe as one argument to why this could be useful upstream: While the code can easily be implemented downstream, there is also basically one way to write it and hence every downstream project will contain a copy of this same code creating quite a bit of unnecessary busy work. So I think shipping something like this could be good community service and make the project more approachable.

> If you want to send that first Clippy commit has a separate PR I'd be happy to merge that

This is #807

cberner:
Sorry it took me a while to reply here. I thought about this some more and decided not to merge it now. If you'd like to submit it as a PR in the `examples/` directory I'd be happy to put it there, and then people can copy if they want

adamreichold:
Understood. Will rework this into an example.

adamreichold:
Reworked into an example based on the existing `int_keys` one.

cberner:
Merged, thanks!

Issue #808, opened 2024-05-12
Using direct I/O on Linux


adamreichold:
Understanding that redb manages it owns user space page cache, I used the small diff

<details>
<summary>changes to make the lmdb_benchmark use direct I/O</summary>

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 649c21b..0819edd 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ pyo3-build-config = { version = "0.20.0", optional = true }
 [dependencies]
 log = {version = "0.4.17", optional = true }
 pyo3 = {version = "0.20.0", features=["extension-module", "abi3-py37"], optional = true }
+rustix = { version = "0.38.34", features = ["fs"] }
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.104"
diff --git a/benches/lmdb_benchmark.rs b/benches/lmdb_benchmark.rs
index 53b79b1..1cf2f88 100644
--- a/benches/lmdb_benchmark.rs
+++ b/benches/lmdb_benchmark.rs
@@ -297,7 +297,25 @@ fn main() {
         benchmark(table)
     };
 
-    let lmdb_results = {
+    let redb_directio_results = {
+        let tmpfile: NamedTempFile = NamedTempFile::new_in(&tmpdir).unwrap();
+        use std::fs::OpenOptions;
+        use std::os::unix::fs::OpenOptionsExt;
+        let file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .custom_flags(libc::O_DIRECT)
+            .open(tmpfile.path())
+            .unwrap();
+        let db = redb::Database::builder()
+            .set_cache_size(4 * 1024 * 1024 * 1024)
+            .create_file(file)
+            .unwrap();
+        let table = RedbBenchDatabase::new(&db);
+        benchmark(table)
+    };
+
+    /*let lmdb_results = {
         let tmpfile: TempDir = tempfile::tempdir_in(&tmpdir).unwrap();
         let env = lmdb::Environment::new().open(tmpfile.path()).unwrap();
         env.set_map_size(4096 * 1024 * 1024).unwrap();
@@ -325,7 +343,7 @@ fn main() {
         let db = sanakirja::Env::new(tmpfile.path(), 4096 * 1024 * 1024, 2).unwrap();
         let table = SanakirjaBenchDatabase::new(&db);
         benchmark(table)
-    };
+    };*/
 
[...]

PR #809, opened 2024-05-12
Move cleaned pages into read cache instead of discarding them.

adamreichold:
Just after dirty pages have been written out to disk, they are currently discard and need to be read in again when these pages are accessed in the future.

This changes moves them into the read cache instead thereby keeping the clean data available and cache hot in memory.

It turned out that this was also what was making direct I/O slower in #808, i.e. this was not that noticeable when buffered I/O is use because the pages were read back in from the kernel page cache most of the time but it should be a win in both scenarios.

### DIFF
--- a/src/tree_store/page_store/cached_file.rs
+++ b/src/tree_store/page_store/cached_file.rs
@@ -303,6 +303,7 @@ impl PagedCachedFile {
             } else {
                 self.read_cache_bytes
                     .fetch_sub(buffer.len(), Ordering::AcqRel);
+                break;
             }
         }
         for (offset, buffer) in write_buffer.low_pri_cache.iter_mut() {
@@ -318,6 +319,7 @@ impl PagedCachedFile {
             } else {
                 self.read_cache_bytes
                     .fetch_sub(buffer.len(), Ordering::AcqRel);
+                break;
             }
         }
         self.write_buffer_bytes.store(0, Ordering::Release);
--- a/src/tree_store/page_store/cached_file.rs
+++ b/src/tree_store/page_store/cached_file.rs
@@ -225,7 +225,7 @@ pub(super) struct PagedCachedFile {
     #[cfg(feature = "cache_metrics")]
     reads_hits: AtomicU64,
     fsync_failed: AtomicBool,
-    read_cache: Vec<RwLock<PrioritizedCache>>,
+    read_cache: Box<[RwLock<PrioritizedCache>]>,
     // TODO: maybe move this cache to WriteTransaction?
     write_buffer: Arc<Mutex<PrioritizedWriteCache>>,
 }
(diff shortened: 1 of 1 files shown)


adamreichold:
The intermediate fuzzer failure was me trying to be clever and avoid reference count bump by `Option::take`ing the buffers out of the write cache but that was not so clever if one of the intermediate writes would fail, I would never call `clear` and leave the empty buffers in place for the next flush or another call to write to assert on them. Should be fixed now.

adamreichold:
> The intermediate fuzzer failure was me trying to be clever and avoid reference count bump by `Option::take`ing the buffers out of the write cache but that was not so clever if one of the intermediate writes would fail, I would never call `clear` and leave the empty buffers in place for the next flush or another call to write to assert on them. Should be fixed now.

Turns out I can avoid that atomic reference count bump, but I think it is safer to just do this in two passes in any case.

src/tree_store/page_store/cached_file.rs
```diff
@@ -291,6 +290,38 @@ impl PagedCachedFile {
...
+        for (offset, buffer) in write_buffer.cache.iter_mut() {
+            let buffer = buffer.take().unwrap();
+            let cache_size = self
+                .read_cache_bytes
+                .fetch_add(buffer.len(), Ordering::AcqRel);
+
+            if cache_size + buffer.len() <= self.max_read_cache_bytes {
+                let cache_slot: usize = (offset % Self::lock_stripes()).try_into().unwrap();
+                let mut lock = self.read_cache[cache_slot].write().unwrap();
+                lock.insert(*offset, buffer.clone(), CachePriority::High);
```
cberner, line 302: Is it necessary to call `.clone()` here? It looks like it should be possible to just pass `buffer`

src/tree_store/page_store/cached_file.rs
```diff
@@ -291,6 +290,38 @@ impl PagedCachedFile {
...
+        for (offset, buffer) in write_buffer.low_pri_cache.iter_mut() {
+            let buffer = buffer.take().unwrap();
+            let cache_size = self
+                .read_cache_bytes
+                .fetch_add(buffer.len(), Ordering::AcqRel);
+
+            if cache_size + buffer.len() <= self.max_read_cache_bytes {
+                let cache_slot: usize = (offset % Self::lock_stripes()).try_into().unwrap();
+                let mut lock = self.read_cache[cache_slot].write().unwrap();
+                lock.insert(*offset, buffer.clone(), CachePriority::Low);
```
cberner, line 318: Is it necessary to call `.clone()` here? It looks like it should be possible to just pass `buffer`

src/tree_store/page_store/cached_file.rs
adamreichold: No, that is a left-over from doing this in a single pass via `.iter`. Will fix both cases.

cberner:
Merged. Thanks!

PR #813, opened 2024-06-02
Fix ReadTransaction::close()

### LINKED ISSUE #812: Error ReadTransactionStillInUse(ReadTransaction) when try to close transaction
I just want to make sure that the transaction is over

```rust
#[test]
fn test_close_read_transaction() {
    let db_path = "test_close_read_transaction.db";
    let db = redb::Database::create(db_path).unwrap();
    let read_tx = db.begin_read().unwrap();
    let close_result = read_tx.close().unwrap();
}
```
and the test result:
```rust Error
---- auth::redb_persistence::test::test_close_read_transaction stdout ----
thread 'redb_persistence::test::test_close_read_transaction' panicked at redb_persistence.rs:207:44:
called `Result::unwrap()` on an `Err` value: ReadTransactionStillInUse(ReadTransaction)

### PR DESCRIPTION
**Pull request by cberner**
Previously it always returned ReadTransactionStillInUse

Fixes #812

### DIFF
--- a/src/transactions.rs
+++ b/src/transactions.rs
@@ -1183,7 +1183,6 @@ impl Drop for WriteTransaction {
 pub struct ReadTransaction {
     mem: Arc<TransactionalMemory>,
     tree: TableTree,
-    transaction_guard: Arc<TransactionGuard>,
 }
 
 impl ReadTransaction {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/transactions.rs [hunk 1]
```diff
@@ -1298,7 +1296,9 @@ impl ReadTransaction {
     ///
     /// Returns `ReadTransactionStillInUse` error if a table or other object retrieved from the transaction still references this transaction
     pub fn close(self) -> Result<(), TransactionError> {
-        if Arc::strong_count(&self.transaction_guard) > 1 {
+        let cloned = self.tree.clone_transaction_guard();
```
adamreichold, line 1299: As an alternative, one could have added an accessor like

```rust
fn transaction_guard(&self) -> &Arc<TransactionGuard>;
```

to avoid the unnecessary clone here. The cloning itself could then still just be written as

```rust
Arc::clone(self.tree.transaction_gurard())
```

without adding another method.

cberner: Oh ya, that's cleaner. https://github.com/cberner/redb/pull/814

PR #817, opened 2024-06-27
chore: production ready borsh(with schema) example

dzmitry-lahoda:


cberner:
Thanks for working on this, but I don't want to increase the MSRV if it's only to support a new example. Also, I think the bincode example is simpler and similar enough to this one, that it is a better example for people to work from.

dzmitry-lahoda:
when you will update rust version for some reason, would it be possible to look at PR again? 

as for bincode, it is only bincode_keys, not values (at least by name : ), also it does not shows shema usage (along with no fixed size). so borsh is example of how schema can be used to automate keys/values for redb.

thank you

PR #839, opened 2024-08-04
benchmark: increase rocksdb block cache capacity

marvin-j97:
Compared to redb (and sled), RocksDB gets way too little cache capacity (32MB), which negatively impacts its read performance. Setting it to 4 GB (same as redb) gives it a 1.5x read performance boost on my system.

I haven't updated the readme, because it's system dependent, so you would need to rerun it on yours.


cberner:
Merged. Thanks!

PR #853, opened 2024-08-28
Add compaction to lmdb_benchmark

cberner:
Note: compact() is currently too slow to practically run with the default 1M elements. It takes 43secs on my computer when set to 1000 elements

### DIFF
--- a/benches/lmdb_benchmark.rs
+++ b/benches/lmdb_benchmark.rs
@@ -297,6 +297,7 @@ fn database_size(path: &Path) -> u64 {
 
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 enum ResultType {
+    NA,
     Duration(Duration),
     SizeInBytes(u64),
 }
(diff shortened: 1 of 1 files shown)


Kerollmops:
Hey 👋

As I understand correctly, you are compacting the redb database but not the other databases just before measuring the disk size?

It feel unfair to me. Doesn't RocksDB and Sled support compaction, too?

This kind of benchmark is only ran once every now and then, right? Can't we display the very long compaction time of LMDB (and others) with the associated size. Another solution could be to display both, before and after compaction.

Have a nice end of the year 🎄

cberner:
Can you point me to the method to call to compact the other dbs? I looked
but didn't find one

Kerollmops:
For sure 😌

You can use [the heed `Env::copy_to_file/fd` method](https://docs.rs/heed/latest/heed/struct.Env.html#method.copy_to_file) to write the content of the database into another file or fd. Note that you then need to reopen the destination file to read the content and, probably open it with the `NO_SUB_DIR` option.

When it comes to RocksDB, calling [the `DBCommon::compact_range/_opt` method](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.compact_range) on every database should be enough to compact everything. I don't know if it is blocking but you should be able to decide about that with some options.

PR #855, opened 2024-08-31
feat: make Database `Clone`

cfcosta:
I don't know if you require this, feel free to close if not. I added it to my own fork because I need it, as being `Clone` is required to share an instance on axum, for example.

Since all internal tools are already guarded by Arcs, there's no reason to block the user from being able to `clone` the Database.

This PR adds support for it.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -310,6 +310,7 @@ impl Drop for TransactionGuard {
 /// # Ok(())
 /// # }
 /// ```
+#[derive(Clone)]
 pub struct Database {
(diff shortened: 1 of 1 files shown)


adamreichold:
> Since all internal tools are already guarded by Arcs, there's no reason to block the user from being able to clone the Database.

I am not sure this is correct: The methods `check_integrity` and `compact` take `&mut self` meaning they currently enjoy exclusive access to the underlying database. When cloning a `Database` is possible, they could be called concurrently.

cberner:
Ya, this is not correct

PR #871, opened 2024-09-30
Optimize first() and last()

cberner:
This change makes these methods almost 2x faster, by reducing the number of tree traversals from two to one.

### DIFF
--- a/src/tree_store/btree.rs
+++ b/src/tree_store/btree.rs
@@ -474,6 +474,18 @@ impl<K: Key + 'static, V: Value + 'static> BtreeMut<'_, K, V> {
         self.read_tree()?.get(key)
     }
 
+    pub(crate) fn first(
+        &self,
+    ) -> Result<Option<(AccessGuard<'static, K>, AccessGuard<'static, V>)>> {
+        self.read_tree()?.first()
+    }
+
+    pub(crate) fn last(
+        &self,
+    ) -> Result<Option<(AccessGuard<'static, K>, AccessGuard<'static, V>)>> {
+        self.read_tree()?.last()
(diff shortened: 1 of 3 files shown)


src/tree_store/btree.rs
```diff
@@ -726,6 +738,71 @@ impl<K: Key, V: Value> Btree<K, V> {
...
+                let accessor = LeafAccessor::new(page.memory(), K::fixed_width(), V::fixed_width());
+                let (key_range, value_range) = accessor.entry_ranges(0).unwrap();
+                let key_guard = AccessGuard::with_page(page.clone(), key_range);
+                let value_guard = AccessGuard::with_page(page, value_range);
+                Ok(Some((key_guard, value_guard)))
+            }
+            BRANCH => {
+                let accessor = BranchAccessor::new(&page, K::fixed_width());
+                let child_page = accessor.child_page(0).unwrap();
+                self.first_helper(self.mem.get_page_extended(child_page, self.hint)?)
```
adamreichold, line 767: I understand tree depth is somewhat limited, but this looks highly amenable to an iterative implementation by basically putting all of this in a loop and writing

```rust
page = self.mem.get_page_extended(child_page, self.hint)?;
```

here with a

```
return Ok(Some((key_guard, value_guard)));
```

above to avoid the any questions of call overhead and/or stack usage.

cberner:
Thanks for the suggestion. I'm going to stick with this recursive
implementation for consistency with most of the rest of the code base, but
if you have any benchmarks on the overhead of the function calls, I'd be
very interested and open to changing if there's meaningful overhead

PR #875, opened 2024-10-08
Switch page cache to approximate LRU

cberner:


adamreichold:
Two questions/remarks:

* Is keeping `PrioritizedWriteCache` around an artifact or is there something non-obvious of the priority mechanism left?
* Adding a rationale for switch to the commit would be nice. Most importantly what problem of the current does this resolve?

cberner:
It doesn't do any prioritization. It's just a wrapper around LRUCache to handle the way that cached pages are "borrowed" by the rest of the codebase.

I updated the commit

adamreichold:
> I updated the commit

Thank you! Much easier to follow for me now.

PR #884, opened 2024-11-01
Fix region tracker page leak and initial sizing

mconst:
When allocating a new region tracker, we need to free the old one.  Also, when recreating the region tracker during repair, it has to be large enough for the actual number of regions (which might be more than INITIAL_REGIONS).

I've expanded the existing out_of_regions() test case to exercise both of these code paths; the test will fail without these fixes.


cberner:
Nice find, thanks for the fixes!

PR #886, opened 2024-11-01
Turn on pedantic clippy lints where possible

casey:
I like the pedantic clippy lints, since they often turn up things that are either a little suspicious, or which just have better alternatives.

I went through all the `clippy::pedantic` lints which weren't enabled, and one by one either denied or enabled them, usually with some kind of rationale. Sometimes I allowed lints that should probably be denied, either because they're subjective, or would take more knowledge of `redb` than I have to actually fix.

I tried to do each one in a separate commit, to keep things readable and let you review the rationale for each, which are in the commit messages.

Finally, I added `clippy::pedantic` to the `#![deny]` block, and tried to remove any `allow` which was not actually being triggered.

As of Rust 1.74, you can put lints in `Cargo.toml`, which is a nice way to enforce them project wide, but until then putting them at the top of `lib.rs` is reasonable.

Feel free to take or leave anything in this PR, since obviously a lot of it is subjective. And definitely squash this commit before merging!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ Cargo.lock
 # Profiling
 perf.data*
 flamegraph.svg
+
+# benchmark and test temporary files
+/.tmp*
--- a/src/complex_types.rs
+++ b/src/complex_types.rs
@@ -9,7 +9,6 @@ fn encode_varint_len(len: usize, output: &mut Vec<u8>) {
         output.push(254);
         output.extend_from_slice(&u16_len.to_le_bytes())
     } else {
-        assert!(len <= u32::MAX as usize);
         let u32_len: u32 = len.try_into().unwrap();
         output.push(255);
         output.extend_from_slice(&u32_len.to_le_bytes())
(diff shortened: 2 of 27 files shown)


casey:
Just noticed the errors on CI on x86, will fix those and undraft.

casey:
Fixed! I didn't realize that the xxh3 module was copied from another repo, so I reverted the changes and added `#[allow(clippy::pedantic)]` to the whole module.

cberner:
Thanks! I'll look through these

cberner:
These seem great, thanks!

casey:
Nice! Consider taking a look at these and seeing if they seem worth enabling. In particular, `Option<Option<T>>` is a travesty 😂

```
clippy::if_not_else
clippy::inline_always
clippy::let_underscore_drop
clippy::must_use_candidate
clippy::option_option
clippy::redundant_closure_for_method_calls
clippy::unnecessary_wraps
clippy::unreadable_literal
```

cberner:
Haha, ya that return type is a bit sketch =P

Issue #878, opened 2024-10-18
Avoid recovery when using Durability::Paranoid


mconst:
Currently, redb sets the needs_recovery flag as soon as you open the database, which means any unclean shutdown will trigger recovery.  For large databases, this isn't a great experience -- it means a simple Ctrl-C can end up locking you out of your database for a half-hour or more while it recovers.  It would be really nice if there were a way to avoid this.

Of course, recovery is unavoidable after an unclean shutdown when using Durability::Immediate, since there's no way to know whether the last commit was complete.  But if you're using Durability::Paranoid, it seems like recovery shouldn't be necessary!  This would let applications opt into slower commits in exchange for not needing recovery, which would be a big usability win when the database is large.

To make this work, I believe redb would need to write out the allocator state as part of Durability::Paranoid commits (or this could be a new durability level above Paranoid, to avoid slowing down existing Durability::Paranoid users who don't care about recovery speed).  But I think it could be done in a fully backwards- and forwards-compatible way.

Does this sound right to you?  And if so, do you think it would be a worthwhile change?

cberner:
Ya, that seems useful. It should probably be separate from `Paranoid` since the allocator state grows linearly in the size of the database, and could be quite large.

Do you think a new `Durability` level, or a separate `Database::flush(&self)` method is more idiomatic?

mconst:
Either way works, but I think a new Durability level would be nicer.  This minimizes the performance impact (since calling `Database::flush()` would mean an extra `fsync()`, whereas a new Durability level lets us do it as part of the regular two-phase commit), and it also avoids having a gap in between `commit()` and `Database::flush()` where a crash would still require recovery.

If it's useful, I can describe the design I had in mind -- I haven't implemented it yet, but I think I have a decent idea of how it would work.

cberner:
Ya, a new Durability level seems good to me. Happy to merge a PR if you want to implement it!

mconst:
Awesome!  I'll try implementing this and I'll send you a PR in a week or two.

mconst:
I'm just finishing up the code here; I'll have a PR for you soon!  In the meantime, I've also got a few small fixes for preexisting redb bugs that I ran into -- I'm going to submit those first, because it'll make testing the main PR easier.

PR #888, opened 2024-11-04
Fix several crashes in repair from corrupted pages

mconst:
During repair, it's normal to encounter a partially-written tree, which can contain arbitrary invalid data.  We can't crash when that happens; we need to gracefully fail so the repair can try the other commit slot instead.

This first batch of fixes only covers the crashes most likely to occur in practice: the ones that can happen on a 64-bit machine, without malicious input.  There are several more arithmetic overflows that can only happen on 32-bit machines, which are simple to fix but I haven't done yet (I can probably look at these after #878).  There are also a number of trickier crashes that can only happen when there's corrupt data with a valid checksum, which is unlikely to happen accidentally but definitely possible with malicious input.  #878 will greatly improve the situation there.

If it's useful, I have .redb files which will reproduce most of these crashes; if you'd like them, just let me know.


cberner (approved):
Thanks!

PR #899, opened 2024-11-19
Don't mark the region tracker page as allocated

mconst:
When writing the allocator state table, it's better not to include the region tracker page as one of the allocated pages. Instead, we should allocate the region tracker page during quick-repair after loading the rest of the allocator state, exactly like we do for full repair.

This avoids an unlikely corruption that could happen for databases larger than 4 TB, if we crash on shutdown after writing out a new region tracker page number but before clearing the recovery_required bit. It also means that in the future, when we drop support for the old allocator state format, we'll be able to get rid of the region tracker page entirely instead of having to keep around some of the allocation code for compatibility.

### DIFF
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -1129,7 +1129,7 @@ impl TransactionalMemory {
 
 impl Drop for TransactionalMemory {
     fn drop(&mut self) {
-        if thread::panicking() {
+        if thread::panicking() || self.needs_recovery.load(Ordering::Acquire) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


mconst:
Okay, I think this is ready.  Sorry for the delay!  The first two commits are fixes for preexisting region-tracker bugs I ran into; the last one is the main change.

cberner:
Ah nice, ya I like this! I think this conflicted with your other PR that I just merged. Can you rebase and I'll merge this one too?

cberner (approved):

mconst:
Oh yeah, sorry about that!  Here's the rebased version.

PR #900, opened 2024-11-20
[Benchmark] Explicitly set cache sizes

marvin-j97:
Turns out sled's cache size was still set incorrectly, but it doesn't really make an impact in the results.

I also updated the results because they were very outdated, feel free to run them on your hardware.


cberner:
Thanks! I merged the first commit, and will re-run them on my machine

PR #912, opened 2024-12-12
Some very small adj.

yukibtc:
* Derive `Clone` for `Database`
* Change `compact` method to take `&self` instead of `&mut self`

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -423,7 +423,7 @@ impl Database {
     /// Compacts the database file
     ///
     /// Returns `true` if compaction was performed, and `false` if no futher compaction was possible
-    pub fn compact(&mut self) -> Result<bool, CompactionError> {
+    pub fn compact(&self) -> Result<bool, CompactionError> {
         if self
             .transaction_tracker
             .oldest_live_read_transaction()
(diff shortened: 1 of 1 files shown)


cberner:
This is not safe. `compact()` needs to take `&mut self` to ensure there are no outstanding transactions since it needs to relocate all the pages

yukibtc:
Is the commit related to the derive `Clone` acceptable? If looks good, I'll reopen a PR with only that commit.

---

> This is not safe. compact() needs to take &mut self to ensure there are no outstanding transactions since it needs to relocate all the pages

Ah, sorry, I saw now the comment in the code. 
Is there a way we can remove the `&mut self` from the `compact`? At the moment this force me to wrap the `Database` inside an `Arc<RwLock<T>>`, only to support compaction.

cberner:
> Is the commit related to the derive Clone acceptable? If looks good, I'll reopen a PR with only that commit.

No, that has the same problem as removing `&mut` from `compact()`. Then it would be possible to have two `Database`s and call `compact()` while the other had outstanding transactions

cberner:
> Is there a way we can remove the &mut self from the compact?

I don't think so. This is an inherent design limitation of the MVCC model in redb

Issue #555, opened 2023-04-15
Question about usage in async code


algermissen:
I am wondering how to properly use redb in multithreaded, async IO code (using Tokio). Can someone clarify the following?

Where and how many Database variables do I have to create? Do I create one in main and wrap it into an Arc<Mutex<..>> and clone it between threads? [Edit: I have just seen that Database and ReadTransactions are Send and Sync and WriteTransaction is Sync - I'll try to work from that and just share everything :-)  ... but still it would be nice the hear what others think]

What should be shared across threads and what should not be shared?  I saw the multithreaded benchmark, but that only works when I have control over what happens on which thread - in async IO code, that kind of control is not possible and I wonder how to integrate redb into async programs properly.

Any ideas or pointers?

Jan

cberner:
I haven't tried async, but redb does its own concurrency control, so you don't need the `Mutex`. Just `Arc<Database>` should work fine

algermissen:
Thank you - yes, seems to work just fine.

PR #933, opened 2025-01-12
Add multithreaded example

quietlychris:
I've seen a couple questions pop up in the issues about using `redb` in a multi-threaded context or async-task context such as #555. I thought it might be useful to have an example of what that might look like. If anything's incorrect or there are changes that would make this more representative, just let me know!

### DIFF
--- a/examples/multithread.rs
+++ b/examples/multithread.rs
@@ -12,9 +12,9 @@ fn main() -> Result<(), Error> {
     let write_txn = db.begin_write()?;
     {
         let mut table = write_txn.open_table(definition)?;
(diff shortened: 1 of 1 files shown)


cberner:
I think the example would be better if you had some, say 8, threads performing reads, while a few, say 2, threads did the writes. One feature that's useful to demonstrate is that readers are not blocked by writers

quietlychris:
Sorry for the delay! I've tried updating this to have 8 readers and writers; the `println!` statements seems a little messy, but not sure if there's a better way to demonstrate the behavior.

PR #944, opened 2025-02-03
Add fjall 2.5 to benchmark suite

EmmChriss:
I was curious about how redb fairs against [fjall](https://github.com/fjall-rs/fjall).

Others might be as well

### DIFF
--- a/benches/common.rs
+++ b/benches/common.rs
@@ -823,3 +823,148 @@ impl<'db, 'txn> BenchIterator for SanakirjaBenchIterator<'db, 'txn> {
         })
     }
 }
+
+pub struct FjallBenchDatabase<'a> {
+    db: &'a mut fjall::TxKeyspace,
+}
+
+impl<'a> FjallBenchDatabase<'a> {
+    #[allow(dead_code)]
+    pub fn new(db: &'a mut fjall::TxKeyspace) -> Self {
+        FjallBenchDatabase { db }
+    }
+}
+
+impl<'a> BenchDatabase for FjallBenchDatabase<'a> {
+    type W<'db>
+        = FjallBenchWriteTransaction<'db>
+    where
+        Self: 'db;
+    type R<'db>
+        = FjallBenchReadTransaction
+    where
+        Self: 'db;
+
+    fn db_type_name() -> &'static str {
+        "fjall"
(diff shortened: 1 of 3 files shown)


benches/common.rs
```diff
@@ -823,3 +823,148 @@ impl<'db, 'txn> BenchIterator for SanakirjaBenchIterator<'db, 'txn> {
...
+    where
+        Self: 'out;
+
+    fn get<'a>(&'a self, key: &[u8]) -> Option<Self::Output<'a>> {
+        self.txn.get(&self.part, key).unwrap()
+    }
+
+    fn range_from<'a>(&'a self, key: &'a [u8]) -> Self::Iterator<'a> {
+        // let iter = self.table.range(key..).unwrap();
+        // FjallBenchIterator { iter }
```
cberner, line 907: Let's remove this

benches/common.rs
```diff
@@ -823,3 +823,148 @@ impl<'db, 'txn> BenchIterator for SanakirjaBenchIterator<'db, 'txn> {
...
+    where
+        Self: 'txn;
+
+    fn get_inserter(&mut self) -> Self::W<'_> {
+        let FjallBenchWriteTransaction { part, txn } = self;
+        FjallBenchInserter { part, txn }
+    }
+
+    fn commit(self) -> Result<(), ()> {
+        self.txn.commit().map_err(|_| ())
```
cberner, line 951: Does this fsync the data? I'm not familiar with fjall, but it looks like you might need to call `keyspace.persist(PersistMode::SyncAll)?;`

cberner (changes requested):

EmmChriss:
Anything to fix/change? @marvin-j97

cberner (approved):

marvin-j97:
> Anything to fix/change? @marvin-j97

You are not setting the [block cache](https://docs.rs/fjall/latest/fjall/struct.Config.html#method.block_cache), so it will repeatedly decompress read data.

Also you disabled compression (because the feature flag is not set), while RocksDB uses the default compression (snappy). I would probably set both to LZ4.

EmmChriss:
Oh wait let me fix mac build first

PR #961, opened 2025-03-16
Add canopydb to the benchmarks

arthurprs:
Adding Canopydb to the benchmarks. I had this for a while but realized I never opened a PR.

Other additional changes
* Make a lot of values module-level constants
* Use a sync barrier in multi-threaded sections to reduce variance
* Use rng fill functions whenever possible
* Add some basic configuration to Rocksdb. These should be closer to Fjall defaults, for example.
* Instantiate an env_logger instance. This is a noop in practice, but helps development.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,8 @@ sanakirja-core = "=1.4.1"
 rocksdb = { version = "0.22.0", default-features = false, features = ["lz4"] }
 fjall = "2.6"
 comfy-table = "7.0.1"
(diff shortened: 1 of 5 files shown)


cberner:
canopy looks a bit immature, so I don't want to add it now. Feel free to re-open when it's more widely used and mature.

Those other changes sound reasonable though, if you'd like to send just those, I can review them

arthurprs:
That's fair. I'll close this one for now and will follow up with a subset in another PR.

PR #992, opened 2025-06-06
Initial Chrono Support

### LINKED ISSUE #891: Chrono Support
Hey, I store some timestamps inside Redb. It would be nice if Redb could directly support [chrono](https://crates.io/crates/chrono). 

Maybe also directly support the [time](https://crates.io/crates/time) crate while adding chrono support.

wyatt-herkamp:
So this is my initial attempt with Chrono support. 

Please let me know how this can be improved. 

Closes #891


Cargo.toml
```diff
@@ -20,6 +20,7 @@ pyo3-build-config = { version = "0.24.1", optional = true }
 [dependencies]
 log = { version = "0.4.17", optional = true }
 pyo3 = { version = "0.24.1", features=["extension-module", "abi3-py37"], optional = true }
+chrono = { version = "0.4", optional = true }
```
cberner, line 23: Hmm, I'm concerned about the versioning here. What happens when `chrono 0.5` is released, and we want to update this? It seems like changing the version here would require bumping the major version on `redb`.

I'm not familiar with how this is typically handled in the ecosystem. Can you find some examples?

wyatt-herkamp: Honestly, I have zero idea. We could do chrono_v4 as the package name and support both until we are ready to do a major version bump, and at that point, we drop it.

wyatt-herkamp:
Awesome. Half of those were me just trying to prove a concept on how to implement this. But I am heading out for the weekend, so I'll finish this up on Monday.

src/types/chrono.rs
```diff
@@ -0,0 +1,269 @@
...
+        let nanoseconds = value.time().nanosecond().to_le_bytes();
+
+        thirty_two_bit_bytes![year, month, day, time_since_midnight, nanoseconds]
+    }
+
+    fn type_name() -> super::TypeName {
+        TypeName::internal("NaiveDate")
+    }
+}
+impl Value for DateTime<FixedOffset> {
```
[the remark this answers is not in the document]
wyatt-herkamp: No, other time zone implementations are for set time zones. Such as Utc or Local so we can't convert them back on read.

wyatt-herkamp:
I rebased my branch.

cberner (changes requested):
Couple more minor comments, but overall this looks good to me

cberner (approved):
lgtm. Thanks!

PR #1049, opened 2025-07-28
Support WASI preview 2 and all WASI platforms on stable toolchain

metent:
As `FileExt` trait is not getting stabilized [anytime soon](https://github.com/rust-lang/rust/issues/71213#issuecomment-1097758095) on WASI, I added the implementations of `read_exact_at` and `write_all_at` in `redb` itself as these are the only two functions that being used from the trait implementation. WASIp2 currently does not have this trait implemented as all, so with this addition, we get support for WASIp2 and we also eliminate the requirement of nightly toolchain for compiling on WASI platforms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,9 @@ log = { version = "0.4.17", optional = true }
 chrono_v0_4 = { package = "chrono", version= "0.4.41", optional = true }
 uuid = { version= "1.17.0", optional = true }
 
+[target.'cfg(target_os = "wasi")'.dependencies]
+libc = "0.2.174"
+
 # Common test/bench dependencies
 [dev-dependencies]
 rand = "0.9"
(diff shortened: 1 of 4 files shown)


cberner (approved):
Couple more minor comments, but otherwise it looks good to me

cberner:
We can `cfg!` out those backwards compatibility tests, that are failing, on wasi. Since they rely on redb 2.x, I don't see an easy way to fix that

metent:
Added redb 2.6 as a conditional dependency. The tests will, however, still fail as tempfile requires nightly rust for WASI targets. Should I install latest nightly toolchain in the justfile? Or is there a better solution?

cberner:
Oh, that's unfortunate. OK, ya let's keep using nightly for the wasi tests then

tests/backward_compatibility.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(not(target_os = "wasi"))]
```
cberner, line 1: Let's re-enable these, since we have to use nightly anyway

cberner:
I merged the std_flock branch. Can you rebase your PR on master?

PR #1085, opened 2025-09-20
Run cargo check on `wasm32-unknown-unknown` target

alexisfontaine:
Following on from #1084, plus some minor changes.


cberner:
Thanks!

PR #1138, opened 2026-04-16
perf: avoid Vec->Arc copy and other allocation hygiene in page_store

jlucaso1:
## Summary

A handful of small, independent allocation fixes in the page cache and bitmap serialization paths. No behavior change — every output is byte-identical to master — and no API changes. The headline win is that every `PagedCachedFile` cache miss and every `overwrite` write used to allocate the buffer **twice**; this PR makes it one allocation.

### `PagedCachedFile` page buffers — `cached_file.rs`

`vec![0; len].into()` and `read_direct(..).into()` both allocated a `Vec<u8>` first, then allocated a fresh `Arc<[u8]>` and **copied** the bytes into it, because `Arc::<[u8]>::from(Vec<u8>)` cannot reuse the Vec's allocation (the `ArcInner` header must precede the data buffer).

Two helpers now allocate the `Arc` directly via `Arc::new_uninit_slice` (stable since 1.82):

- `zero_filled_arc(len)` — returns a zero-filled `Arc<[u8]>` in one allocation. Used on the `overwrite` path and as the starting buffer for the read path.
- `read_direct_into_arc(offset, len)` — zero-fills via `zero_filled_arc` and then delegates to `StorageBackend::read` (which takes `&mut [u8]`, so the buffer must be initialized before we call it).

The only `unsafe` block is inside `zero_filled_arc` and is a trivial `assume_init()` after filling every slot with `MaybeUninit::new(0)`. Handing out `&mut [u8]` aliasing `MaybeUninit` storage was proposed in an earlier revision of this PR but is UB even when the backend fully overwrites the buffer — it is intentionally avoided.

### `BuddyAllocator::to_vec` — `buddy_allocator.rs`

Each `BtreeBitmap` was serialized twice — once to measure its length (to write the offset header), then again to copy into the output. Serialize once into a `Vec<Vec<u8>>`, use those sizes for the offsets, then `extend_from_slice` them into a pre-sized output.

### `U64GroupedBitmap::{to_vec, from_bytes}` — `bitmap.rs`

Preallocate with the known final size instead of `vec![]` + repeated `extend`, avoiding cascading Vec reallocations on serialize/restore.

### `InMemoryBackend::set_len` — `backends.rs`

Replace `reserve(n) + push(0)` loop with `resize(len, 0)` — an idiomatic single memset instead of a push-per-byte loop.

## Why this is safe

- No format changes, no API changes, no semantic changes.
- All existing tests pass (unmodified). Two small tests for `zero_filled_arc` are added to document the invariants (length preserved, all bytes zero, `len == 0` works).

## Measurements

Same reproduction crate as #1137. Two focused benchmarks:

- `bench_bulk`: 500K inserts / 100K reads / 50K range scans into a single disk-backed DB with a 4 GiB cache.
- `bench_memory`: 32 small databases opened and held live (shown here as a regression check — this PR doesn't target the per-region bookkeeping).

### `bench_bulk` (direct target — every page cache miss/write benefits)

| Metric                         | master       | this PR      | delta    |
| :----------------------------- | -----------: | -----------: | -------: |
| Total heap allocated (DHAT)    | **1,200 MB** | **755 MB**   | **−37%** |
| Total alloc blocks (DHAT)      | 1,648,756    | 1,538,389    | −6.7%    |
| bulk_insert (500K)             | 1,190 ms     | 968 ms       | ~faster  |
| random_reads (100K)            | 137 ms       | 128 ms       | ~neutral |
| range_scans (50K × 10)         | 128 ms       | 109 ms       | ~faster  |
| individual_writes (1K)         | 51 ms        | 37 ms        | ~neutral |
| Max RSS                        | 263,720 KB   | 263,872 KB   | neutral  |
| Peak live heap (DHAT `gmax`)   | 257.6 MB     | 257.6 MB     | neutral  |

**445 MB fewer bytes allocated** across the run. Wall-clock deltas are within run-to-run noise (redb's own benchmark is noisy this way), but the allocation delta is stable and repeatable.

### `bench_memory` (regression check — not this PR's target)

| Metric                          | master       | this PR      |
| :------------------------------ | -----------: | -----------: |
| VmRSS after 32 DBs              | 12,892 KB    | 12,740 KB    |
| Total heap allocated            | 27.2 MB      | 23.9 MB      |

(#1137 is the one that drops the per-region RSS here.)

## Reproducing

Reproduction archive: [redb-measure.zip](https://github.com/user-attachments/files/26796089/redb-measure.zip) — same archive as #1137.

1. Unpack `redb-measure.zip` next to your redb clone (so `../redb` resolves).
2. `cd redb-measure && BRANCHES="master perf/page-store-allocation-hygiene" ./run_all.sh`

The script builds and runs both benchmarks (with and without DHAT) and writes per-run numbers to `./results/`.

### DIFF
--- a/src/tree_store/page_store/backends.rs
+++ b/src/tree_store/page_store/backends.rs
@@ -86,16 +86,7 @@ impl StorageBackend for InMemoryBackend {
     fn set_len(&self, len: u64) -> Result<(), io::Error> {
         let mut guard = self.write();
         let len = usize::try_from(len).map_err(|_| Self::out_of_range())?;
-        if guard.len() < len {
-            let additional = len - guard.len();
-            guard.reserve(additional);
-            for _ in 0..additional {
-                guard.push(0);
-            }
-        } else {
-            guard.truncate(len);
-        }
-
+        guard.resize(len, 0);
         Ok(())
     }
 
--- a/src/tree_store/page_store/bitmap.rs
+++ b/src/tree_store/page_store/bitmap.rs
@@ -258,19 +258,20 @@ impl U64GroupedBitmap {
     // 4 bytes: number of elements
     // n bytes: serialized groups
     pub fn to_vec(&self) -> Vec<u8> {
-        let mut result = vec![];
-        result.extend(self.len.to_le_bytes());
-        for x in &self.data[..Self::required_words(self.len)] {
-            result.extend(x.to_le_bytes());
+        let words = Self::required_words(self.len);
+        let mut result = Vec::with_capacity(size_of::<u32>() + words * size_of::<u64>());
+        result.extend_from_slice(&self.len.to_le_bytes());
+        for x in &self.data[..words] {
+            result.extend_from_slice(&x.to_le_bytes());
         }
         result
     }
 
     pub fn from_bytes(serialized: &[u8]) -> Self {
         assert_eq!(0, (serialized.len() - size_of::<u32>()) % size_of::<u64>());
-        let mut data = vec![];
         let len = u32::from_le_bytes(serialized[..size_of::<u32>()].try_into().unwrap());
         let words = (serialized.len() - size_of::<u32>()) / size_of::<u64>();
+        let mut data = Vec::with_capacity(words);
         for i in 0..words {
             let start = size_of::<u32>() + i * size_of::<u64>();
             let value = u64::from_le_bytes(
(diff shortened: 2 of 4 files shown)


jlucaso1:
@cberner what about creating a CI benchmark that posts results in PRs comparing to master?

Something like:
image

https://github.com/jlucaso1/whatsapp-rust/blob/main/.github/workflows/benchmark-comment.yml

I can set this up in another PR if you like the idea

cberner:
I replaced the unsafe code, and merged the rest. Thanks!

cberner:
I prefer running them manually, at the moment. Github Actions with write permission tokens open up security issues, that I prefer not to think about. I've also found the benchmark is pretty noisy on shared VMs

PR #1169, opened 2026-04-21
Remove checksums from DeletionResult::Subtree variant

cberner:
## Summary
Simplifies the B-tree deletion logic by removing checksums from the `DeletionResult::Subtree` variant, leveraging the deferred checksum system that already handles checksum computation for modified pages.

## Key Changes
- **Removed checksum parameter from `Subtree` variant**: Changed `Subtree(PageNumber, Checksum)` to `Subtree(PageNumber)`, eliminating redundant checksum tracking
- **Updated deletion helper signatures**: Removed `checksum` parameter from `delete_helper`, `delete_leaf_helper`, and `delete_branch_helper` methods
- **Simplified root page handling**: When a deletion doesn't modify the tree (`found.is_none()`), the root page is left untouched to preserve its already-valid checksum
- **Standardized checksum usage**: All modified pages now use `DEFERRED` checksum, which is computed during finalization
- **Refined skip-path optimization**: Updated the child page skip-path check to compare against `DEFERRED` instead of carrying checksums through the variant
- **Improved documentation**: Added clarifying comments explaining why `PartialBranch` and `DeletedBranch` retain checksums (for propagating clean page checksums)

## Implementation Details
The change leverages the existing deferred checksum system where:
- Modified pages automatically get `DEFERRED` checksums
- Clean/unchanged pages retain their existing checksums
- The finalization phase recomputes checksums only for uncommitted pages

This eliminates the need to track checksums through the deletion result enum, simplifying the code while maintaining correctness through the deferred checksum mechanism.

https://claude.ai/code/session_01GutSf5gN9dEvRyqxCY2x1o

### DIFF
--- a/src/tree_store/btree_mutator.rs
+++ b/src/tree_store/btree_mutator.rs
@@ -27,11 +27,10 @@ enum DeleteTarget<'a> {
     Last,
 }
 
-// TODO: it seems like Checksum can be removed from most/all of these, now that we're using deferred checksums
 #[derive(Debug)]
 enum DeletionResult {
     // A proper subtree
-    Subtree(PageNumber, Checksum),
+    Subtree(PageNumber),
     // A leaf with zero children
     DeletedLeaf,
     // A leaf with fewer entries than desired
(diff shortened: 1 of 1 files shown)


cberner:
@codex review

PR #1181, opened 2026-04-22
Refactor database header recovery to handle file length changes

cberner:
## Summary
Refactored the database header recovery logic to properly detect and handle cases where the backing file has been truncated or extended externally. The recovery check now always considers both the on-disk recovery flag and whether the stored layout matches the actual file length.

## Key Changes
- **Enhanced `recovery_required()` method**: Now takes the actual file length as a parameter and returns true if either the recovery flag is set OR the stored layout doesn't match the file length
- **Refactored `finalize()` method**: 
  - Now takes `file_len` as a parameter to enable layout reconciliation
  - Automatically recalculates the database layout if it's stale (doesn't match actual file length)
  - Returns a `clean` flag that is true only when no reconciliation was needed (primary was kept and layout already matched)
  - Extracted slot selection logic into a separate `select_primary_slot()` helper method
- **Simplified caller logic in `page_manager.rs`**:
  - Removed manual layout recalculation code from `open()` - now handled by `finalize()`
  - Consolidated recovery detection into a single `recovery_required()` call
  - Improved `check_integrity()` to properly use the `clean` flag returned by `finalize()`
  - Eliminated redundant TODO comments and simplified the repair decision logic

## Implementation Details
- The `finalize()` method now owns the responsibility for both layout reconciliation and slot selection, ensuring these operations are always performed together
- File length is passed through the call chain to ensure consistent state checking
- The `clean` flag accurately reflects whether any repair work was performed, enabling proper write-back of the header when needed

https://claude.ai/code/session_01AozxQQYZQZ7ackWThrURfN

### DIFF
--- a/src/tree_store/page_store/header.rs
+++ b/src/tree_store/page_store/header.rs
@@ -151,23 +151,48 @@ impl UnrepairedDatabaseHeader {
         self.inner.page_size
     }
 
-    pub(super) fn layout(&self) -> DatabaseLayout {
-        self.inner.layout()
+    // Returns true if the header needs to be repaired before use: either the recovery_required
+    // flag is set on disk, or the stored layout no longer matches the current file length (e.g.
+    // the file was truncated or extended externally). Callers must pass the actual file length
+    // so both conditions are always checked together.
+    pub(super) fn recovery_required(&self, file_len: u64) -> bool {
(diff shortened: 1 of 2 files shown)


cberner:
@codex review

PR #1194, opened 2026-04-27
Avoid creating system tables during metadata checks

cberner:
Read system table roots directly for metadata-only paths instead of opening them with get_or_create_table. This keeps savepoint and pending-free checks from dirtying otherwise empty transactions.

Also share the DATA_FREED_TABLE and SYSTEM_FREED_TABLE extraction logic so the freeing path is ready for the follow-up page reuse work without duplicating the table scan.


cberner:
@codex review

Issue #19, opened 2021-07-07
Allow database to be opened by multiple processes/threads


cberner:

yaofan:
Hi, is there any processing?

PR #1195, opened 2026-04-27
redb-python: add Database.create() binding

cberner:
Exposes redb::Database::create as a Python classmethod so callers can
create (or open) a database file from Python. Adds a pytest sanity test
plus a hypothesis test that fuzzes the filename.

Refs: https://github.com/cberner/redb/issues/19

Assisted-by: Claude <noreply@anthropic.com>


cberner:
@codex review

PR #1226, opened 2026-05-04
Add B-tree balance validation to fuzzing tests

cberner:
## Summary
This PR adds validation to the redb fuzzing tests to ensure that B-trees maintain proper balance properties. The changes introduce helper functions to calculate expected maximum tree heights based on the number of entries and page size, then assert that actual tree heights don't exceed these bounds.

## Key Changes
- Added `page_size` parameter to `apply_crashable_transaction()` and `apply_crashable_transaction_multimap()` functions to enable balance validation
- Implemented `expected_max_height()` function that calculates the theoretical maximum height of a B-tree based on:
  - Number of entries
  - Page size and the merge threshold (page_size/3) which determines minimum branch width
  - A fanout calculation: ceil(page_size / 96.0) representing bytes per child (16B checksum + 8B page number + 8B key)
- Added `assert_balanced()` to validate single-table B-tree heights with a +2 slack factor for root level and multi-page values
- Added `assert_balanced_multimap()` to validate multimap B-tree heights by separately bounding outer tree (keyed by u64) and inner subtrees (per-key value collections)
- Integrated balance assertions into transaction execution to check tree stats after each operation batch
- Updated all call sites to pass the `page_size` parameter through the transaction execution chain

## Implementation Details
- The balance validation uses logarithmic calculations to determine expected fanout and maximum height
- Multimap validation accounts for the two-level structure: outer tree for keys and inner subtrees for values per key
- The +2 slack in single-table validation accommodates the root node (which can be shallower) and leaves with single multi-page values
- Assertions are performed after table operations but before transaction commit, catching imbalance issues early in fuzzing

https://claude.ai/code/session_018M8GfZ5TUmdL6xdAyZm6oP

### DIFF
--- a/fuzz/fuzz_targets/fuzz_redb.rs
+++ b/fuzz/fuzz_targets/fuzz_redb.rs
@@ -660,6 +660,7 @@ fn exec_table_crash_support<T: Clone + Debug>(
         &mut BTreeMap<u64, T>,
         &FuzzTransaction,
         &mut SavepointManager<T>,
+        usize,
     ) -> Result<(), redb::Error>,
 ) -> Result<(), redb::Error> {
     let mut redb_file: NamedTempFile = NamedTempFile::new().unwrap();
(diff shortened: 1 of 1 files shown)


cberner:
@codex review

PR #1228, opened 2026-05-05
Prepare range iterator branch state for visitors

cberner:
Represent range iterator branch positions with an explicit traversal state and cache leaf entry counts. This keeps the existing iterator API unchanged while making the branch state easier to extend with traversal visitor events in the extract_if optimization.

Assisted-by: Codex


cberner:
@codex review

Issue #826, opened 2024-07-22
consider reducing log level to trace/debug


flokli:
We're using redb in an application, and with the default `RUST_LOG` (`info`) there's a lot of log messages from `redb` popping up, at `info` level.

Opening and closing transactions is quite normal, I'd not expect `redb` to emit generic debug logs at `info` level during normal operation, but it'd help to have it available for instrumentation usecases (so I don't want to feature-flag it out).

Would you be up to a PR moving these to `debug` or `trace` level?

flokli:
I'm talking about the log messages that are currently at info (emitted when doing transactions).

Usually applications use `info` as a log level default, and this means downstream applications get log lines about redb transactions happening.

Changing these to debug/trace would cause them only to appear if the log level is increased.

cberner:
You can just filter them out. `RUST_LOG=redb=warn` is the syntax I think

flokli:
Yes, but it need to be set explicitly by users of the application, which is confusing, especially if used transitively. I wouldnt expect a library to log at info level.

cberner:
You should be able to set it via the logging APIs, I think, then users won't need to. I don't know the exact API call off hand though.

> I wouldnt expect a library to log at info level.
Is there any official guidance you know of? Happy to look at that, but other libraries I've used also log at `info!`

flokli:
I cannot, if I'm providing a library itself, that users consume.

It'd be on them to set up logging, and explicitly filter out log messages that come from redb.

Seems there's no official guidelines for this, but https://github.com/rust-lang/api-guidelines/discussions/65#discussioncomment-233350  dives a bit into this topic, that logging with debug and trace only for regular operations is preferred.

> You should be able to set it via the logging APIs, I think, then users won't need to. I don't know the exact API call off hand though.
> 
> > I wouldnt expect a library to log at info level.
> > Is there any official guidance you know of? Happy to look at that, but other libraries I've used also log at `info!`

cberner:
That link doesn't seem very authoritative. Do you have any examples of high profile libraries that do it this way? I looked at `serde`, but I don't see any logging calls in it

flokli:
I could mention tonic/tower. They also don't log an info on every new connection being made / request being processed. But move this to trace. I'd expect redb to behave the same.

cberner:
I left one at `info!`, but changed the rest. Thanks for the references!

flokli:
Thanks for changing this!

flokli:
[...]

PR #1262, opened 2026-05-22
Downgrade remaining log messages from info! to debug!

flokli:
Not logging at info level for noncritical messages seems common practice for libary code, and previously the amount of log lines logging at this level was already reduced to one single occasion (#828), after the report in #826.

Since then, two more occurences have popped up, causing two log lines for each redb database opened. Switch them to debug!.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -25,7 +25,7 @@ use crate::transactions::{
 };
 use crate::tree_store::file_backend::FileBackend;
 #[cfg(feature = "logging")]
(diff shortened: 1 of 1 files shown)


flokli:
See https://github.com/cloud-hypervisor/fuse-backend-rs/pull/225 / https://github.com/cloud-hypervisor/fuse-backend-rs/pull/195 for another example of other library code downgrading these to debug level.

flokli:
Thanks!

PR #1268, opened 2026-06-11
Coalesce sparse leaf rewrites into packed replacement runs

cberner:
Ports the `codex/retain-owned-leaf-builder-prototype` run-coalescing onto the merged cursor machinery, generalized per review to both scan directions and to extract_if.

When removals leave a leaf below the merge threshold, the cursor used to merge it into a sibling one mutator call at a time, re-copying the accumulated survivors as the scan advanced across the parent. Now each consumed leaf's retained entries are buffered in owned memory and packed replacement leaves are spliced into the parent in a single update, freeing the originals.

Beyond the prototype:
- Runs work in both scan directions: buffered entries are kept in key order, so backward scans still emit leaves in key order.
- Runs apply to extract_if's deferred batches too: the buffers hold no page references, so they are safe with outstanding guards, and a range cursor splices its open run when the other end activates (park/close are the existing mutation boundaries).
- Only leaves left below the merge threshold join a run; the first leaf that stays healthy on its own ends it (its pending removals ride the splice). Runs never consume the dense tail, and the buffer is structurally bounded at roughly fanout x page_size / 3 plus one final leaf — no explicit byte cap needed (previously up to fanout x 1MiB transient heap on large-value tables).
- The cursor owns only run bookkeeping (which children, which direction, when to splice); all rewrite policy lives in the mutator's `replace_leaf_children`, next to `plan_leaf_delete`: entries that pack below the merge threshold absorb the same adjacent child `plan_leaf_delete` would merge with (skipping single-large-value neighbors), and the trailing two replacement leaves are divided with the same `build_split` used by ordinary inserts, preserving the same minimum-fill bound as single-key deletes (previously repeated bounded retains could decay leaves toward one entry each).
- Explicit poisoning: an error that interrupts removals already reported to the caller discards the cursor state, every later operation re-raises `StorageError::PreviousIo` (the storage layer has latched the underlying failure), and the transaction is poisoned. An extract iterator latches the first error and later calls keep re-raising it instead of continuing, so iteration cannot re-test entries through a stale range bound.

Benchmarks (1M entries, 24B/150B, committed data, interleaved with master): retain keep-half ~1.9x, retain keep-1-in-8 ~2.1x, and extract_from_if removing every other entry over a third of the keyspace improves a further ~1.8x on top of the cursor rework — matching the throughput of the eager whole-range-rebuild prototype while keeping lazy semantics and memory structurally bounded.

Verified: `just test` green (302 tests, including new regression tests for the sparse-boundary merge, descending-removal decay, and large-value runs), clean fuzz runs (including interleaved-direction extract ops that drive backward runs and park-time splices), and adversarial scenario tests (interior-band retain, alternating double-ended extract with held guards, oversized values).

Known follow-up: replacement leaves other than the balanced tail are packed completely full, which shifts cost to later inserts.

https://claude.ai/code/session_0112BvsuWfD2hRL6VNiLzyyt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,12 @@
 ## 4.2.0 - 2026-XX-XX
 * Fix a crash during a transaction that grows the database file leaving the database permanently
   unopenable afterward.
-* Optimize `Table::retain()` and `Table::retain_in()`. Some benchmarks on large tables show a 25x speedup.
+* Optimize `Table::retain()` and `Table::retain_in()`. Some benchmarks on large tables show a 25-50x speedup.
 * Optimize `Table::extract_if()` and `Table::extract_from_if()`. Benchmarks on large tables show
-  a 1.6-4x speedup, depending on the fraction of entries extracted, and iterating the extract
+  a 2-4x speedup, depending on the fraction of entries extracted, and iterating the extract
   iterator from both ends no longer degrades the removal batching.
+* After the extract iterator returns an error, later calls keep returning an error instead of
+  continuing.
 * Add `Table::entry()` and the associated `Entry`, `OccupiedEntry`, and `VacantEntry`
   types, mirroring `std::collections::BTreeMap::entry`. Supports `or_insert`,
   `or_insert_with`, `or_insert_with_key`, `and_modify`, and the usual `OccupiedEntry`
--- a/src/table.rs
+++ b/src/table.rs
@@ -148,6 +148,11 @@ impl<'txn, K: Key + 'static, V: Value + 'static> Table<'txn, K, V> {
     ///
     /// Note: values not read from the iterator will not be removed
     ///
+    /// If the iterator returns an error, later calls keep returning an error
+    /// (the failure is not recoverable). Entries already yielded stay
+    /// removed; if finalizing their removal fails too, the write transaction
+    /// is poisoned and cannot be committed.
+    ///
     /// The predicate must not panic. If it panics, the write transaction is
     /// poisoned and [`crate::WriteTransaction::commit`] will return
     /// [`crate::CommitError::TransactionPoisoned`].
@@ -162,6 +167,11 @@ impl<'txn, K: Key + 'static, V: Value + 'static> Table<'txn, K, V> {
     /// `predicate` evaluates to `true` are returned in an iterator, and those which are read from the iterator are removed
     ///
     /// Note: values not read from the iterator will not be removed
+    ///
+    /// If the iterator returns an error, later calls keep returning an error
+    /// (the failure is not recoverable). Entries already yielded stay
+    /// removed; if finalizing their removal fails too, the write transaction
+    /// is poisoned and cannot be committed.
     ///
     /// The predicate must not panic. If it panics, the write transaction is
     /// poisoned and [`crate::WriteTransaction::commit`] will return
@@ -190,12 +200,12 @@ impl<'txn, K: Key + 'static, V: Value + 'static> Table<'txn, K, V> {
         predicate: F,
     ) -> Result {
         let mut panic_guard = RetainPanicGuard::new(self.transaction);
-        let mut lost_removals = false;
+        let mut poisoned = false;
         let result = self
             .tree
-            .retain_in::<K::SelfType<'_>, F>(predicate, .., &mut lost_removals);
+            .retain_in::<K::SelfType<'_>, F>(predicate, .., &mut poisoned);
         panic_guard.disarm();
-        if lost_removals {
+        if poisoned {
             self.transaction.poison();
         }
         result
@@ -217,10 +227,10 @@ impl<'txn, K: Key + 'static, V: Value + 'static> Table<'txn, K, V> {
         KR: Borrow<K::SelfType<'a>> + 'a,
     {
         let mut panic_guard = RetainPanicGuard::new(self.transaction);
-        let mut lost_removals = false;
-        let result = self.tree.retain_in(predicate, range, &mut lost_removals);
+        let mut poisoned = false;
+        let result = self.tree.retain_in(predicate, range, &mut poisoned);
         panic_guard.disarm();
-        if lost_removals {
+        if poisoned {
             self.transaction.poison();
         }
         result
@@ -660,7 +670,8 @@ impl<
     /// Entries already returned by the iterator remain removed, and unread
     /// entries are not tested by the predicate or removed. Dropping the iterator
     /// also closes it, but this method returns any error encountered while
-    /// finalizing the iterator.
+    /// finalizing the iterator, including when the iterator already closed
+    /// itself after an iteration error.
     pub fn close(mut self) -> Result {
         self.inner.close()
     }
@@ -675,9 +686,11 @@ impl<
     fn drop(&mut self) {
         // Entries already yielded may have removals pending; if a flush
         // failed the table would silently keep them, so poison the
-        // transaction instead of letting it commit.
-        let close_failed = self.inner.close().is_err() || self.inner.close_failed();
-        if (close_failed || self.inner.predicate_panicked())
+        // transaction instead of letting it commit. An iteration error
+        // alone must not poison: close_failed reports lost removals, and
+        // close() also re-raises after a cleanly finalized error.
+        let _ = self.inner.close();
+        if (self.inner.close_failed() || self.inner.predicate_panicked())
             && let Some(transaction) = self.poison_target
         {
             transaction.poison();
--- a/src/transactions.rs
+++ b/src/transactions.rs
@@ -1285,9 +1285,25 @@ impl WriteTransaction {
             };
             let mut system_tables = self.system_tables.lock().unwrap();
             let mut data_freed = system_tables.open_system_table(self, DATA_FREED_TABLE)?;
-            for entry in data_freed.extract_from_if(lower.., |_, _| true)? {
-                entry?;
+            let drain = || -> Result<(), StorageError> {
+                let mut iter = data_freed.extract_from_if(lower.., |_, _| true)?;
+                for entry in &mut iter {
+                    entry?;
+                }
+                // Defensive: exhaustion has already closed the iterator and
+                // surfaced any finalization error through the loop; this only
+                // keeps errors from being swallowed if the loop ever gains an
+                // early exit.
+                iter.close()
+            };
+            let result = drain();
+            if result.is_err() {
+                // The table tree root was already restored above, so a partial
+                // purge must not be committed. System-table extract iterators
+                // carry no poison target, so poison the transaction directly.
+                self.poison();
             }
+            result?;
             // No need to process the system freed table, because it only rolls forward
         }
 
--- a/src/tree_store/btree_cursor.rs
+++ b/src/tree_store/btree_cursor.rs
@@ -1,11 +1,14 @@
 use crate::AccessGuard;
-use crate::Result;
-use crate::tree_store::btree_base::{BRANCH, BranchAccessor, LEAF, LeafAccessor};
+use crate::tree_store::btree_base::{
+    BRANCH, BranchAccessor, LEAF, LeafAccessor, OwnedEntryBuffer, leaf_below_merge_threshold,
+    retained_after_removals,
+};
 use crate::tree_store::btree_iters::EntryGuard;
 use crate::tree_store::btree_mutator::MutateHelper;
 use crate::tree_store::page_store::{Page, PageHint, PageImpl};
 use crate::tree_store::{BtreeHeader, PageAllocator, PageNumber, PageResolver, PageTrackerPolicy};
 use crate::types::{Key, Value};
+use crate::{Result, StorageError};
 use std::cmp::Ordering;
 use std::collections::Bound;
 use std::collections::Bound::{Excluded, Included, Unbounded};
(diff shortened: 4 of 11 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 4.2.0 - 2026-XX-XX
 * Optimize `Table::retain()` and `Table::retain_in()`. Some benchmarks on large tables show a 25x speedup.
+* Optimize `Table::retain()` and `Table::retain_in()` further for sparse retains: runs of underfull
+  leaves are now coalesced into packed replacement leaves in a single parent update, roughly
+  doubling sparse retain throughput.
```
cberner: Edit the previous row which is about this same function. No implementation details. Just the user facing impact

src/tree_store/btree_cursor.rs
```diff
@@ -543,21 +652,41 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> CursorMut<'a, 'b, K, V> {
             if position.has_entry(direction) {
                 return Ok(true);
             }
+            if matches!(direction, Direction::Next) {
```
cberner: It seems suspect that this only handles one direction

src/tree_store/btree_cursor.rs
```diff
@@ -543,21 +652,41 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> CursorMut<'a, 'b, K, V> {
...
             if !self.state.removed_indexes.is_empty() {
                 let resume_key = self
                     .flush_removed_entries(direction)?
                     .expect("pending removals exist");
-                let resume = match direction {
-                    Direction::Next => Position::After(resume_key.as_slice()),
-                    Direction::Previous => Position::Before(resume_key.as_slice()),
-                };
-                self.seek_to(resume)?;
+                self.seek_to(Position::Before(resume_key.as_slice()))?;
```
cberner: Hmm, why does this only handle one direction now?

CHANGELOG.md
cberner: Done — folded into the existing retain row as "25-50x speedup" with the implementation details dropped.

---

src/tree_store/btree_cursor.rs
cberner: Agreed, the forked control flow obscured the actual constraint. Restructured: both directions now go through the same close_current_leaf(direction)/advance_past_closed_leaf(direction) flow, and the forward-only rule is one explicit `run_eligible` condition with a comment. The underlying reason runs are forward-only: the only guard-free batching caller (retain) scans forward, and backward scans exist solely for extract's deferred batches, which carry guards and always rewrite directly — a debug_assert now pins that a run can never coexist with a backward close.

---

src/tree_store/btree_cursor.rs
cberner: Fixed along with the thread above — this arm no longer exists; ensure_has_entry is direction-uniform again and the run gating lives in close_current_leaf behind a single commented condition.

---

src/tree_store/btree_cursor.rs
```diff
@@ -226,6 +229,155 @@ struct Leaf {
...
+        };
+        for buffer in buffers {
+            for (key, value) in buffer.entries() {
+                if builder.would_split_with(key, value) {
+                    flush(&mut builder, leaves)?;
+                }
+                builder.push(key, value);
+            }
+        }
+        flush(&mut builder, leaves)
```
cberner: This doesn't seem right. We don't need an OwnedLeafBuilder here. It looks like could use the normal unowned one.

src/tree_store/btree_cursor.rs
```diff
@@ -226,6 +229,155 @@ struct Leaf {
...
+// foreign mutations; only the parent path captured at splice time must be
+// current. Runs grow in either scan direction.
+struct LeafRunRewrite {
+    parent_page: PageNumber,
+    direction: Direction,
+    // The parent's children consumed so far: a contiguous range extended
+    // upward by forward scans and downward by backward scans.
+    replaced_children: Range<usize>,
+    // One buffer per consumed leaf, in scan order.
+    leaf_buffers: Vec<OwnedEntryBuffer>,
```
cberner: Why are we buffering into a Vec of buffers? We should be buffering into an OwnedLeafBuilder I think

cberner: Done in 63d29af. I kept the `Vec<OwnedEntryBuffer>` (backward scans consume leaves in descending order but replacement leaves must be built ascending, so they're reordered at build time), added `would_split_with()` to `LeafBuilder`, and dropped `OwnedLeafBuilder` entirely — the build step now packs with the normal `LeafBuilder` borrowing directly from the buffers. Net −96 lines, and it removes a copy since `LeafBuilder` references the buffered bytes instead of `OwnedLeafBuilder` re-owning them. (`OwnedLeafBuilder` existed only because `LeafBuilder` couldn't emit a stream of full leaves; `would_split_with()` closes that gap.)

---

cberner: Updated in c181473 per your VecDeque suggestion: collapsed the `Vec<OwnedEntryBuffer>` into a single `OwnedEntryBuffer` whose `pairs` is a `VecDeque`. Forward scans `push_back`, backward scans `push_front` (iterating the leaf in reverse so its entries stay ascending at the front), so the buffer is always in final ascending order and the build step is now a flat single loop with no reversal. Benchmarked min-of-N (noisy box): retain8 285ms vs 310ms, retain2 726ms vs 790ms — no regression, marginally faster and more consistent. All tests + the both-direction stress pass; 45k fuzz runs clean.

---

cberner:
Addressed both open review comments in e5e500e (replying here because a pending review on the PR blocks in-thread replies):

- **"Is this clean unwinding code necessary?"** (build_replacement_leaves) — agreed, removed. Any error there poisons the cursor and the transaction, and rollback already reclaims every page allocated since the last commit, so the manual unwinding was redundant. The matching `free_pages` cleanup after a failed splice was redundant for the same reason and is deleted too.
- **"This should have a debug_assert! that the page was uncommitted"** — the `free_if_uncommitted` loop in question was part of that unwinding block and no longer exists.

e5e500e also folds in fixes from a deeper review pass: the run buffer is now capped at roughly one parent's worth of packed pages (fixes a fanout x value-size transient-memory blowup on large-value tables: 1.76s / +192MiB → sub-ms on a 200 x 1MiB retain), splices that would strand an under-threshold leaf absorb an adjacent sibling and rebalance the trailing two replacement leaves (restores the single-delete minimum-fill bound; a descending single-key retain pattern previously decayed to 31 leaf pages where master used 1), the run invariant debug_asserts are promoted to hard asserts, extract iterators fuse on the first error, and the split/merge-threshold policies are consolidated into shared helpers so the cursor and mutator cannot drift. New regression tests cover the sparse-boundary merge, descending-removal decay, and large-value cap.

---

src/tree_store/btree_cursor.rs
```diff
@@ -219,13 +228,230 @@ fn key_data<K: Key + 'static, V: Value + 'static>(leaf: &Leaf, position: usize)
...
+                let tail_healthy = !leaf_below_merge_threshold(
+                    len - division,
+                    second_bytes,
+                    fixed_key,
+                    fixed_value,
+                    page_size,
+                );
+                if first_fits && tail_healthy {
+                    plan.push((combined.start..combined.start + division, first_bytes));
+                    plan.push((combined.start + division..combined.end, second_bytes));
```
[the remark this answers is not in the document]
cberner: Done in 706c84c — added a `second_fits` check symmetric with `first_fits`, so the rebalanced boundaries are adopted only when both halves satisfy the split policy and the tail clears the merge threshold. I could not construct a violating case within the offset-arithmetic bounds of a valid greedy plan, but the extra condition uses values already computed and makes the invariant local rather than resting on that argument.

---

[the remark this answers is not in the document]
cberner: Superseded in 9d67efa: the hand-rolled midpoint rebalance (division walk, clamp, first_fits/second_fits/tail_healthy) is gone. When the greedy plan strands a sparse trailing page, the last two ranges are now rebuilt with the same `LeafBuilder::build_split()` used by ordinary inserts, which already handles the byte-midpoint division and u16::MAX pair clamping and allocates whatever page order each half needs — so the split policy can't be bypassed by construction.

---

src/tree_store/btree_cursor.rs
```diff
@@ -219,13 +228,186 @@ fn key_data<K: Key + 'static, V: Value + 'static>(leaf: &Leaf, position: usize)
...
+                &page_allocator,
+                &allocated,
+                range.len(),
+                fixed_key,
+                fixed_value,
+            );
+            for (key, value) in &entries[range.clone()] {
+                builder.push(key, value);
+            }
+            let (page1, split_key, page2) = builder.build_split()?;
```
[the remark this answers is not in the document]
cberner: This is the intended behavior, not a bypass: `build_split()` here is deliberately unguarded because it is the same canonical division the rest of the tree uses on identically-shaped input. The existing single-key delete path (btree_mutator.rs, `apply_child_deletion_result`) merges a below-merge-threshold leaf with a sibling that can be nearly full and calls `child_builder.build_split()` with no fit checks either — same envelope as here (one greedy-packed range that fits a page plus one below-threshold tail, since greedy would not have cut a combinable range). If a byte-midpoint half ever needs more than the nominal page size, `build_split` allocates a higher-order page, which is first-class in redb (the same thing happens today on that merge path and on insert overflow). Adding a fit guard here would make the run path stricter than the policy it is meant to match, and rejecting the split would just re-strand the sparse tail the rebalance exists to fix.

---

src/table.rs
```diff
@@ -148,6 +148,9 @@ impl<'txn, K: Key + 'static, V: Value + 'static> Table<'txn, K, V> {
     ///
     /// Note: values not read from the iterator will not be removed
     ///
+    /// If the iterator returns an error, it closes itself: subsequent calls
+    /// return `None`, and entries already yielded stay removed.
+    ///
```
cberner: This behavior is wrong. If the iterator returns an error, it should continue returning an error (unless the error is recoverable and the code implements that recovery)

src/table.rs
```diff
@@ -163,6 +166,9 @@ impl<'txn, K: Key + 'static, V: Value + 'static> Table<'txn, K, V> {
...
```
cberner: Same comment

cberner (commented):
The cursor splicing / leaf closing still has pretty complex heuristics. Possible to simplify it?

src/table.rs
cberner: Fixed in 2fba8db. The iterator now latches the error: later `next()`/`next_back()` calls return `Err(StorageError::PreviousIo)` instead of `None`. These errors are not recoverable — every error reachable on this path has latched the storage layer's io_failed flag, which already blocks commit — so re-raising is the honest contract. The docs here and the CHANGELOG entry are updated to match.

---

src/table.rs
cberner: Fixed together with the thread above (2fba8db) — both extract_if and extract_from_if share the same iterator, and its docs now describe the re-raising behavior.

---

cberner:
Re "the cursor splicing / leaf closing still has pretty complex heuristics" — simplified in 2fba8db by moving all of the rewrite policy out of the cursor and into the mutator, next to the code it must agree with:

- `replace_leaf_children` now takes the raw entry buffer and owns everything downstream: packing into full leaves, the balanced tail via `build_split`, and the sparse-result merge (it absorbs the same adjacent child `plan_leaf_delete` would merge with, skipping single large values). The cursor's `absorb_adjacent_sibling`, `build_replacement_leaves`, and the splice-time threshold gate are all deleted — `splice_run` is now just "drop the leaf ref, assert the parent matches, hand the buffer to the mutator".
- `close_current_leaf` is a single decision ladder with one splice exit: a leaf that would underfill joins the run (and the run splices when its parent runs out of children); a healthy leaf ends the run, its pending removals riding the splice; otherwise it's the ordinary flush/no-op.

So the cursor now contains only run bookkeeping (which contiguous children, which direction, when the run ends) and the heuristics — merge threshold, sibling choice, split policy — each exist in exactly one place, in the mutator, shared with the single-key delete path. Cursor is ~155 lines lighter; all tests, the adversarial scenarios, and fuzzing pass with retain/extract benchmarks unchanged.

---

CHANGELOG.md
```diff
@@ -3,10 +3,11 @@
...
 * Fix a crash during a transaction that grows the database file leaving the database permanently
   unopenable afterward.
-* Optimize `Table::retain()` and `Table::retain_in()`. Some benchmarks on large tables show a 25x speedup.
+* Optimize `Table::retain()` and `Table::retain_in()`. Some benchmarks on large tables show a 25-50x speedup.
 * Optimize `Table::extract_if()` and `Table::extract_from_if()`. Benchmarks on large tables show
-  a 1.6-4x speedup, depending on the fraction of entries extracted, and iterating the extract
-  iterator from both ends no longer degrades the removal batching.
+  a 2-4x speedup, depending on the fraction of entries extracted, and iterating the extract
+  iterator from both ends no longer degrades the removal batching. After the extract iterator
+  returns an error, later calls keep returning an error instead of continuing.
```
cberner: `After the extract iterator returns an error, later calls keep returning an error instead of continuing` should be in a separate bullet point

cberner: Done in 31dbaf8.

---

src/transactions.rs
```diff
@@ -1285,9 +1285,23 @@ impl WriteTransaction {
             };
             let mut system_tables = self.system_tables.lock().unwrap();
             let mut data_freed = system_tables.open_system_table(self, DATA_FREED_TABLE)?;
-            for entry in data_freed.extract_from_if(lower.., |_, _| true)? {
-                entry?;
+            let drain = || -> Result<(), StorageError> {
```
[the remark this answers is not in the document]
cberner: This compiles: the closure moves `lower` (a non-Copy `TransactionIdWithPagination`) into the `lower..` range expression, so it implements only `FnOnce`, and calling a once-used `FnOnce` closure through an immutable binding is by-value and legal. E0596 would only apply if the closure were `FnMut` (nothing moved). Verified empirically: the crate builds with `--deny warnings` and CI's coverage job has compiled this exact code on the last several pushes.

---

PR #1269, opened 2026-06-11
Fix check_integrity() falsely reporting corruption on a healthy database after commits that freed pages

cberner:
## Problem

`Database::check_integrity()` on a live database handle deterministically returned `Ok(false)` ("failed integrity checks but was repaired") on a perfectly healthy database after any committed transaction that freed data pages — e.g. committing a multi-level table and then removing a single key. It also performed a spurious repair commit. Reopening the same file and re-running the check returned `true`, proving the persisted state was consistent.

## Root cause

Durable commits run a post-commit epilogue (`process_data_freed_pages_after_commit`) that frees the pages pending in `DATA_FREED_TABLE` in the live in-memory allocator and removes the entries via a non-durable commit, so the live state is routinely ahead of the durable state. `check_integrity()` captured `allocator_hash` from this live post-epilogue state, but `clear_cache_and_reload()` reverts to the last durable state — where the freed-table entries still exist and the listed pages still count as allocated — so the repair-rebuilt allocator legitimately differed from the captured hash.

This is an unreleased regression from the 4.2.0 "reuse pages freed by a durable write transaction one transaction sooner" work, so no changelog entry is included.

## Fix

Compare like with like:

1. Validate the live allocator against an allocator rebuilt from the **live** roots (new `rebuild_allocator_state()` helper, factored out of `do_repair()`; it performs no file writes, so externally-injected corruption is not disturbed before the reload). This preserves leak/double-free detection, including state loaded from a stale allocator state table.
2. Verify the durable state exactly as before: `clear_cache_and_reload()` (header/layout reconciliation), `do_repair()` (checksum verification, primary-slot repair), and the root comparison. Unclean shutdowns and external modifications are still reported as `false`-and-repaired.

No double-free can result from reverting the live state to the durable roots: the still-registered pending non-durable commit pins freed-page processing at its durable ancestor until the next durable commit.

## Testing

* New regression test `check_integrity_clean_after_delete` (fails before the fix, passes after), asserting that a healthy live handle reports clean after a freeing commit, including on repeated calls.
* `just test` passes: cargo deny, fmt, clippy (`-D warnings`), and the full test suite, including the existing repair, savepoint, and integrity tests.

---

Found by a codebase audit; the original failing repro is on the `claude/codebase-bug-audit-h638r8` branch.

https://claude.ai/code/session_01GGotvNGnZVJjsf5stQzS8o

---
_Generated by [Claude Code](https://claude.ai/code/session_01GGotvNGnZVJjsf5stQzS8o)_

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -579,9 +579,34 @@ impl Database {
     /// Returns [`DatabaseError::TransactionInProgress`] if any read or write transaction is still
     /// alive when this method is called.
     pub fn check_integrity(&mut self) -> Result<bool, DatabaseError> {
-        let allocator_hash = self.mem.allocator_hash();
-        let mem = Arc::get_mut(&mut self.mem).ok_or(DatabaseError::TransactionInProgress)?;
-        let mut was_clean = mem.clear_cache_and_reload()?;
+        if Arc::get_mut(&mut self.mem).is_none() {
+            return Err(DatabaseError::TransactionInProgress);
+        }
+
+        // The live in-memory state may be ahead of the durable state, because durable commits
+        // process pending freed pages after committing and publish the result as a non-durable
+        // commit. The live allocator state corresponds to the live roots, so it must be checked
+        // against a rebuild from the live roots: comparing it against the rebuild from the
+        // durable roots (after the reload below) would falsely report corruption.
+        let mut was_clean = match Self::verify_primary_checksums(self.mem.clone()) {
+            Ok(true) => {
+                let live_allocator_hash = self.mem.allocator_hash();
+                match Self::rebuild_allocator_state(&mut self.mem, &|_| {}) {
+                    Ok(_) => live_allocator_hash == self.mem.allocator_hash(),
+                    Err(DatabaseError::Storage(StorageError::Corrupted(_))) => false,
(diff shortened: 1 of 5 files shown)


src/db.rs
```diff
@@ -907,12 +950,6 @@ impl Database {
             Self::check_repaired_allocated_pages_table(system_root, mem.clone())?;
         }
 
-        mem.end_repair()?;
```
cberner, line 910: It feels off that this function calls begin_repair() but no longer calls end_repair() and leaves that to the caller. begin_repair() may get called twice now I think, which maybe is ok, but makes the function name seem bad.

cberner: Agreed — the begin/end pairing no longer matched what the functions do. Reworked in the latest push: `begin_repair()` is now `reset_allocator_state()` (it just swaps in fresh empty allocators, and `rebuild_allocator_state()` owns it entirely, so running it twice per `check_integrity()` is self-evidently fine), and `end_repair()` is now `clear_recovery_required()`, which stays the repair path's responsibility since it's about the header flag rather than the allocator rebuild. No behavior change; full suite still green.

---

PR #1288, opened 2026-07-04
Extend leaf runs across every rewritten leaf under a parent

cberner:
A replacement run used to end at the first leaf that stayed above the merge threshold, so removal patterns that shrink every leaf without underfilling it (dropping every other entry, for example) rewrote each leaf 1:1 into a half-empty replacement: no packing, nearly twice the dirty pages, and commit time dominating the operation.

This keeps the rule that only an underfilling leaf opens a run, but once one is open, extends it across every later leaf that has pending removals: those leaves are being rewritten anyway, so buffering them costs one copy and saves a page. A leaf with no removals still ends the run, so rewrites never spread past the region the scan is already dirtying. A leaf whose survivors exceed a page rides the splice but also ends the run, keeping the buffer bounded by roughly fanout x page_size plus one such leaf. The fits-in-one-page predicate is shared with the split policy (`leaf_split_required` is now defined in terms of it), so the cursor's packing decision and the mutator's rebuild policy cannot disagree.

## Benchmarks

Interleaved A/B against master (1M entries, 24-byte keys / 150-byte values, redb_benchmark shape, times include commit):

| workload | master | this PR | |
|---|---|---|---|
| `retain` keep 1-in-2 | 1582ms | 1078ms | ~1.5x faster |
| `extract_if` keep 1-in-2 | 1742ms | 1138ms | ~1.5x faster |
| `retain` keep 1-in-3 / 1-in-8 | — | — | parity |
| `retain` keep 1-in-4 | 675ms | 712ms | ~5% slower |
| leaf pages after retain 1-in-2 | 38,574 | 24,095 | |
| fragmented bytes after | 69MB | 8.5MB | |

The keep 1-in-4 cost is the extra buffer copy paid by the occasional healthy leaf that previously flushed 1:1 and now rides its run. The packed result also absorbs later random inserts slightly faster than a naturally grown tree of the same contents, so packing does not shift cost onto subsequent writes.

This closes the moderate-density gap against both earlier prototypes: it matches `codex/retain-owned-leaf-builder-prototype`'s packed layout (identical post-retain leaf counts) while never rewriting untouched leaves, and it beats `claude/optimize-extract-if-jJD8n`'s eager-rebuild throughput while keeping lazy extract semantics and a bounded buffer.

## Validation

- `just test` equivalent: cargo deny licenses, fmt, clippy (0 warnings), full `cargo test --all-features` — all green
- 10 minutes of `just fuzz` (206k runs; the harness drives `Retain`/`RetainIn`/`ExtractIf` directly): no crashes
- New regression test `retain_packs_moderate_density_survivors` fails on the old policy (survivors spread over 87% of the original leaves) and passes with packing

## Notes for review

- The run buffer bound grows from ~fanout x page_size / 3 to ~fanout x page_size. At the default 4KiB page size that is ~1MB transient per run; with a user-configured large page size (e.g. 64KiB+) it scales quadratically with page size, as the old bound already did. If that is a concern, an explicit byte budget on the run buffer would bound it independently of page size and could subsume the fits-one-page run-ending rule; left out here to keep the policy free of arbitrary constants.
- Pre-existing and unchanged: a run-ending leaf whose survivors repack from 1 child into 2 can grow the parent branch, and `finalize_branch_builder` never splits branches — at production page sizes this only produces an oversized branch page, and the u16 child-count panic needs ~2MiB pages, but a follow-up guard may be worth it.

https://claude.ai/code/session_01QMCHcLQyNCvHw6DhQGcjbh

---
_Generated by [Claude Code](https://claude.ai/code/session_01QMCHcLQyNCvHw6DhQGcjbh)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,10 @@
 ## 4.2.0 - 2026-XX-XX
 * Fix a crash during a transaction that grows the database file leaving the database permanently
   unopenable afterward.
-* Optimize `Table::retain()` and `Table::retain_in()`. Some benchmarks on large tables show a 25-50x speedup.
+* Optimize `Table::retain()` and `Table::retain_in()`. Some benchmarks on large tables show a
+  30-100x speedup, depending on the fraction of entries retained.
 * Optimize `Table::extract_if()` and `Table::extract_from_if()`. Benchmarks on large tables show
-  a 2-4x speedup, depending on the fraction of entries extracted, and iterating the extract
+  an 18-65x speedup, depending on the fraction of entries extracted, and iterating the extract
   iterator from both ends no longer degrades the removal batching.
 * After the extract iterator returns an error, later calls keep returning an error instead of
   continuing.
--- a/src/tree_store/btree_base.rs
+++ b/src/tree_store/btree_base.rs
@@ -913,14 +913,30 @@ pub(super) fn leaf_split_required(
     fixed_value_size: Option<usize>,
     page_size: usize,
 ) -> bool {
+    !leaf_fits_one_page(
+        num_pairs,
+        keys_values_bytes,
+        fixed_key_size,
+        fixed_value_size,
+        page_size,
+    ) && num_pairs > 1
+}
+
+// True when a leaf holding these pairs fits into a single order zero page.
+pub(super) fn leaf_fits_one_page(
+    num_pairs: usize,
+    keys_values_bytes: usize,
+    fixed_key_size: Option<usize>,
+    fixed_value_size: Option<usize>,
+    page_size: usize,
+) -> bool {
     let required = RawLeafBuilder::required_bytes(
         num_pairs,
         keys_values_bytes,
         fixed_key_size,
         fixed_value_size,
     );
-    let too_many_pairs = num_pairs > usize::from(u16::MAX);
-    (required > page_size || too_many_pairs) && num_pairs > 1
+    required <= page_size && u16::try_from(num_pairs).is_ok()
 }
 
 // Merge a leaf when it is less than 33% full: splits occur when a page is full and produce two
--- a/src/tree_store/btree_cursor.rs
+++ b/src/tree_store/btree_cursor.rs
@@ -1,7 +1,7 @@
 use crate::AccessGuard;
 use crate::tree_store::btree_base::{
     BRANCH, BranchAccessor, LEAF, LeafAccessor, OwnedEntryBuffer, leaf_below_merge_threshold,
-    retained_after_removals,
+    leaf_fits_one_page, retained_after_removals,
 };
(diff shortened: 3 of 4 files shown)


src/tree_store/btree_base.rs
```diff
@@ -912,15 +912,33 @@ pub(super) fn leaf_split_required(
...
+        keys_values_bytes,
+        fixed_key_size,
+        fixed_value_size,
+        page_size,
+    ) && num_pairs > 1
+}
+
+// True when the pairs fit in a single regular page, i.e. they can share a replacement leaf
+// with a run's other survivors. The one predicate underlying both the split policy above and
+// the cursor's decision to keep a run open, so the two cannot disagree.
```
cberner: The comment is too specific to the leaf run merging code. It should just state what this function checks: that a leaf fits into a single order zero page

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 * Optimize `Table::extract_if()` and `Table::extract_from_if()`. Benchmarks on large tables show
   a 2-4x speedup, depending on the fraction of entries extracted, and iterating the extract
   iterator from both ends no longer degrades the removal batching.
+* Pack the leaves rewritten by `retain()`, `retain_in()`, `extract_if()`, and
+  `extract_from_if()`. Removal patterns that shrink many neighboring leaves without emptying
+  them (for example, dropping every other entry) run about 1.5x faster and leave a defragmented
+  table.
```
cberner: Remove this bullet point, and just update the numbers for the above optimization lines. Make sure you use the latest release (4.1.0) as the baseline

src/tree_store/btree_cursor.rs
```diff
@@ -975,40 +975,72 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> CursorMut<'a, 'b, K, V> {
...
-            self.leaf_would_underfill(&position.leaf) && !position.path.is_empty()
-        };
-        if underfilling {
+        if self.state.removed_indexes.is_empty() {
+            if !run_open {
+                return Ok(LeafCloseOutcome::Unchanged);
+            }
+            // A fully retained leaf may still gain entries when the spliced
+            // run packs below the merge threshold and absorbs it as the
+            // adjacent child.
```
cberner: It's safe to fallthrough here? We don't need to push the entries or something like the else branch?

cberner: Yes — a leaf with no removals is not one of the run's `replaced_children`, so the splice leaves it in place as a preserved child of the parent. Pushing its entries into the buffer would duplicate them, since the splice doesn't free this leaf's page. The case where the run does need this leaf's entries (a packed run below the merge threshold) is handled inside `replace_leaf_children`, which absorbs the adjacent child and extends the replaced range over it — same as the single-delete merge path. I've expanded the comment here to state that.

---

PR #1290, opened 2026-07-05
Preserve torn commit slot bytes during recovery

cberner:
## Problem

A crash mid-`fsync` of a 1-phase commit legitimately leaves the secondary slot torn (invalid checksum). `select_primary_slot()` correctly keeps the valid primary — but on reopen with `needs_recovery`, the header is immediately rewritten via `to_bytes(true)`, and `TransactionHeader::to_bytes()` **always recomputes the slot checksum**. This re-serializes the parsed torn secondary with a freshly computed, now-*valid* checksum, "laundering" it.

The laundered slot then sits on disk for the entire repair (three full tree walks — minutes on a large database). If the process crashes again in that window, the next open sees two valid slots and promotes the laundered one (its transaction id `T+1 > T`). Tree-checksum verification is the only remaining defense, and torn byte-mixes exist that pass it:

- old roots + new txn id -> silent rollback of a durably-committed transaction;
- old user root + new system root -> the new `DATA_FREED_TABLE` lists pages the old user root still references, so a later commit frees and reuses live pages -> corruption.

Preconditions are narrow (byte-granular tear + double crash), but this is exactly the "poorly behaved media" model `docs/design.md` commits to defending, and the laundering defeats the checksum design: without it, a torn slot stays invalid forever and none of this is reachable.

This affects the two recovery rewrite paths (the open path and `clear_cache_and_reload`), and is present in released **v4.1.0** (verified against the tagged source — same `to_bytes` re-checksum + `pick_primary_for_repair` promotion), so it is not a regression on `master`.

## Fix

Preserve the original on-disk bytes of any slot that fails checksum verification, and write them back verbatim during a header rewrite, so a torn slot keeps its invalid checksum until a genuine commit overwrites it.

- `TransactionHeader` gains a `corrupt_bytes: Option<[u8; TRANSACTION_SIZE]>` field, set by `from_bytes()` whenever the parsed slot's checksum did not verify.
- `to_bytes()` returns those bytes verbatim when present, instead of recomputing a fresh (valid) checksum. This fixes both rewrite sites at the source, and means the newer-secondary promotion in `select_primary_slot()` can never see a laundered slot.
- `secondary_slot_mut()` is replaced by `write_secondary_slot()`, which clears `corrupt_bytes` as an atomic part of installing a new commit — so a torn slot's bytes are preserved only until the slot is genuinely overwritten, with no way to take a mutable reference and forget to clear it.

## Testing

- Added `recovery_does_not_launder_torn_slot`: fabricates a torn, newer secondary slot, opens with the repair aborted (freezing the exact post-rewrite on-disk state), and asserts the torn slot is byte-for-byte unchanged. Confirmed it fails on the pre-fix behavior ("recovery laundered the torn secondary slot") and passes with the fix.
- `cargo fmt --check` and `cargo clippy --all-targets --all-features` clean.
- Full `cargo test --all-features` suite passes, including `crash_consistency`.
- 90s / ~57k-iteration `cargo fuzz run fuzz_redb` smoke run on the crash-recovery target — no failures.
- Added a `CHANGELOG.md` entry.

https://claude.ai/code/session_018uK6KdWZRdM91oRQyiEGF8

---
_Generated by [Claude Code](https://claude.ai/code/session_018uK6KdWZRdM91oRQyiEGF8)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,10 @@
 * Fix `Database::check_integrity()` silently discarding transactions committed with
   `Durability::None` that had not yet been made durable by a later commit; a passing check now
   preserves them (making them durable) instead of rolling them back.
+* Fix a bug that could silently roll back or corrupt durably committed transactions if a crash
+  occurred while recovering from an earlier crash. Triggering it required two crashes -- one
+  interrupting a commit and another during the subsequent repair on the next open -- and it did
+  not affect transactions committed with two-phase commit.
 
 ### Python bindings
 * Add `Database.create(path)` for creating or opening a database file.
--- a/src/tree_store/page_store/header.rs
+++ b/src/tree_store/page_store/header.rs
@@ -365,8 +365,18 @@ impl DatabaseHeader {
         &self.transaction_slots[self.primary_slot ^ 1]
     }
 
-    pub(super) fn secondary_slot_mut(&mut self) -> &mut TransactionHeader {
-        &mut self.transaction_slots[self.primary_slot ^ 1]
+    // Overwrite the secondary slot with a newly committed transaction.
+    pub(super) fn write_secondary_slot(
+        &mut self,
+        transaction_id: TransactionId,
+        user_root: Option<BtreeHeader>,
+        system_root: Option<BtreeHeader>,
+    ) {
+        let slot = &mut self.transaction_slots[self.primary_slot ^ 1];
+        slot.transaction_id = transaction_id;
+        slot.user_root = user_root;
+        slot.system_root = system_root;
+        slot.corrupt_bytes = None;
     }
 
     pub(super) fn swap_primary_slot(&mut self) {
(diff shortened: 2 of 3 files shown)


cberner (commented):
The comments are too specific to the fix. Like the write secondary should not mention the issue. The change log should focus only on the user visible issue and its conditions to trigger, not the details

cberner:
Addressed both points (force-pushed):

- Trimmed the code comments to describe invariants rather than this specific fix — `write_secondary_slot` no longer references the issue, and the `corrupt_bytes`/`to_bytes` comments just explain that a slot which failed checksum verification is written back verbatim.
- Rewrote the CHANGELOG entry to focus on the user-visible symptom (durably committed transactions silently rolled back or corrupted) and the conditions to trigger it (two crashes — one during a commit, one during the subsequent repair; not with two-phase commit), dropping the implementation details.

---

PR #1306, opened 2026-08-04
Add owned accessors that keep the read transaction alive

cberner:
## Problem

`ReadOnlyTable::get()` documents that its guard "is reference counted and keeps the transaction alive until it is dropped", but nothing is actually kept alive: the guard holds only the page. The same applies to the guards and entries yielded by the `'static` iterators from `ReadOnlyTable::range()`, `ReadOnlyMultimapTable::get()`, and `ReadOnlyMultimapTable::range()`, which may outlive the iterator that produced them.

Holding any of these past the `ReadTransaction` unregisters the read snapshot, so a concurrent writer is free to reclaim the referenced pages: the writer's `commit()` panics on the `read_page_ref_counts` assertion in debug builds (reproduced with a simple insert/get/drop-txn/overwrite loop; the first commit panics).

## Change

Adds `get_owned()` and `range_owned()` to `ReadOnlyTable` and `ReadOnlyMultimapTable`, returning new `OwnedAccessGuard`, `OwnedRange`, `OwnedMultimapValue`, and `OwnedMultimapRange` types. These wrap the plain guard or iterator together with the transaction guard, keeping the transaction alive until the last of them is dropped (naming follows tokio's `OwnedMutexGuard` convention). `ReadTransaction::close()` correctly reports the transaction as still in use while any of them is alive.

The existing methods are untouched (`src/table.rs` has zero deletions); deprecating them is planned as a follow-up PR.

## Why wrapper types

Storing the transaction inside `AccessGuard` (as a field or an `EitherPage` variant) slowed range scans by 4-7% on the paths that never pin, through a combination of guard size (96 -> 104 bytes), drop glue that no longer inlined cross-crate, and branchier iterators. With wrapper types, cachegrind shows the existing paths at exact instruction parity with master (2,888.6M vs 2,888.9M instructions on 2M-entry range scans), and the owned paths cost about 1.5% more instructions than the borrowing ones.

The benchmark harness now calls the `ReadableTable` trait methods explicitly so it measures the lifetime-bound accessors, which are the ones comparable to the other engines' cursor semantics.

## Testing

- Four new regression tests covering `get_owned()` guards, `range_owned()` yields, and multimap owned values (inline and subtree) held past table + transaction drop while a writer commits repeatedly; each fails on master
- Full suite passes (327 tests), clippy pedantic and fmt clean, cargo-deny clean
- Examples and the crate-level doc example showcase the new accessors

https://claude.ai/code/session_01YCYPuM4nb94pF36ZX7R8sa

---
_Generated by [Claude Code](https://claude.ai/code/session_01YCYPuM4nb94pF36ZX7R8sa)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 * `check_integrity()` now recomputes the table counts stored alongside the data and system roots,
   repairing a file whose counts disagree with its trees. Such a file previously passed the check
   and then panicked, including from `Database::drop`.
+* Add `ReadOnlyTable::get_owned()`, `ReadOnlyTable::range_owned()`,
+  `ReadOnlyMultimapTable::get_owned()`, and `ReadOnlyMultimapTable::range_owned()`, which
+  return the new `OwnedAccessGuard`, `OwnedRange`, `OwnedMultimapValue`, and
+  `OwnedMultimapRange` types. These keep the read transaction alive until they are dropped,
+  including the guards yielded by the iterators, which may outlive the iterator that produced
+  them.
 * Fix a crash during a transaction that grows the database file leaving the database permanently
   unopenable afterward.
 * Fix a potential deadlock when removing a value from a multimap table causes its value-set to
--- a/src/table.rs
+++ b/src/table.rs
@@ -576,6 +576,17 @@ impl<K: Key + 'static, V: Value + 'static> ReadOnlyTable<K, V> {
         self.tree.get(key.borrow())
     }
 
+    /// This method is like [`ReadableTable::get()`], but the returned [`OwnedAccessGuard`] is
+    /// reference counted and keeps the transaction alive until it is dropped.
+    pub fn get_owned<'a>(
+        &self,
+        key: impl Borrow<K::SelfType<'a>>,
+    ) -> Result<Option<OwnedAccessGuard<V>>> {
+        Ok(self
+            .get(key)?
+            .map(|x| OwnedAccessGuard::new(x, self.transaction_guard.clone())))
+    }
+
     /// This method is like [`ReadableTable::range()`], but the iterator is reference counted and keeps the transaction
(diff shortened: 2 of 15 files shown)


src/table.rs
```diff
@@ -769,6 +793,82 @@ impl<K: Key + 'static, V: Value + 'static> DoubleEndedIterator for Range<'_, K,
...
 }
 
+/// An [`AccessGuard`] which also keeps the transaction alive
+///
+/// Returned by the reference-counted accessors of [`ReadOnlyTable`] and
+/// [`crate::ReadOnlyMultimapTable`], such as [`ReadOnlyTable::get_owned()`]: in addition to
+/// providing access to the data, it keeps the read transaction alive until it is dropped.
+///
+/// This is a separate type, rather than a field of every [`AccessGuard`], so that the
+/// lifetime-bound accessors do not pay for the reference counting.
```
cberner: Remove this implementation detail

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 * `check_integrity()` now recomputes the table counts stored alongside the data and system roots,
   repairing a file whose counts disagree with its trees. Such a file previously passed the check
   and then panicked, including from `Database::drop`.
+* Add `ReadOnlyTable::get_owned()`, `ReadOnlyTable::range_owned()`,
+  `ReadOnlyMultimapTable::get_owned()`, and `ReadOnlyMultimapTable::range_owned()`, which
+  return the new `OwnedAccessGuard`, `OwnedRange`, `OwnedMultimapValue`, and
+  `OwnedMultimapRange` types. These keep the read transaction alive until they are dropped,
+  including the guards yielded by the iterators, which may outlive the iterator that produced
+  them. The existing accessors are unchanged and do not pay for the reference counting.
```
cberner: Remove the note about existing accessors

Issue #1310, opened 2026-08-05
Ascending-order inserts leave leaf pages ~50% full


farhan-syah:
Inserting keys in ascending order leaves every leaf page about half full, so the same data occupies ~39% more space than inserting it in random order.

Repro, single table, `tempfile` the only extra dep:

```rust
use redb::{Database, ReadableDatabase, ReadableTableMetadata, TableDefinition};

const T: TableDefinition<&[u8], &[u8]> = TableDefinition::new("x");

fn main() {
    for sorted in [false, true] {
        let mut state = 0x2545f4914f6cdd1du64;
        let mut next = || {
            state ^= state << 13;
            state ^= state >> 7;
            state ^= state << 17;
            state
        };

        let mut keys: Vec<[u8; 24]> = (0..1_000_000)
            .map(|_| {
                let mut key = [0u8; 24];
                key[..8].copy_from_slice(&next().to_le_bytes());
                key[8..16].copy_from_slice(&next().to_le_bytes());
                key[16..].copy_from_slice(&next().to_le_bytes());
                key
            })
            .collect();
        if sorted {
            keys.sort_unstable();
        }

        let value = vec![0u8; 150];
        let file = tempfile::NamedTempFile::new().unwrap();
        let db = Database::create(file.path()).unwrap();
        let txn = db.begin_write().unwrap();
        {
            let mut table = txn.open_table(T).unwrap();
            for key in &keys {
                table.insert(key.as_slice(), value.as_slice()).unwrap();
            }
        }
        txn.commit().unwrap();

        let read = db.begin_read().unwrap();
        let stats = read.open_table(T).unwrap().stats().unwrap();
        println!(
            "{:>6} order: {} leaf pages, {} MiB fragmented",
            if sorted { "sorted" } else { "random" },
            stats.leaf_pages(),
            stats.fragmented_bytes() / 1024 / 1024,
        );
    }
}
```

```
random order: 64848 leaf pages, 80 MiB fragmented
sorted order: 83333 leaf pages, 155 MiB fragmented
```

Identical data and identical `stored_bytes`, but 28.6% more leaf pages. Accounting for per-entry offsets, the sorted leaves sit around 50% occupancy against roughly 70% for random.

[...]

PR #1312, opened 2026-08-05
Add Table::insert_with_hint() and InsertHint::Append

farhan-syah:
## Problem

Inserting keys in ascending order leaves every leaf about half full. `LeafBuilder::build_split()` divides at the byte midpoint, and an ascending insert always splits the rightmost leaf, so the leading half is sealed at half capacity with nothing to refill it. Loading 1M pairs in ascending order costs 28.6% more leaf pages than loading the same pairs in random order, where later inserts land in the half-empty pages and fill them.

## Fix

Add `InsertHint` and `Table::insert_with_hint()`, as discussed in #1310:

```rust
#[non_exhaustive]
pub enum InsertHint {
    Append,
}

table.insert_with_hint(&key, &value, InsertHint::Append)?;
```

`InsertHint::Append` divides after the second to last pair instead of evenly, so the leading page stays full. The hint is threaded through `BtreeMut::insert_with_hint()` and `MutateHelper` to the split site; `insert()` is unchanged and passes `None`.

The hint is advisory and verified rather than trusted: the split site still checks that the insert actually landed past the leaf's last key, so a hint that turns out to be wrong is ignored rather than producing a malformed leaf. This differs from LMDB's `MDB_APPEND`, which rejects out-of-order keys with `KEYEXIST`. Happy to switch to rejecting if you would rather match LMDB.

Packing leaves fully does cost something: a later key falling inside a packed leaf has to split it again. Loading 1M pairs ascending and then inserting 200k random ones uses 90,129 leaf pages with the hint against 83,362 without, so this is a poor trade for a table that keeps taking scattered writes. That is why it is opt-in per insert rather than automatic.

## Testing

`cargo test --all-features` passes. New `tests/insert_hint_tests.rs` covers packing fewer leaves, stored data being unchanged, a wrong hint being harmless (5,000 descending keys inserted with the hint set, then `check_integrity()`), replacing an existing key, and hinted and unhinted inserts interleaved in one transaction.

New `insert_hint_benchmark`, 1M pairs of 24-byte keys and 150-byte values:

| | load | leaf pages | data | DB size |
|---|---|---|---|---|
| random order | 1794ms | 64799 | 64.5% | 514MiB |
| ascending order | 1233ms | 83333 | 49.7% | 514MiB |
| ascending order, Append hint | 910ms | 45455 | 91.1% | 257MiB |

The `data` column is the share of the table's bytes holding user data, rather than padding left behind by splitting.

On a larger load (1M vertices, 8M edges, a vertex table plus forward and reverse adjacency tables, 17M pairs in one transaction) the compacted database goes from 901.55 MiB to 456.88 MiB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,10 @@
   Existing databases created by older versions remain readable. If an older database already used
   such a colliding composite name, its stored type identity remains ambiguous and may still open
   under either spelling.
+* Add `Table::insert_with_hint()` and `InsertHint`. `InsertHint::Append` packs a leaf fully
+  instead of splitting it evenly when an insert past the leaf's last key forces a split. This
+  reduces free space left behind when loading data in ascending key order, at the cost of
+  leaves needing to split again to accept a later key that falls inside them.
 
 ### Python bindings
 * Add `Database.create(path)` for creating or opening a database file.
--- a/crates/redb-bench/Cargo.toml
+++ b/crates/redb-bench/Cargo.toml
@@ -64,3 +64,7 @@ harness = false
 [[bench]]
 name = "syscall_benchmark"
 harness = false
+
+[[bench]]
+name = "insert_hint_benchmark"
+harness = false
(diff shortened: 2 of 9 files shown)


cberner:
Merged. Thanks!

PR #1313, opened 2026-08-05
Tag user-defined field types in the derived TypeName

cberner:
Last in the redb-derive stack (#1314 → #1311 → this). Both predecessors have merged; this PR is rebased onto master and is the only one remaining. It carries the changelog for the whole stack, under a `redb-derive (unreleased)` heading (no version changes in the stack; versioning happens at release).

## The bug

Derived `type_name()` composed the struct's `TypeName` from only the *name strings* of its field types, discarding the user-defined vs built-in classification. A user type named `String` and the built-in `String` produced byte-identical `TypeName`s for their containing structs, so a table written with one opened as the other without error and silently misread the stored bytes.

## The fix

This is the derive-side analog of the composite-type collision fixed in 4388342/27160a5, but bubbling the classification up cannot work here: a derived struct is itself always user-defined, so the outer classification carries no signal. Instead, user-defined field types are tagged in the composed name (`Holder {id: Inherent#user}`); `#` cannot appear in any built-in type name, so tagged and untagged renderings cannot collide.

The generated code reads the classification by rebuilding the field's `TypeName` through `TypeName::new` and comparing — `TypeName` equality is exactly (classification, name) — rather than calling `is_user_defined()`. That keeps it compiling when `#[redb(crate = "...")]` names an older redb: `new`, `name()`, and `PartialEq` are public back to redb 3.0, which was already the floor for field-bearing derives (`name()` is private in ≤2.6). `TypeName::is_user_defined()` stays `pub(crate)` — nothing outside the crate needs it (per review).

## Compatibility

- Derived structs whose fields are **all built-in types** (including tuples, arrays, `Option`/`Vec` of built-ins) keep a byte-identical `TypeName` — their existing tables still open. The pre-existing tests pin this: all expected name strings are unchanged.
- Structs containing **user-defined field types** change identity; their derive-0.1 tables report `TableTypeMismatch` and need migration. Per review, this includes the `chrono` types whose `Value` implementations redb provides under the `chrono_v0_4` feature — those are classified user-defined (`TypeName::new`) and, unlike `uuid::Uuid` (which never compiled under 0.1, see #1311), chrono fields did compile, so such tables can exist. The changelog calls this out.
- The `#[redb(crate = "...")]` multi-version use case keeps working (per review): field-bearing derives compile against redb 3.0+, pinned by a new `redb3_0 = "=3.0.0"` dev-dependency.

## Tests

- User type named `String` vs built-in `String` → distinct `TypeName`s, and an end-to-end `open_table` mismatch error.
- Nested derived structs, and `Vec`/tuple composites wrapping user types → tagged; composites of built-ins → untagged.
- Equality pin that built-in-only structs keep the exact pre-change identity.
- A field-bearing struct with a user-defined field derives against redb 3.0.0 via `#[redb(crate = "redb3_0")]`, asserts its tagged name, and round-trips through a 3.0 database; a `types.rs` unit test keeps the reconstruction agreeing with `is_user_defined()` for every classification.

https://claude.ai/code/session_01EuLGiEERgJR51vhJNtYwRA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,19 @@
   instead of splitting it evenly when an insert past the leaf's last key forces a split. This
   reduces free space left behind when loading data in ascending key order, at the cost of
   leaves needing to split again to accept a later key that falls inside them.
+### redb-derive (unreleased)
+* Fix the derived implementations calling inherent methods named `fixed_width`, `from_bytes`,
+  `as_bytes`, or `type_name` on field types, instead of the `Value` trait methods. The
+  generated code now uses fully qualified paths, and no longer requires the `Value` and `Key`
+  traits to be in scope at the derive site.
+* Add `#[redb(crate = "...")]` to name the crate the implementations are generated for, when
+  redb is renamed or present in several versions. Structs with fields require redb 3.0+.
+* Fix derived structs producing the same `TypeName` when two field types with different
+  definitions share a name, such as a user-defined type named `String` and the built-in
+  `String`. User-defined field types (including the `chrono` types redb provides) are now
+  tagged in the derived `TypeName`, so structs containing them change type identity: their
+  existing tables report `TableTypeMismatch` and must be migrated. Structs whose fields are
+  all built-in types are unaffected.
 
 ### Python bindings
 * Add `Database.create(path)` for creating or opening a database file.
--- a/crates/redb-derive/src/lib.rs
+++ b/crates/redb-derive/src/lib.rs
@@ -45,7 +45,12 @@ fn redb_crate_path_for(input: &DeriveInput) -> syn::Result<proc_macro2::TokenStr
 /// The generated implementation refers to the redb crate as `::redb`. When it should be
 /// generated for a crate under another name -- a renamed dependency
 /// (`my_redb = { package = "redb" }`), or one of several versions of redb -- name that crate
-/// with `#[redb(crate = "my_redb")]`.
+/// with `#[redb(crate = "my_redb")]`. For a struct with fields, the generated implementation
+/// uses `TypeName` API that redb makes public from 3.0 on; a unit struct works with older
+/// versions as well. The tagging of user-defined field types in the composed type name follows
+/// the named version's classification: redb 4.1 and older classify composites (`Option`,
+/// `Vec`, tuples, arrays) of user-defined types as built-in, so such composite fields render
+/// untagged there, and their names collide exactly where that version's own type names do.
 #[proc_macro_derive(Key, attributes(redb))]
 pub fn derive_key(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
@@ -88,7 +93,12 @@ fn generate_key_impl(input: &DeriveInput) -> syn::Result<proc_macro2::TokenStrea
 /// The generated implementation refers to the redb crate as `::redb`. When it should be
 /// generated for a crate under another name -- a renamed dependency
 /// (`my_redb = { package = "redb" }`), or one of several versions of redb -- name that crate
-/// with `#[redb(crate = "my_redb")]`.
+/// with `#[redb(crate = "my_redb")]`. For a struct with fields, the generated implementation
+/// uses `TypeName` API that redb makes public from 3.0 on; a unit struct works with older
+/// versions as well. The tagging of user-defined field types in the composed type name follows
+/// the named version's classification: redb 4.1 and older classify composites (`Option`,
+/// `Vec`, tuples, arrays) of user-defined types as built-in, so such composite fields render
+/// untagged there, and their names collide exactly where that version's own type names do.
 #[proc_macro_derive(Value, attributes(redb))]
 pub fn derive_value(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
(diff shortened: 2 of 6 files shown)


crates/redb-derive/src/lib.rs
```diff
@@ -183,6 +183,27 @@ fn generate_self_type(
...
+// tagged to keep the composed names distinct. Built-in field types render as the bare name,
+// which keeps the `TypeName` of structs made only of built-in types unchanged.
+fn render_field_type_name(
+    field_type: &syn::Type,
+    redb: &proc_macro2::TokenStream,
+) -> proc_macro2::TokenStream {
+    quote! {
+        {
+            let type_name = <#field_type as #redb::Value>::type_name();
+            if type_name.is_user_defined() {
```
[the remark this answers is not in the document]
cberner: Confirmed — fixed in f6fd07e. The generated code no longer calls `is_user_defined()`: it reads the classification by rebuilding the field's `TypeName` through `TypeName::new` and comparing, which is equivalent because `TypeName` equality is exactly (classification, name). `new`, `name()`, and `PartialEq` are public back to redb 3.0, which was already the floor for field-bearing derives (`name()` is private in ≤2.6, as the unit-struct test notes).

Pinned in CI by a new `redb3_0 = { package = "redb", version = "=3.0.0" }` dev-dependency: `crate_attr_tests.rs` now derives a field-bearing struct with a user-defined field type against it and round-trips through a 3.0 database, and a unit test in `types.rs` keeps the reconstruction agreeing with `is_user_defined()` for every classification. The macro docs and changelog state the 3.0 floor. `is_user_defined()` itself stays public — hand-written composite `Value` implementations still want it, and they ship in lockstep with redb.

---

src/types.rs
```diff
@@ -109,7 +109,15 @@ impl TypeName {
...
-    pub(crate) fn is_user_defined(&self) -> bool {
+    /// Whether this name involves a user-defined type: either it was created with
+    /// [`TypeName::new`], or it names a composite type (such as a tuple or `Vec`) that contains
+    /// a user-defined type.
+    ///
+    /// A user-defined type may share its name string with a built-in type; only the
+    /// classification carried by the `TypeName` distinguishes the two. `Value` implementations
+    /// that compose their [`type_name()`](Value::type_name) from the names of inner types should
+    /// therefore incorporate this flag, so that the composed names cannot collide either.
+    pub fn is_user_defined(&self) -> bool {
```
cberner: Does this really have to be public?

CHANGELOG.md
```diff
@@ -101,6 +101,35 @@
...
+  share a name, such as a user-defined type named `String` and the built-in `String`. A table
+  written using one could be opened with the other without error, silently misinterpreting the
+  stored bytes. User-defined field types are now tagged in the derived `TypeName`. This changes
+  the type identity of derived structs that contain user-defined field types (including derived
+  ones, the `chrono` types whose `Value` implementations redb provides under the `chrono_v0_4`
+  feature, and `Option`/`Vec`/tuple/array composites of them): their tables written by redb-derive
+  0.1 are reported as `TableError::TableTypeMismatch` and must be migrated, for example by
+  reading them under a hand-written `Value` implementation that reproduces the previous
+  `type_name()` and copying into a new table. Derived structs whose fields are all built-in
+  types keep their previous identity, and their existing tables are unaffected.
```
cberner: Can we make this less verbose please?

crates/redb-derive/tests/derive_tests.rs
```diff
@@ -287,7 +287,7 @@ fn test_inherent_methods_do_not_shadow_trait_methods() {
     // The inherent `type_name` returns "bogus"
     assert_eq!(
         InherentHolder::type_name().name(),
-        "InherentHolder {a: Inherent, b: u64}"
+        "InherentHolder {a: Inherent#user, b: u64}"
```
cberner, line 290: Is "#user" the same as what we used for Tuples containing user types?

src/types.rs
cberner: No — not anymore. The generated code stopped calling it when it switched to the `TypeName::new` reconstruction (for redb-3.0 compat), so nothing outside the crate uses it. Reverted to `pub(crate)` in d2c5635, and dropped the changelog entry that announced it.

---

CHANGELOG.md
cberner: Trimmed in d2c5635 — about half the length now. Dropped the failure-mode walkthroughs and the migration recipe; kept the chrono callout and the `TableTypeMismatch` symptom.

---

crates/redb-derive/tests/derive_tests.rs
cberner: Same classification, different encoding channel. For redb's own composites (tuples/arrays/`Option`/`Vec`), the composed name string is unchanged and the user-defined-ness bubbles up into the `TypeName`'s *classification* via `into_composite`. That channel isn't available here: a derived struct's `TypeName` is built with `TypeName::new`, so its classification is always user-defined no matter what the fields are — the bit carries no signal. So the derive encodes the per-field classification in the *name string* instead, as the `#user` tag (`#` can't occur in any built-in type name, so tagged and untagged renderings can't collide).

The two mechanisms compose rather than overlap: a `Vec<UserType>` field is classified user-defined by the bubbling, and the derive reads that classification and renders the tag — the `WrappedUser` test below pins `"WrappedUser {v: Vec<SingleField {value: i32}>#user, w: Vec<u32>}"`.

---

crates/redb-derive/src/lib.rs
```diff
@@ -183,6 +187,34 @@ fn generate_self_type(
...
+// `name()`, and `PartialEq` reach back to redb 3.0, which `#[redb(crate = "...")]` may name
+// (`name()` is private before that, so field-bearing structs already floor there).
+fn render_field_type_name(
+    field_type: &syn::Type,
+    redb: &proc_macro2::TokenStream,
+) -> proc_macro2::TokenStream {
+    quote! {
+        {
+            let type_name = <#field_type as #redb::Value>::type_name();
+            if type_name == #redb::TypeName::new(type_name.name()) {
```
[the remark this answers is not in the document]
cberner: Confirmed, and resolved by documenting + pinning (the second of the suggested options) in 00e8c67.

Recursive tagging was considered and rejected as unsound: the macro only sees the field's type *syntax*, so an alias (`type A = Option<MyString>`) would render differently from the spelled-out composite, and a version-dependent rendering (`Option<String#user>` under 3.0 vs `Option<String>#user` under 4.2) would change a struct's composed name when the app upgrades redb, breaking its existing tables — the failure mode this PR exists to prevent.

The residual collision is exactly the named version's own: redb ≤ 4.1 classifies `Option<MyString>` as built-in, so a plain 3.0 table of `Option<MyString>` already collides with `Option<String>` with no derive involved (that's the bug 4.2 fixes). The derive against an old redb reproduces that version's semantics rather than pretending to retrofit it. Now stated in both derive macro docs and in `render_field_type_name`'s comment, and pinned by a new `Option<Old3Inner>` field in the `redb3_0` test, which asserts the untagged rendering alongside the tagged direct field.

---

PR #1316, opened 2026-08-06
Add InsertWithHint fuzz operation

farhan-syah:
## Summary

Adds `FuzzOperation::InsertWithHint` so the fuzz target exercises `Table::insert_with_hint`. Nothing currently reaches it.

The hint is advisory: it changes how a leaf is split but never what is stored, and a caller may be wrong about the order without any consequence. That property is what wants adversarial coverage, since the fuzzer will freely emit hinted inserts for keys that are not greater than everything already present, interleaved with plain inserts, removals and the rest.

The operation records the same expectation in the reference model as a plain insert, so any divergence in stored data fails the existing comparison. Multimap tables have no insert hints, so it is a no-op there.

## Testing

`cargo fuzz run --sanitizer=none fuzz_redb -- -max_len=10000 -max_total_time=600`: 275255 runs, no crashes.

### DIFF
--- a/fuzz/fuzz_targets/common.rs
+++ b/fuzz/fuzz_targets/common.rs
@@ -112,6 +112,10 @@ pub(crate) enum FuzzOperation {
         key: BoundedU64<KEY_SPACE>,
         value_size: BinomialDifferenceBoundedUSize<MAX_VALUE_SIZE>,
     },
+    InsertWithHint {
+        key: BoundedU64<KEY_SPACE>,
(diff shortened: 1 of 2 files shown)


cberner (approved):

farhan-syah:
Closing this — with `CursorMut` replacing `insert_with_hint`, the operation this adds would be fuzzing an API that is on its way out. If the append policy ends up derived from tree position instead, I will send fuzz coverage for that path rather than reopen this.

PR #1318, opened 2026-08-06
Keep non-durable commits in the write buffer instead of flushing them

cberner:
Resolves the TODO in `cached_file.rs`: `Durability::None` commits previously flushed every dirty page to the OS through `write_barrier()` — one write syscall per page, every commit — which made them much slower than the non-durable modes of other KV stores. They now do no file I/O at all.

## Design

`write_barrier()` leaves the dirty pages in the write buffer and sets an atomic flag (`committed_pages_buffered`) recording that the buffer holds pages belonging to the committed, reader-visible state. The pages are written out by the next durable commit, a clean close, or cache pressure — and repeated writes to the same page across non-durable commits now coalesce into a single eventual write instead of one write per commit.

- **Reader visibility**: reads with `PageHint::Clean` (read transactions) could previously skip the write buffer because commits always drained it. While the flag is set, they consult it. `read()` checks the striped read cache before the buffer mutex, and a Clean-hint buffer hit promotes the page into the read cache as a clone of the same `Arc` — so the two caches cannot diverge, and repeated reads of hot non-durable pages reach them through the striped locks instead of contending on the mutex (without the promotion, multithreaded reads of recently committed data regressed 70–90%; with it they match baseline).
- **Cache accounting**: unchanged. Retained pages keep counting against the write buffer's existing half-of-cache budget, and the existing spill logic in `write()` bounds the memory that accumulated non-durable commits can hold; spilled pages remain readable from the file.
- **Rollback**: `clear_cache_and_reload()` now discards the buffer instead of flushing it — the buffered pages belong to the state being rolled back, and writing them out could even fail beyond end-of-file after an external truncation. `check_integrity()`'s promotion path is unchanged and writes the buffered pages out through the promoting durable commit.

## Why readers cannot observe uncommitted data

The buffer also holds uncommitted pages of the in-progress write transaction, and the flag makes readers look into that buffer — so this was audited explicitly:

- A reader only requests offsets reachable from the root snapshot it captured under the state mutex, and `write_barrier()` runs (with `debug_assert_no_dirty_pages`) before the new root is published.
- Pages are copy-on-write: an in-progress transaction's dirty pages are either freshly allocated offsets or in-place mutations of pages it allocated itself, never pages reachable from a committed root.
- A freed offset is only reallocated (and re-buffered) after no live read transaction can reference it. Every free path of reader-reachable pages is gated on reader liveness: deferred freed-table processing, `process_freed_pages_nondurable` (gated on `oldest_live_read_nondurable_transaction`), and savepoint restore (queues to the freed tables). The system-tree pages freed immediately at commit are never traversed by read transactions — the pre-existing invariant `durable_commit()` already relies on.

So the flag only changes *where* committed pages are found, never *which* pages a reader observes. This is the same reachability invariant that `PageHint::None` reads (which always consulted the buffer) have relied on all along. As defense in depth, a protocol violation would panic (`LRUWriteCache::get` unwraps a taken entry; `WritablePage::mem_mut` unwraps `Arc::get_mut`) rather than silently return wrong data. The invariant is now documented on the flag field.

## Benchmarks

The full `redb_benchmark` showed ±30–40% run-to-run noise in the benchmarking environment, so an alternating A/B micro-benchmark (3 rounds each; 500k-row seed; write-syscall counts from `/proc/self/io`) was used for the targeted phases:

| Phase | Before | After |
|---|---|---|
| 20k single-insert nosync commits | 663 / 670 / 683 ms | **591 / 591 / 595 ms** |
| Write syscalls during those commits | 161,564 (~8/commit) | **0** |
| 1M random reads, 1 thread (mixed durable + non-durable) | 738–809 ms | 776–862 ms |
| 1M random reads x 4 threads | 868–1013 ms | 923–958 ms |
| Durable commit flushing the accumulated pages | 10–17 ms | 15–23 ms |

~12% faster nosync commits even on a fast virtio disk where syscalls are cheapest — the win grows on slower storage — with reads and the deferred flush unchanged within noise. In the full suite, all other phases (bulk load, durable writes, batch writes, reads, removals, compaction) are within baseline self-variance.

## Validation

- Full test suite passes (353 tests), fmt and clippy clean.
- `fuzz_redb` run: 127,099 executions (non-durable commits, savepoints, simulated crashes, tiny caches) with zero failures.
- New tests pin the behavior: `write_barrier()` issues zero backend writes while Clean-hint readers see the data; pages spilled under pressure stay readable; discarded buffers roll back; and an end-to-end test that `Durability::None` commits issue zero backend writes. The existing disk-corruption `check_integrity` test now uses a zero-size cache so the page it corrupts is genuinely spilled to the file, matching the new semantics (pages still buffered are authoritative in memory and rewritten at promotion).

---
_Generated by [Claude Code](https://claude.ai/code/session_01D6r1JSUJ54zhBzWoVNxsYx)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   be added behind the `experimental_cursor` feature flag, like the table cursors' were.
 
 ## 4.2.0 - 2026-XX-XX
+* `Durability::None` commits are about 2x faster.
 * Commits now flush table root updates in a deterministic order, removing a source of
   nondeterminism that could make identical operation sequences produce differing database files
 * Add a `std` feature, enabled by default, in preparation for a future `no_std` mode. It gates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # redb - Changelog
 
 ## 4.2.0 - 2026-XX-XX
+* Improve performance of `Durability::None` commits. They no longer write dirty pages to the file:
+  the pages are retained in the in-memory cache, where readers see them, and are written back by
+  the next durable commit or when evicted by cache pressure.
 * Fix `WriteTransaction::stats()` returning garbage statistics, or panicking when debug assertions
   are enabled, when called while a table is open and modified in the same transaction.
 * Fix a deadlock when a `Database` was dropped while a `WriteTransaction` was live. A live
--- a/src/transactions.rs
+++ b/src/transactions.rs
@@ -363,6 +363,10 @@ impl DatabaseStats {
 pub enum Durability {
     /// Commits with this durability level will not be persisted to disk unless followed by a
     /// commit with [`Durability::Immediate`].
+    ///
+    /// Such commits do not write to the file. The changes are held in the in-memory cache, where
+    /// other transactions see them, until they are written back by a durable commit, a clean
+    /// database close, or cache pressure.
     None,
     /// Commits with this durability level are guaranteed to be persistent as soon as
     /// [`WriteTransaction::commit`] returns.
--- a/src/tree_store/page_store/cached_file.rs
+++ b/src/tree_store/page_store/cached_file.rs
@@ -195,6 +195,24 @@ impl CheckedBackend {
         }
         result.map_err(StorageError::from)
     }
+
+    // Like write(), but a failure does not permanently fail the backend. For a write that only an
+    // optimization depends on, latching would turn every later operation into a PreviousIo error
+    // over data that nothing was waiting on.
+    fn write_best_effort(&self, offset: u64, data: &[u8]) -> Result<()> {
+        self.check_failure()?;
+        self.file.write(offset, data).map_err(StorageError::from)
+    }
+}
+
+// Whether the data written by a write buffer flush has to reach the file. Either way a page whose
+// write fails stays buffered; they differ in what a failure does to the backend.
+#[derive(Copy, Clone)]
+enum Writeback {
+    // A failure is propagated and permanently fails the backend
+    Required,
+    // A failure leaves the backend usable
+    BestEffort,
 }
 
 pub(super) struct PagedCachedFile {
@@ -213,6 +231,10 @@ pub(super) struct PagedCachedFile {
     // writes are in progress, while write-heavy workloads never starve readers
     // below 50%.
     //
+    // Non-durable commits leave their pages in the write buffer (see write_barrier()), so it is
+    // not necessarily empty between transactions. Invariant 1 bounds what they accumulate, and
+    // read() flushes them out under pressure, so they cannot hold the read cache below 100%.
+    //
     // We track usage with two atomic counters and compute the total on the fly.
     // The resulting read is not perfectly atomic (between loading the two
     // counters a concurrent operation could change one), but the budget is a
@@ -221,6 +243,15 @@ pub(super) struct PagedCachedFile {
     // negligible accuracy gain.
     read_cache_bytes: AtomicUsize,
     write_buffer_bytes: AtomicUsize,
+    // True when the write buffer holds committed, reader-visible pages, left there by a
+    // non-durable commit (see write_barrier()) instead of being written to the file. While set,
+    // PageHint::Clean reads must consult the buffer, since a clean page may live there rather
+    // than in the file.
+    //
+    // The buffer also holds the in-progress write transaction's uncommitted pages, but readers
+    // cannot observe them: pages are copy-on-write, and a freed offset is only reallocated once
+    // no live read transaction can reference it.
+    committed_pages_buffered: AtomicBool,
     max_cache_size: usize,
     // Rotates the starting stripe for read-cache eviction
     next_eviction_stripe: AtomicUsize,
@@ -261,6 +292,7 @@ impl PagedCachedFile {
             page_size,
             read_cache_bytes: AtomicUsize::new(0),
             write_buffer_bytes: AtomicUsize::new(0),
+            committed_pages_buffered: AtomicBool::new(false),
             max_cache_size,
             next_eviction_stripe: AtomicUsize::new(0),
             #[cfg(feature = "cache_metrics")]
@@ -396,8 +428,28 @@ impl PagedCachedFile {
             }
             write_buffer.clear();
         }
+        // The buffer is empty, so readers no longer need to consult it
+        self.committed_pages_buffered
+            .store(false, Ordering::Release);
 
         Ok(())
+    }
+
+    // Drop all buffered writes without writing them to the file. Only valid when the state they
+    // belong to is being abandoned (e.g. rolling back to the on-disk state).
+    pub(super) fn discard_write_buffer(&self) {
+        self.committed_pages_buffered
+            .store(false, Ordering::Release);
+        for stripe in &self.write_buffer {
+            let mut write_buffer = stripe.lock().unwrap();
+            for (_, buffer) in write_buffer.cache.iter() {
+                // Like flush_write_buffer(), decrement per page rather than storing zero: a
+                // concurrent writer may be adding to a stripe this loop has already passed.
+                self.write_buffer_bytes
+                    .fetch_sub(buffer.as_ref().unwrap().len(), Ordering::AcqRel);
+            }
+            write_buffer.clear();
+        }
     }
 
     // Caller should invalidate all cached pages that are no longer valid
@@ -443,11 +495,12 @@ impl PagedCachedFile {
         self.file.sync_data()
     }
 
-    // Make writes visible to readers, but does not guarantee any durability
-    pub(super) fn write_barrier(&self) -> Result {
-        // TODO: non-durable commits would be much faster, if this did not issues writes to disk,
-        // and instead just made the data visible to readers
-        self.flush_write_buffer()
+    // Make buffered writes visible to readers, without writing them to the file or guaranteeing
+    // durability. They stay in the buffer, against its budget, until flushed or evicted.
+    pub(super) fn write_barrier(&self) {
+        if self.write_buffer_bytes.load(Ordering::Acquire) > 0 {
+            self.committed_pages_buffered.store(true, Ordering::Release);
+        }
     }
 
     // Read directly from the file, ignoring any cached data
@@ -473,7 +526,8 @@ impl PagedCachedFile {
         #[cfg(feature = "cache_metrics")]
         self.reads_total.fetch_add(1, Ordering::AcqRel);
 
-        if !matches!(hint, PageHint::Clean) {
+        // A write transaction's own dirty pages are in the write buffer, so look there first.
+        if matches!(hint, PageHint::None) {
             let lock = self.write_buffer_stripe(offset).lock().unwrap();
             if let Some(cached) = lock.get(offset) {
                 #[cfg(feature = "cache_metrics")]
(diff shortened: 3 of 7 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -532,7 +532,12 @@ impl TransactionalMemory {
     }
 
     pub(crate) fn clear_cache_and_reload(&mut self) -> Result<bool, DatabaseError> {
-        self.storage.flush()?;
+        // The in-memory state is being discarded in favor of the on-disk state, so buffered
+        // writes (which can only belong to the discarded state) are dropped rather than written
+        // out. Writing them could even fail: after an external truncation they may fall beyond
+        // the end of the file.
+        self.storage.discard_write_buffer();
+        self.storage.sync_file()?;
```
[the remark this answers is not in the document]
cberner: The exact sequence is unreachable today: `clear_cache_and_reload()`'s only caller is `check_integrity_inner()`, whose pending-commit branch calls `clear_read_cache()` before reaching it, and no promotions can happen during the check itself — it holds exclusive access (`&mut self` plus the `Arc::get_mut` guard), and verification reads use `PageHint::None`, which does not promote. A failed `sync_file()` also latches `io_failed`, so post-error reads that miss the caches fail with `PreviousIo` rather than reading the file.

That said, the ordering made the consistency depend on those non-local caller invariants, which is fragile. Hardened in 7d8ead2: the read cache is now invalidated together with the buffer discard, before the fallible sync, so any early error return leaves a self-consistent cache state; `discard_write_buffer()`'s contract now documents that read-cache clones of buffered pages must be invalidated by callers.

---

src/transactions.rs
```diff
@@ -363,6 +363,10 @@ impl DatabaseStats {
 pub enum Durability {
     /// Commits with this durability level will not be persisted to disk unless followed by a
     /// commit with [`Durability::Immediate`].
+    ///
+    /// Such commits do not write to the file. The changes are held in the in-memory cache, where
+    /// other transactions see them, until they are written back by a durable commit, a clean
+    /// database close, or cache pressure.
```
cberner: Remove this. We shouldn't document and guarantee implementation details like this

cberner: Removed in 37f3167 — the doc block is gone, leaving just the "not persisted unless followed by a `Durability::Immediate` commit" guarantee.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -494,7 +530,37 @@ impl PagedCachedFile {
...
+        // Checking the read cache first keeps the buffer's mutex, which serializes readers that
+        // the striped read cache lets run concurrently, off the path of pages cached for reading.
+        if matches!(hint, PageHint::Clean) && self.committed_pages_buffered.load(Ordering::Acquire)
+        {
+            let lock = self.write_buffer_stripe(offset).lock().unwrap();
+            if let Some(cached) = lock.get(offset) {
+                #[cfg(feature = "cache_metrics")]
+                self.reads_hits.fetch_add(1, Ordering::Release);
+                debug_assert_eq!(cached.len(), len);
+                return Ok(cached.clone());
```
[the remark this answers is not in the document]
cberner: Confirmed by measurement and fixed in 2cf686a. An earlier revision of this PR did promote clean buffer hits into the read cache; I dropped it after master striped the write buffer, on the theory that 131 stripes made the contention immaterial, and a benchmark showed no regression. That benchmark was the weak part of the argument: its reads were spread over the whole keyspace, so only ~4% of them hit the buffer.

Re-measured with reads concentrated on the non-durably committed keys, so nearly every read hits a buffer stripe (3 alternating rounds, 4 threads, 2M reads each):

| | without promotion | with promotion |
|---|---|---|
| 4 threads | 1894 / 2151 / 1262 ms | **1193 / 1118 / 1160 ms** |
| 1 thread | 579 / 760 / 567 ms | 560 / 619 / 548 ms |

So the mutex does serialize concurrent readers of the same page even when striped, ~40% at 4 threads, and the promotion is back.

On the accounting, which was my reason for dropping it: both caches then hold the same `Arc` and each counts it, so the budget is conservative rather than wrong, and it resolves itself — whichever copy is dropped first (the buffer's when flushed, the read cache's when evicted) leaves the page counted exactly once. `flush_write_buffer()` subtracts the replaced entry when it reinserts, and `free` drops both copies via `invalidate_cache` + `cancel_pending_write`, so no path double-subtracts.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -494,7 +530,37 @@ impl PagedCachedFile {
...
+        // Pages a non-durable commit left in the write buffer would otherwise hold up to half the
+        // cache until the next durable commit. They can always be written out, so reclaim their
+        // space rather than evicting read cache entries for them. Done before the read cache lock
+        // is taken, to keep the ordering of write buffer before read cache.
+        let mut reclaimed = false;
+        if self.committed_pages_buffered.load(Ordering::Acquire) {
+            let read_bytes = self.read_cache_bytes.load(Ordering::Acquire);
+            let write_bytes = self.write_buffer_bytes.load(Ordering::Acquire);
+            if read_bytes + len + write_bytes > self.max_cache_size {
+                reclaimed = self.flush_buffered_pages(len)? >= len;
```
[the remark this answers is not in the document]
cberner: Agreed that an opportunistic writeback should not fail the read it is optimizing. Fixed in 2cf686a: the reclaim's result is discarded, so the read returns the page it already has and the eviction below makes room instead. `flush_lowest_priority()` reinserts a page whose write failed, so nothing is lost.

One correction to the reasoning: the database is latched either way. `CheckedBackend::write` sets `io_failed` before the error reaches this call, so the `PreviousIo` state is entered by the failed write itself, not by propagating it here. What the change buys is that this read succeeds and the error surfaces from the next write or flush -- an operation that actually required durability -- rather than from a read.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -511,7 +577,7 @@ impl PagedCachedFile {
         let write_bytes = self.write_buffer_bytes.load(Ordering::Acquire);
         let over_total = cache_size + len + write_bytes > self.max_cache_size;
         let mut removed = 0;
-        if over_total {
+        if over_total && !reclaimed {
```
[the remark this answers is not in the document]
cberner: Real bug, fixed in 2cf686a. Reproduced first with a `set_cache_size(0)` unit test (`zero_size_cache_stays_empty_after_reclaim`): it fails on the previous revision with `read_cache_bytes` at 128 after the reclaim, and passes now.

The fix removes the `reclaimed` flag rather than adding a second condition. `write_bytes` is loaded after the reclaim runs, so `over_total` already reflects the space it freed: when the reclaim covered the shortfall the check is false on its own, and when it did not — a zero-size cache, or a stripe that could not be locked — the eviction runs as before. So the guard was both redundant and, as you note, wrong.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -494,7 +530,58 @@ impl PagedCachedFile {
...
+        // is taken, to keep the ordering of write buffer before read cache.
+        //
+        // Reclaiming is best effort, so a writeback error is not propagated: the page this call
+        // returns was already read, the flushed page stays buffered, and the eviction below makes
+        // room instead. The backend records the failure, so the next write or flush reports it.
+        if self.committed_pages_buffered.load(Ordering::Acquire) {
+            let read_bytes = self.read_cache_bytes.load(Ordering::Acquire);
+            let write_bytes = self.write_buffer_bytes.load(Ordering::Acquire);
+            if read_bytes + len + write_bytes > self.max_cache_size {
+                let _ = self.flush_buffered_pages(len);
```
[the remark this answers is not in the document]
cberner: Right, and this was the substantive half of the earlier comment — I fixed the propagation and left the latch, which was the part that actually mattered. Ignoring the error made it worse, not better: the read then succeeded and the latch surfaced later as a `PreviousIo` from an unrelated operation.

CI caught it independently on 2cf686a. The fuzzer failed with `called Result::unwrap() on an Err value: PreviousIo`, exactly this: a read triggers the reclaim, the write trips the simulated I/O failure, `io_failed` latches, and the harness then hits `PreviousIo` where it expects an operation to succeed.

Fixed in 6415579 by giving `CheckedBackend` a `write_best_effort()` that does not latch, selected per call site with an explicit `Writeback::{Required, BestEffort}`, so the rule 1 spill in `write()` keeps latching while the read-path reclaim does not. A failed writeback now leaves the page buffered, the read succeeds, and the error is reported by the next real write or flush.

Verified against the fuzzer rather than by inspection: the crash reproduces locally on the pre-fix code (`crash-d66889485ed5f14f2022978bdb49775cbd28d36a`, same `PreviousIo` panic), that artifact passes with the fix, and a fresh 10-minute run over the corpus is clean -- 208,956 runs, no artifacts.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -443,11 +477,12 @@ impl PagedCachedFile {
...
-    pub(super) fn write_barrier(&self) -> Result {
-        // TODO: non-durable commits would be much faster, if this did not issues writes to disk,
-        // and instead just made the data visible to readers
-        self.flush_write_buffer()
+    // Make buffered writes visible to readers, without writing them to the file or guaranteeing
+    // durability. They stay in the buffer, against its budget, until flushed or evicted.
+    pub(super) fn write_barrier(&self) {
+        if self.write_buffer_bytes.load(Ordering::Acquire) > 0 {
+            self.committed_pages_buffered.store(true, Ordering::Release);
+        }
```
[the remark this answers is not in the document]
cberner: The description of the mechanism is right -- pages this transaction frees go into `DATA_FREED_TABLE` and are only released by `process_data_freed_pages_after_commit()`, after `mem.commit()` has flushed the buffer, so the durable commit does write pages its final roots no longer reference. But it is not a regression, because master writes those same pages, just earlier.

Measured with a counting backend: durable baseline, a `Durability::None` commit inserting 199 x 100KiB values, then a durable commit deleting all of them.

| | non-durable commit | durable delete | total |
|---|---|---|---|
| master | 207 writes / 24 MiB | 7 writes | 224 writes / 25 MiB |
| this PR | **0 writes** | 212 writes / 24 MiB | 222 writes / 25 MiB |

Identical bytes; the change only moves when they are written. Where the page *can* be dropped before the flush the saving is already realized -- `free_helper()` calls `cancel_pending_write()`, so a page freed while still buffered is never written at all, which is why the totals come out marginally lower here rather than higher.

Making the remaining case free before the flush would mean releasing pages ahead of `mem.commit()`, which is exactly what the deferred-free design avoids: readers of the old root may still need them. That is a change to commit ordering with its own correctness argument, and out of scope for this PR.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -575,12 +680,17 @@ impl PagedCachedFile {
...
         let mut flushed = 0;
         while flushed < bytes_needed {
             if let Some((offset, buffer)) = stripe.pop_lowest_priority() {
                 let removed_len = buffer.len();
-                let result = self.file.write(offset, &buffer);
+                let result = if writeback == Writeback::Required {
+                    self.file.write(offset, &buffer)
+                } else {
+                    self.file.write_best_effort(offset, &buffer)
+                };
```
cberner: I think we should use a match statement here, so that this is safe against adding another variant to the enum

cberner: Agreed — with `==` plus an `else`, a third variant would silently take the best-effort path, which is the wrong default for a write that has to reach the file. Changed to an exhaustive match in 3c71856, and dropped the now-unused `PartialEq` from the enum.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -195,6 +195,23 @@ impl CheckedBackend {
         }
         result.map_err(StorageError::from)
     }
+
+    // A write whose data does not have to reach the file: the caller keeps the page buffered for
+    // a later flush to retry. A failure must not be latched, or an optimization that nothing
+    // depends on would turn every subsequent operation into a PreviousIo error.
```
cberner: This comment seems a bit misleading. It looks like the caller keeps the page buffered in either the write() or write_best_effort() case. The difference is that if write_best_effort() fails it does not permanently fail the backend

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@
   be added behind the `experimental_cursor` feature flag, like the table cursors' were.
 
 ## 4.2.0 - 2026-XX-XX
+* Improve performance of `Durability::None` commits. They no longer write dirty pages to the file:
+  the pages are retained in the in-memory cache, where readers see them, and are written back by
+  the next durable commit or when evicted by cache pressure.
```
cberner: Say that it's about 2x faster, and don't include the implementation details

cberner (commented):
I left some comments. Otherwise lgtm

src/tree_store/page_store/cached_file.rs
cberner: You're right — `flush_lowest_priority()` reinserts the page on error regardless of which write it used, so the re-buffering is not the distinction. The only difference is whether the failure permanently fails the backend. Reworded in d134a4f, and fixed the same implication in the `Writeback` enum's doc, which said the page "stays buffered" only for `BestEffort`.

---

CHANGELOG.md
cberner: Done in d134a4f — the entry is now just `* \`Durability::None\` commits are about 2x faster.`

One caveat on the number, since it is a user-facing claim: I measure this PR on its own at roughly **1.2-1.5x**, not 2x (20k single-key non-durable commits over a 500k row table: overwrites 878ms -> 694ms, inserts 935ms -> 760ms; the `redb_benchmark` no-sync row went 10618ms -> 6914ms in one run).

2x is what the pair of changes gets. Deferring each non-durable commit's freed-page records to the next durable commit -- which I ported onto `claude/nondurable-freed-records-deferral`, stacked on this branch -- takes the same benchmark to 364ms and 394ms, about 2.7x and 2.4x overall.

So the entry is accurate for 4.2.0 if both land, which is the plan. If the follow-up slips past the release, this should come down to "about 1.5x". Happy to write it either way -- say the word and I'll adjust.

---

PR #1319, opened 2026-08-07
Fix stats() walking stale table roots while a table is open

cberner:
## The bug

A write transaction only learns a table's root when the table handle is closed: `Table::drop` stages it in `pending_table_updates`, where it waits until commit. While a handle is open, the live root exists only inside that handle, so the staged entry is the root from the table's *previous* close.

That staleness is dangerous because of how freeing works mid-transaction: pages allocated by **this** transaction are returned to the allocator immediately when freed (nothing committed can reference them), so they are eligible for reuse right away. A root staged on close is built of exactly such pages. Reopening the table and modifying it frees those pages, and the allocator may recycle them for unrelated tables — while the staged entry still points at them.

`WriteTransaction::stats()` preferred the staged root unconditionally. Calling it in that window walked recycled pages:

1. Open table X, insert 5000 rows, drop the handle → staged root R1, made of this-transaction pages
2. Reopen X, delete all rows, **keep the handle open** → R1's pages are freed immediately; the map still says R1
3. Open table X2, insert 2000 × 16KB values → the allocator recycles R1's pages for X2's leaves
4. `txn.stats()` walks R1 and parses X2's data as X's btree nodes

**Release builds** returned `Ok` with garbage numbers (the repro reported a phantom 16,008-byte row read out of a recycled page). **Debug builds** abort the process: the walk trips an internal assertion (`cached_file.rs` buffer-length check or a btree parse assert, depending on what landed in the page) while `stats()` holds the `tables` mutex; the unwind poisons the mutex, the still-open `Table`'s destructor then panics on the poisoned lock, and a panic inside a destructor during unwinding is fatal.

## The fix

`stats()` now takes the set of currently-open tables into account (the transaction already tracks them to reject double-opens): for an open table it skips the staged root and walks the **committed** root, which cannot be recycled during the transaction — pages from previous transactions are only queued for freeing at commit. The semantics are documented on the method: a table currently open for writing is reported as of the start of the transaction; tables that were modified and closed are reported with their in-progress contents, as before.

The sibling pending-update walkers (compaction's `relocate_tables` and `highest_index_pages`) only run without open tables, so `stats()` was the sole exposed path. Reporting the live root instead was considered and rejected: it would require every open table to publish root updates into shared locked state on every mutation, taxing the write hot path for a diagnostics call.

## Testing

- New regression test `stats_with_open_table` reproduces the full recycle dance and asserts exact `stored_bytes` both while the table is open (committed view) and after it closes (staged view). Without the fix it aborts under debug assertions.
- Standalone repro confirms debug and release now agree on the exact byte count where release previously reported garbage and debug aborted.
- Full test suite passes; clippy and rustfmt clean.

https://claude.ai/code/session_01EuLGiEERgJR51vhJNtYwRA

---
_Generated by [Claude Code](https://claude.ai/code/session_01EuLGiEERgJR51vhJNtYwRA)_

### DIFF
--- a/src/tree_store/table_tree.rs
+++ b/src/tree_store/table_tree.rs
@@ -549,10 +549,8 @@ impl TableTreeMut {
         new_name: &str,
         table_type: TableType,
     ) -> Result<(), TableError> {
-        // Move the definition as stored in the tree, without applying pending_table_updates:
-        // a root staged by a table close must stay out of the master tree until commit,
-        // because reopening the table can free that root's pages. stats() relies on the
-        // staged roots of open tables living only in pending_table_updates, which it skips.
+        // Move the definition as stored, not the pending update: staged roots must stay out
+        // of the master tree, since reopening the table can free their pages (see stats())
         let stored_definition = if let Some(guard) = self.tree.get(&name)? {
             let definition = guard.value();
             definition.check_match_untyped(table_type, name)?;
@@ -706,14 +704,8 @@ impl TableTreeMut {
         Ok(())
     }
 
-    // `is_open` must return true for tables that are currently open in this transaction.
-    // An open table may have staged a root here on a previous close, and then freed pages
-    // reachable from that root: pages allocated by this transaction are returned to the
-    // allocator immediately when freed, so they may already have been reallocated and
-    // rewritten with unrelated data. Walking such a root reads garbage. The committed root
-    // is always safe to walk, because pages allocated by previous transactions are only
-    // queued for freeing at commit, so for open tables we ignore the staged root and report
-    // the table as of the start of the transaction.
+    // Staged roots of tables that are currently open (per `is_open`) may reference recycled
+    // pages, so those tables are reported from their stored root: as of the transaction start.
     pub fn stats(&self, is_open: impl Fn(&str) -> bool) -> Result<DatabaseStats> {
         let master_tree_stats = self.tree.stats()?;
         let mut max_subtree_height = 0;
--- a/src/tree_store/table_tree.rs
+++ b/src/tree_store/table_tree.rs
@@ -549,7 +549,21 @@ impl TableTreeMut {
         new_name: &str,
         table_type: TableType,
     ) -> Result<(), TableError> {
-        if let Some(definition) = self.get_table_untyped(name, table_type)? {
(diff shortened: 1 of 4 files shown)


src/tree_store/table_tree.rs
```diff
@@ -711,7 +719,9 @@ impl TableTreeMut {
         for entry in self.tree.range::<RangeFull, &str>(&(..))? {
             let entry = entry?;
             let mut definition = entry.value();
-            if let Some((updated_root, length, _)) = self.pending_table_updates.get(entry.key()) {
+            if !is_open(entry.key())
+                && let Some((updated_root, length, _)) = self.pending_table_updates.get(entry.key())
```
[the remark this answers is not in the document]
cberner: Confirmed — modify/close/rename/reopen walked the stale root through the master tree entry despite the skip; a regression test on that sequence read 1,040 phantom bytes out of recycled pages.

Fixed in 4caad2a at the rename rather than in `stats()`: `rename_table()` now moves the definition as stored in the tree, without applying `pending_table_updates`, so staged roots keep living only in the pending map (which already travels to the new name and is flushed at commit). That restores the invariant the open-table skip relies on, and it also means the committed root is what `stats()` finds in the master tree for a renamed open table — falling back to "the original committed definition" per-name inside `stats()` isn't otherwise possible once the name has changed. Added `stats_with_renamed_open_table` covering the full sequence, and the existing rename-with-modifications tests still pass.

---

src/tree_store/table_tree.rs
```diff
@@ -614,6 +625,9 @@ impl TableTreeMut {
             table
         };
 
+        // Take the staged update: while the table is open its root lives in the handle
+        self.pending_table_updates.remove(name);
```
cberner: This doesn't seem like the best place for this. It's quite implicit. Can we put it closer to the actual Table construction?

cberner: Moved in b2580b1: the take now happens in `TableNamespace::inner_open` next to the `open_tables` registration, and in `open_system_table` for the system tree, via a named `clear_pending_table_update()`. `get_or_create_table` is back to a pure lookup/create. Ordering is unchanged — the take still follows the fallible type check, so `failed_reopen_preserves_staged_table` still passes.

---

cberner (commented):
Lgtm

Issue #1317, opened 2026-08-06
Bulk loading into an empty table


farhan-syah:
We bulk load graph data into redb -- a vertex table plus forward and reverse adjacency, one transaction. `insert_with_hint` fixed the space side of this; thanks for taking it. What's left is time.

Measured on f551b96, 1M vertices and 8M edges as 17M pairs across three tables, one transaction:

| | load | compacted |
|---|---|---|
| arrival order `insert()` | 24.8s | 651.55 MiB |
| sorted `insert()` | 12.8s | 901.55 MiB |
| sorted + `InsertHint::Append` | 9.9s | 456.88 MiB |
| **bottom-up build** | **1.3s** | **446.88 MiB** |

The hint gets the density. A bottom-up build gets the time: 8x faster than the best insert path, at 2.2% smaller.

The idea is the usual one -- given an empty table and a strictly ascending stream, pack the leaves, then build the branch level above them, and so on. Each page written once, no descent, no splits.

Shape I have been running with, for something concrete to react to:

```rust
impl WriteTransaction {
    pub fn build_sorted_table<K, V, F>(
        &self,
        definition: TableDefinition<K, V>,
        producer: F,
    ) -> Result<(), Error>
    where
        F: FnOnce(&mut SortedTableBuilder<K, V>) -> Result<(), Error>;

    /// As above, with an explicit packing policy
    pub fn build_sorted_table_with_options<K, V, F>(
        &self,
        definition: TableDefinition<K, V>,
        options: SortedTableOptions,
        producer: F,
    ) -> Result<(), Error>;
}

impl<K: Key, V: Value> SortedTableBuilder<K, V> {
    /// Keys must be strictly increasing
    pub fn insert(&mut self, key: ..., value: ...) -> Result<(), Error>;
}
```

Used as:

```rust
let txn = db.begin_write()?;
txn.build_sorted_table(TABLE, |builder| {
    for (key, value) in sorted_stream {
        builder.insert(key, value)?;
    }
    Ok(())
})?;
txn.commit()?;
```

The closure is so the builder cannot outlive the build. Any failure inside it, or an out-of-order key, poisons the transaction, so a partially built table can never be committed. `SortedTableOptions` currently carries one knob, a target leaf size, since packing leaves larger than a page helps scan-heavy tables.

If `SortedTableBuilder` is more public surface than you want to commit to, a narrower version would be one method taking an iterator:

```rust
txn.create_table_from_sorted_iter(TABLE, iter)?;
```

[...]

PR #1322, opened 2026-08-07
Add Table::append_sorted()

farhan-syah:
Adds `Table::append_sorted`, per the discussion in #1317.

Stacked on #1321 so the diff shows only the new work — retarget to master once that lands.

## What it does

Appending a stream of ascending keys through `insert()` descends the tree once per pair. This seeks to the end once, buffers into the existing `LeafRunRewrite`, and splices packed leaves in through `replace_leaf_children` when the buffer fills, which is the machinery `retain()` already uses.

```rust
let mut table = txn.open_table(TABLE)?;
table.append_sorted(sorted_pairs)?;
```

Keys must be strictly greater than every key already in the table. It works on a table that already holds data, not only an empty one.

## Numbers

Same benchmark as #1317, 1M vertices and 8M edges as 17M pairs across three tables in one transaction, on top of #1321:

| | load | compacted | forward leaf pages | data |
|---|---|---|---|---|
| arrival order `insert()` | 25.0s | 651.55 MiB | 78713 | 58.8% |
| sorted `insert()` | 10.3s | 456.88 MiB | 54795 | 83.7% |
| `append_sorted` | 1.7s | 453.84 MiB | 54862 | 84.3% |

Load times move with anything else running on the machine; the leaf page counts and compacted sizes are identical run to run.

## Splitting the parent branch

`replace_leaf_children` was written for `retain()` and `extract_if()`, where a branch only ever loses children, so `finalize_branch_builder` handles merging and collapsing but never splitting. Appending inverts that: each splice hands the parent a batch of leaves in place of one, so the branch grew without bound and the tree stayed two levels deep. A large enough load then panicked in `RawBranchBuilder::new`, which stores the key count as a `u16` — 500k pairs of 512 bytes was enough.

`BranchBuilder::should_split()` already existed and already accounted for that `u16` limit; the splice path simply never called it. `finalize_branch_builder` now does, returning a `SplitBranch` that the ancestor path applies like any other result, growing a new root if it reaches the top. Deletions never make a branch grow, so the new arm is unreachable for the existing callers.

Tree shape for an ascending load of 512 byte values, before and after:

| pairs | leaves | branches | height |
|---|---|---|---|
| 60,000 | 8572 | 1 → 33 | 2 → 3 |
| 500,000 | 71429 | panic → 260 | panic → 4 |

Leaf counts are unchanged, so packing is unaffected. Dropping the oversized branch pages is also where the compacted size above improves against a plain sorted load.

## Out-of-order keys

`append_sorted` originally corrupted the tree when given keys that did not ascend, since nothing checked. It now returns `StorageError::KeysNotAscending`, checked against both the previous key of the call and the table's existing maximum. Both error enums are `#[non_exhaustive]`, so the new variant is not a breaking change.

Pairs accepted before the rejected one are kept, as they would be by `insert()` in a loop. That is documented on the method rather than made atomic, since the iterator is consumed lazily and a caller wanting all-or-nothing can abort the transaction.

## Large values

A leaf holding a single value larger than a page is not absorbed into the buffer, which would copy the whole value to add one pair after it. Those fall back to an ordinary insert, matching what `replace_leaf_children` already does for the neighbouring-leaf case.

## Buffer size

`APPEND_BUFFER_BYTES` is an internal constant rather than a parameter. The cost is per splice — each one rewrites the parent chain — so what matters is the flush count, not the buffer. At 64 KiB it is 4x slower than at 1 MiB, and flat from 1 MiB out to 16 MiB.

## Testing

`cargo fmt --check`, `cargo clippy --all --all-targets -- -Dwarnings`, and `cargo test` all pass.

`tests/append_sorted_tests.rs` covers read-back at five sizes, extending a table that already has data, an empty stream, keys that do not ascend within one call, keys at or below an existing table's maximum, a stream long enough to splice several times (asserting the tree actually gained a level and more than one branch), and appending past a leaf holding an 8 MiB value. All call `check_integrity()`.

`FuzzOperation::AppendSorted` appends a run derived from the reference model's current maximum, so the fuzzer exercises the splice path against the model. `cargo fuzz run fuzz_redb -- -max_len=10000 -max_total_time=600` ran 217,725 inputs with no crashes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # redb - Changelog
 
 ## 4.2.0 - 2026-XX-XX
+* Add `Table::append_sorted()`, which loads pairs whose keys are greater than everything already
+  in the table without descending the tree for each one. Keys must be strictly increasing, and
+  `StorageError::KeysNotAscending` is returned if they are not.
 * Fix `WriteTransaction::stats()` returning garbage statistics, or panicking when debug assertions
   are enabled, when called while a table is open and modified in the same transaction.
 * Fix a deadlock when a `Database` was dropped while a `WriteTransaction` was live. A live
--- a/crates/redb-bench/Cargo.toml
+++ b/crates/redb-bench/Cargo.toml
@@ -64,3 +64,7 @@ harness = false
 [[bench]]
 name = "syscall_benchmark"
 harness = false
+
+[[bench]]
+name = "graph_load_benchmark"
+harness = false
--- a/crates/redb-bench/benches/common.rs
+++ b/crates/redb-bench/benches/common.rs
@@ -553,7 +553,7 @@ pub fn benchmark<T: BenchDatabase + Send + Sync>(
     results
 }
 
-fn database_size(path: &Path) -> u64 {
+pub fn database_size(path: &Path) -> u64 {
     let mut size = 0u64;
     for result in walkdir::WalkDir::new(path) {
         let entry = result.unwrap();
--- a/crates/redb-bench/benches/graph_load_benchmark.rs
+++ b/crates/redb-bench/benches/graph_load_benchmark.rs
@@ -1,0 +1,218 @@
+#![allow(dead_code)]
+
+use std::time::{Duration, Instant};
+
+use redb::{
+    Database, ReadableDatabase, ReadableTableMetadata, TableDefinition, TableStats,
+    WriteTransaction,
+};
+use tempfile::NamedTempFile;
+
+mod benchmark_dir;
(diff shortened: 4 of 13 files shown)


cberner:
Ah sorry, by CursorMut, I meant we should expose CursorMut as a public API, not add an append_sorted() API. I should have something you can test later today to see if it reaches the same performance as your implementation

farhan-syah:
> Ah sorry, by CursorMut, I meant we should expose CursorMut as a public API, not add an append_sorted() API. I should have something you can test later today to see if it reaches the same performance as your implementation

Okay, no problem. I'll wait and compare it later.

PR #1327, opened 2026-08-07
Add the experimental CursorMut API for bulk loading sorted data

cberner:
Fourth slice of splitting up #1323. The machinery from #1326 has merged; this now targets master directly.

The public surface over #1326's machinery, behind the `experimental_cursor` feature flag:

```rust
let mut table = txn.open_table(TABLE)?;
let mut cursor = table.upper_bound_mut(Bound::Unbounded)?;
for (key, value) in sorted_stream {
    cursor.insert_before(key, &value)?;
}
cursor.close()?;
```

- `Table::lower_bound_mut()` / `Table::upper_bound_mut()` return a `CursorMut` pointing at a gap between entries, modeled on the standard library's nightly `BTreeMap` cursors as closely as redb's data model allows.
- `insert_before(key, value)` inserts into the gap and leaves the cursor after the new entry. A key that does not sort strictly between the gap's neighbors is rejected with the new `StorageError::UnorderedKey` and nothing changes, matching std's `UnorderedKeyError`; an existing key is never overwritten.
- `peek_prev()` / `peek_next()` return the gap's neighbors as `AccessGuard`s, including pending buffered inserts, without splicing them.
- `close()` splices any pending inserts and surfaces errors. Dropping the cursor also splices; a failed splice on either path poisons the write transaction, so inserts that were reported as applied can never be silently missing from a commit.

Loading 1M sorted 150-byte pairs is ~2x faster than `insert()` at the same leaf density and half the branch pages; on #1322's 17M-pair graph load the gap is ~6x. The changelog marks the feature unstable; `insert_after`, `remove_*`, and multimap coverage are future work.

Integration tests cover bulk loads crossing the buffer threshold, large values, appends across transactions, all six bound positions, peeks over pending inserts, unordered-key rejection, drop-flush, and abort. The fuzzer gains a `CursorInsert` operation that inserts runs derived from the gap around a random key and checks the peeks against the reference model.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01Brnuq3uRWRXYn9iqKk2ZnJ)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ## 4.2.0 - 2026-XX-XX
 * Fix a panic when opening a database containing a corrupted persistent savepoint record;
   `StorageError::Corrupted` is now returned instead.
+* Add an experimental cursor API, behind the `experimental_cursor` feature flag:
+  `Table::lower_bound_mut()` and `Table::upper_bound_mut()` return a `CursorMut` pointing at a gap
+  between entries, modeled on the standard library's `BTreeMap` cursors. Its `insert_before()`
+  method makes bulk loading sorted data much faster than `insert()`. The feature is unstable and
+  may change incompatibly, or be removed, in any release.
 * Fix `WriteTransaction::stats()` returning garbage statistics, or panicking when debug assertions
   are enabled, when called while a table is open and modified in the same transaction.
 * Fix a deadlock when a `Database` was dropped while a `WriteTransaction` was live. A live
--- a/src/table.rs
+++ b/src/table.rs
@@ -1,5 +1,7 @@
 use crate::db::TransactionGuard;
 use crate::sealed::Sealed;
+#[cfg(feature = "experimental_cursor")]
+use crate::tree_store::BtreeCursorMut;
 use crate::tree_store::{
     AccessGuardMutInPlace, Btree, BtreeCursorRange, BtreeExtractIf, BtreeHeader, BtreeMut,
     MAX_PAIR_LENGTH, MAX_VALUE_LENGTH, PageAllocator, PageHint, PageNumber, PageResolver,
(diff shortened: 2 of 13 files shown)


src/table.rs
```diff
@@ -1073,3 +1157,185 @@ impl<'a, K: Key + 'static, V: Value + 'static> VacantEntry<'a, K, V> {
...
+/// around the gap returns [`AccessGuard`]s instead of references, and every
+/// operation can report a storage error.
+///
+/// A cursor makes it cheap to insert a run of keys that arrive in sorted
+/// order, such as when bulk loading a table: inserting through the gap skips
+/// the per-key descent from the root that [`Table::insert`] performs, and
+/// packs finished leaves densely. Pending inserts may be buffered internally.
+/// They are visible through the cursor's own [`peek_prev`](Self::peek_prev)
+/// and [`peek_next`](Self::peek_next), and are spliced into the table when
+/// the buffer fills and when the cursor is closed or dropped. Call
```
cberner: This describes too much internal implementation detail. Bulk loading doesn't need to be mentioned either.

src/table.rs
```diff
@@ -1073,3 +1157,185 @@ impl<'a, K: Key + 'static, V: Value + 'static> VacantEntry<'a, K, V> {
...
+            .inner
+            .insert_before(key_bytes.as_ref(), value_bytes.as_ref())
+        {
+            Ok(true) => Ok(()),
+            Ok(false) => Err(StorageError::UnorderedKey),
+            Err(err) => Err(self.latch_error(err)),
+        }
+    }
+
+    /// Closes the cursor, splicing any pending inserts into the table.
```
cberner: I don't think we need to mention splicing and pending inserts all over the place. Just focus on the user interface and not the details of how the cursor is implemented

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 ## 4.2.0 - 2026-XX-XX
 * Fix a panic when opening a database containing a corrupted persistent savepoint record;
   `StorageError::Corrupted` is now returned instead.
+* Add an experimental cursor API, behind the `experimental_cursor` feature flag:
+  `Table::lower_bound_mut()` and `Table::upper_bound_mut()` return a `CursorMut` pointing at a gap
+  between entries, modeled on the standard library's `BTreeMap` cursors. Its `insert_before()`
+  method makes bulk loading sorted data much faster than `insert()`. The feature is unstable and
```
cberner, line 9: This is a good place to mention bulk loading. Only mention it here, not in the docstrings.

Issue #794, opened 2024-04-13
ReadableTable impl for an Either<Table, ReadOnlyTable>


rklaehn:
I got some complex functions that I want to work with either a Table<'a> or a ReadOnlyTable.

The trait ReadableTable is made for this. You can have a function that takes an `impl ReadableTable<'a, K, V>`, and pass in either a Table<'a, K, V> or a ReadOnlyTable<K, V>.

However, there are some downsides to having to take an impl Trait, and it is not possible to construct a concrete type for either a Table or ReadOnlyTable. In my case I want to define a complex iterator that uses multiple tables, and having to drag around the table type is very inconvenient.

Implementing such a concrete type would be simple enough. You need some kind of Either type, and then implement ReadableTable for it assuming that both Left and Right also implement ReadableTable. But the ReadableTable trait is sealed, so this can not be done outside redb.

So either make the trait non-sealed (I suspect there are good reasons against this), or add an either type and the above impl to the crate. The same would have to be done for the [ReadableMultimapTable](https://docs.rs/redb/latest/redb/trait.ReadableMultimapTable.html#).

All in all it would have to be one either type and impls for all the "readonly view abstraction" traits in the crate.

cberner:
Actually they were just sealed because I couldn't think of a good use case for them being unsealed, and wanted to minimize the API surface area :)

rklaehn:
Thanks for the quick response!

PR #1329, opened 2026-08-08
Seal ReadableDatabase behind experimental-api-5

cberner:
Seals `ReadableDatabase`, gated on `experimental-api-5`. Rebased onto master (79614c5); one commit, two files, no changelog entry.

`ReadableDatabase` is implemented only by `Database` and `ReadOnlyDatabase`, so sealing it keeps the implementation set closed and lets methods be added without breaking callers — matching `TableHandle` and `MultimapTableHandle`.

Sealing it outright would be a source break: a wrapper outside redb can implement the trait today by delegating, even though it cannot construct a `ReadTransaction` or `CacheStats` itself. Gating it on `experimental-api-5` means 4.x keeps compiling for those wrappers, and the seal previews what redb 5 will do.

`cfg` can't be applied to a supertrait bound, so the bound itself is swapped in `sealed.rs`:

```rust
#[cfg(feature = "experimental-api-5")]
pub use self::Sealed as SealedInApi5;

#[cfg(not(feature = "experimental-api-5"))]
pub trait SealedInApi5 {}
#[cfg(not(feature = "experimental-api-5"))]
impl<T: ?Sized> SealedInApi5 for T {}
```

`ReadableDatabase: SealedInApi5` is then a single declaration whose bound is real or vacuous depending on the feature. Verified with a throwaway `struct Wrapper(Database)` that delegates `begin_read()` and `cache_stats()`:

| Build | Result |
|---|---|
| `cargo build --tests` (default) | compiles — no break |
| `cargo build --tests --features experimental-api-5` | `E0277: the trait bound Wrapper: redb::sealed::Sealed is not satisfied` |

`ReadableTableMetadata`, `ReadableTable`, and `ReadableMultimapTable` are untouched and stay unsealed, per #794.

### Testing

`just`/podman aren't available in this environment, so the equivalent checks were run directly, in **both** feature configurations since the behavior differs by flag:

| Check | Default | `--all-features` |
|---|---|---|
| `cargo clippy --all-targets` (`RUSTFLAGS=--deny warnings`) | pass | pass |
| `cargo test` (`RUST_BACKTRACE=1`, `--deny warnings`) | pass (19/19 targets) | pass (19/19 targets) |

Plus `cargo fmt --all -- --check`.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 use std::{io, thread};
 
 use crate::error::TransactionError;
-use crate::sealed::Sealed;
+use crate::sealed::{Sealed, SealedInApi5};
 use crate::transactions::{
     ALLOCATOR_STATE_TABLE_NAME, AllocatorStateKey, AllocatorStateTree, DATA_ALLOCATED_TABLE,
     DATA_FREED_TABLE, PageList, SYSTEM_FREED_TABLE, SystemTableDefinition,
@@ -354,7 +354,7 @@ impl Drop for TransactionGuard {
     }
 }
 
-pub trait ReadableDatabase {
+pub trait ReadableDatabase: SealedInApi5 {
     /// Begins a read transaction
     ///
     /// Captures a snapshot of the database, so that only data committed before calling this method
@@ -414,6 +414,8 @@ pub struct ReadOnlyDatabase {
     mem: Arc<TransactionalMemory>,
     transaction_tracker: Arc<TransactionTracker>,
 }
+
+impl Sealed for ReadOnlyDatabase {}
 
 impl ReadableDatabase for ReadOnlyDatabase {
     fn begin_read(&self) -> Result<ReadTransaction, TransactionError> {
@@ -529,6 +531,8 @@ pub struct Database {
     mem: Arc<TransactionalMemory>,
     transaction_tracker: Arc<TransactionTracker>,
 }
+
+impl Sealed for Database {}
 
 impl ReadableDatabase for Database {
     fn begin_read(&self) -> Result<ReadTransaction, TransactionError> {
--- a/src/sealed.rs
+++ b/src/sealed.rs
@@ -1,1 +1,13 @@
 pub trait Sealed {}
+
+// Seals a trait only when the redb 5 API preview is enabled. `cfg` cannot be applied to a
+// supertrait bound directly, so the bound itself is swapped: with the feature off it resolves to a
+// blanket-implemented marker, leaving the trait implementable outside redb for the rest of 4.x.
+#[cfg(feature = "experimental-api-5")]
+pub use self::Sealed as SealedInApi5;
+
+#[cfg(not(feature = "experimental-api-5"))]
+pub trait SealedInApi5 {}
+
+#[cfg(not(feature = "experimental-api-5"))]
+impl<T: ?Sized> SealedInApi5 for T {}


src/db.rs
```diff
@@ -354,7 +354,7 @@ impl Drop for TransactionGuard {
     }
 }
 
-pub trait ReadableDatabase {
+pub trait ReadableDatabase: Sealed {
```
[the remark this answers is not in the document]
cberner: You're right, and my original justification was wrong. I claimed the trait was already impossible to implement externally because `ReadTransaction` and `CacheStats` have no public constructors; that misses delegation, as you point out. I verified it both ways with a throwaway test in `tests/`:

```rust
struct Wrapper(redb::Database);
impl ReadableDatabase for Wrapper {
    fn begin_read(&self) -> Result<ReadTransaction, TransactionError> { self.0.begin_read() }
    fn cache_stats(&self) -> CacheStats { self.0.cache_stats() }
}
```

Against the PR base (82bce77) that compiles. With the seal applied it fails with `E0277: the trait bound Wrapper: redb::sealed::Sealed is not satisfied`. So this is a real source break, not a formality.

The semver point holds too. `ReadableDatabase` has existed since 3.0.0 ("Move `Database::cache_stats()` and `Database::begin_read()` to `ReadableDatabase` trait"), so it has a year of exposure, and every breaking entry in `CHANGELOG.md` so far has landed in a major section — minor releases like 2.1.0, 2.5.0, and 3.1.0 are additive and fixes only.

I've pushed 7bac227, which corrects the commit message and rewrites the changelog entry to state the break plainly instead of implying it was inert. What I have not done is retarget the release, because which section this belongs in (4.2.0 anyway, a new 5.0.0, or not at all) is a call for @cberner rather than something I should decide inside the PR.

---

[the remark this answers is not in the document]
cberner: Closing the loop on my last comment: the release call has been made, and this is now filed under a new `## 5.0.0` changelog section instead of 4.2.0 (9a479fb), which is the "keep it for a major release" outcome you suggested. The readable table traits stay unsealed, and 0560286 documents that they may gain methods without that counting as a breaking change. Resolving.

---

src/table.rs
```diff
@@ -530,6 +530,11 @@ impl<K: Key + 'static, V: Value + 'static> Debug for Table<'_, K, V> {
     }
 }
 
+/// Metadata that can be read from a table or multimap table
+///
+/// redb's own table types already implement this trait. Implementing it outside redb is only for
+/// advanced uses, such as a concrete type that abstracts over [`Table`] and [`ReadOnlyTable`] by
+/// delegating to whichever it holds; it is not needed to use redb.
```
cberner: revert this

src/table.rs
```diff
@@ -543,6 +548,11 @@ pub trait ReadableTableMetadata {
     }
 }
 
+/// A table that can be read from
+///
+/// redb's own table types already implement this trait. Implementing it outside redb is only for
+/// advanced uses, such as a concrete type that abstracts over [`Table`] and [`ReadOnlyTable`] by
+/// delegating to whichever it holds; it is not needed to use redb.
```
cberner: revert this

src/multimap_table.rs
```diff
@@ -974,6 +974,11 @@ impl<K: Key + 'static, V: Key + 'static> Drop for MultimapTable<'_, K, V> {
     }
 }
 
+/// A multimap table that can be read from
+///
+/// redb's own table types already implement this trait. Implementing it outside redb is only for
+/// advanced uses, such as a concrete type that abstracts over [`MultimapTable`] and
+/// [`ReadOnlyMultimapTable`] by delegating to whichever it holds; it is not needed to use redb.
```
cberner: revert this

src/db.rs
```diff
@@ -354,7 +354,12 @@ impl Drop for TransactionGuard {
     }
 }
 
-pub trait ReadableDatabase {
+/// A database that can be read from
+///
+/// Note: with the `experimental-api-5` feature enabled this trait is sealed, and is implemented
+/// only by [`Database`] and [`ReadOnlyDatabase`]. It is planned to be sealed unconditionally in
+/// redb 5; until then a wrapper outside redb may still implement it by delegating to one of them.
```
cberner: Remove this

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
   library's `BTreeMap` cursors, are behind the `experimental_cursor` feature flag (which enables
   `experimental-api-5`); the split lets the trait surface stabilize before the cursor's own
   methods settle.
+* Seal `ReadableDatabase` behind the `experimental-api-5` feature flag, so that it is implemented
+  only by `Database` and `ReadOnlyDatabase`. With the flag off the trait is unchanged, and a
+  wrapper outside redb may still implement it by delegating `begin_read()` and `cache_stats()` to
+  an inner database. The seal is planned to become unconditional in redb 5.
```
cberner: Remove the release notes

PR #1337, opened 2026-08-08
Take a KeyRange in the range taking methods, behind experimental-api-5

cberner:
Adds the `KeyRange` trait behind `experimental-api-5`, the flag that collects the API changes planned for redb 5. Rebased onto master now that #1339, which carried the internal refactoring, has merged.

## The problem

The range taking methods are generic over `KR: Borrow<K::SelfType<'a>>`, and `RangeFull` constrains nothing, so `..` cannot infer `KR` and every unbounded call needs a turbofish — including redb's own `iter()`, which spells it `self.range::<K::SelfType<'_>>(..)`. A pair of `Bound`s and a range of references are ambiguous for the same reason: `table.range(&0..&10)` and `extract_from_if((Bound::Excluded(k), Bound::Unbounded), f)` both need the key type named today.

## The change

Under the feature, `range()`, `range_owned()`, `retain_in()`, `extract_from_if()`, and their multimap equivalents replace that bound with a `KeyRange<'a, K>` trait:

```rust
// today
fn range<'a, KR>(&self, range: impl RangeBounds<KR> + 'a) -> Result<Range<'_, K, V>>
where KR: Borrow<K::SelfType<'a>> + 'a;
// under experimental-api-5
fn range<'a>(&self, range: impl KeyRange<'a, K>) -> Result<Range<'_, K, V>>;
```

One impl per concrete range shape — `Range`, `RangeFrom`, `RangeTo`, `RangeInclusive`, `RangeToInclusive`, `(Bound, Bound)`, `&R` — plus a standalone `RangeFull` impl. With no blanket impl over `RangeBounds`, `..` resolves to the `RangeFull` impl, which has no key type left to infer, while `2..8` and `&2..&8` still pin their bound type. The trait is sealed, since its one method returns bounds in redb's internal key encoding — which, after #1339, is what the methods convert their range to anyway.

## Compatibility

With the feature off the public signatures are unchanged: the diff adds the gated variants beside them rather than editing them, so the default build's API is identical to master's. Under the feature it is a breaking change — calls that name the key type in a turbofish must drop it, and implementations of `ReadableTable` / `ReadableMultimapTable` must update their `range()` signature (the delegating tables in the tests show both).

Only the calls that the 5.0 signature infers differently need bridging in the tests, through macros that expand to whichever form the feature selects; everything else compiles unchanged under both.

## Testing

New `range_signature_inference` test (feature-gated) covers every shape with no annotation: `..`, `2..8`, `2..=8`, `8..`, `..8`, `..=8`, `&2..&8`, `(Bound::Excluded(2), Bound::Unbounded)`, `&range`, `key.as_str()..` on a `&str` table, plus `iter()`, `retain_in()`, and `extract_from_if()`.

* `cargo test --all-features`: 369 passed, 0 failed
* `cargo test` (default features): 324 passed, 0 failed — same count as master, doctests included
* `cargo clippy --workspace --all-targets` and `--workspace --all-targets --all-features`: clean
* `cargo clippy -p redb --all-targets` in both feature configurations: clean
* `cargo fmt --all -- --check`: clean
* `cargo doc --all-features`: no new warnings
* `redb-bench` compiles; the fuzz target compiles unchanged

### DIFF
--- a/crates/redb-bench/benches/common.rs
+++ b/crates/redb-bench/benches/common.rs
@@ -1056,7 +1056,7 @@ impl BenchInserter for RedbBenchInserter<'_> {
         // returned iterator is consumed.
         let iter = self
             .table
-            .extract_from_if::<&[u8], F>(range, predicate)
+            .extract_from_if(range, predicate)
             .map_err(|_| ())?;
         Ok(RedbExtractIfIterator { iter })
     }
--- a/src/table.rs
+++ b/src/table.rs
@@ -1,9 +1,12 @@
+#[cfg(feature = "experimental-api-5")]
+use crate::KeyRange;
 use crate::db::TransactionGuard;
 use crate::sealed::Sealed;
 #[cfg(feature = "experimental-api-5")]
 use crate::tree_store::BtreeCursor;
 #[cfg(feature = "experimental_cursor")]
 use crate::tree_store::BtreeCursorMut;
+#[cfg(not(feature = "experimental-api-5"))]
 use crate::tree_store::encode_bounds;
 use crate::tree_store::{
     AccessGuardMutInPlace, Btree, BtreeCursorRange, BtreeExtractIf, BtreeHeader, BtreeMut,
@@ -17,6 +20,7 @@ use std::borrow::Borrow;
 use std::fmt::{Debug, Formatter};
 use std::marker::PhantomData;
 use std::ops::Bound;
+#[cfg(not(feature = "experimental-api-5"))]
 use std::ops::RangeBounds;
 use std::sync::{Arc, Mutex};
 use std::thread;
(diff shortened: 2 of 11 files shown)


crates/redb-bench/benches/common.rs
```diff
@@ -1053,10 +1053,11 @@ impl BenchInserter for RedbBenchInserter<'_> {
         F: for<'f> FnMut(&'f [u8], &'f [u8]) -> bool + 'a,
     {
         // redb has a native cursor-based extract_from_if that removes entries lazily as the
-        // returned iterator is consumed.
+        // returned iterator is consumed. experimental_cursor, which this crate enables, turns on
+        // experimental-api-5, whose signature infers the key type from the pair of bounds.
```
cberner: Remove these comments about the feature flag. This is internal code, so we don't need to over explain it.

cberner (commented):
It seems like there are two changes mixed together in this commit: the public API change for api-5 and also the refactoring of methods to use Bound<Vec<u8>> internally more consistently. If the latter can be split out into a separate PR in this stack, then let's do that.

crates/redb-bench/benches/common.rs
cberner: Done — the comment is back to what it was on master, and the whole bench diff is now just dropping the turbofish.

---

cberner:
Split as asked: #1339 now carries the internal refactoring — encoding the bounds once at the API boundary and giving the tree layer `range_bounds()` / `extract_from_bounds()` / `retain_in_bounds()` — with no public signature change. This PR is rebased onto it and is the `KeyRange` API change alone.

The split also shrank this one: the `Bound<Vec<u8>>` conversion the gated and ungated variants share now comes from the base PR, so what is left here is the `KeyRange` trait, the gated variants beside the existing ones, and the test bridging.

---

cberner (commented):
Add the release notes to the 5.0 section. Not the 4.2

cberner:
Moved to a new `## 5.0.0` section at the top of the changelog. Left the existing `experimental-api-5` entry from #1336 under 4.2.0, since that one is an addition available in 4.2.0 behind the flag, while this is the breaking signature change that only lands in 5.0 — say the word if you want that one moved too.

---

src/table.rs
```diff
@@ -803,6 +899,15 @@ impl<K: Key + 'static, V: Value + 'static> ReadOnlyTable<K, V> {
 
     /// This method is like [`ReadableTable::range()`], but the iterator is reference counted and keeps the transaction
     /// alive until it is dropped.
+    #[cfg(feature = "experimental-api-5")]
+    pub fn range<'a>(&self, range: impl KeyRange<'a, K>) -> Result<Range<'static, K, V>> {
```
cberner: Remove this one. This method is buggy. Look at the commit history about the _owned() variants. Don't add any of the buggy versions under the api-5 flag

cberner: Removed, and the same for `ReadOnlyMultimapTable::range()` — 993f06f describes both: their `'static` iterators hold only the page, so a concurrent writer can reclaim it out from under them. Under the flag neither inherent method exists now, leaving the `ReadableTable` / `ReadableMultimapTable` methods and the `_owned()` variants. `ReadOnlyTable::get()` has the same problem but is not a range method, so this PR does not touch it.

Two tests were pinning exactly that unsound shape — `range_arc` and `range_arc_lifetime` let the iterator outlive the table — so under the flag they now go through `range_owned()`, which keeps the transaction alive and preserves what they were testing. Noted the removal in the changelog entry.

---

cberner (commented):
lgtm

PR #1338, opened 2026-08-08
Add CursorMut::insert_after(), inserting through the gap's other side

cberner:
Modeled on the standard library&#39;s nightly `BTreeMap` cursors: `insert_after()` inserts into the gap and leaves the cursor **before** the new entry, so each accepted key becomes the gap&#39;s successor and a run of `insert_after` calls loads descending data with the same buffered splices as `insert_before`&#39;s ascending runs.

A run&#39;s pending inserts all fall on one side of the gap, and the buffer uses only end operations, per review:

- An ascending run buffers the leaf&#39;s entries before the gap and appends each `insert_before`, as before.
- A descending run prepends each `insert_after` (`push_front`), which keeps the buffer ascending since the arrivals descend; the leaf&#39;s entries before the gap join once at the flush, and `peek_prev` reads the untouched leaf at the cursor&#39;s position.
- Switching direction splices the pending run first, so mixed workloads pay one splice per switch while single-direction runs keep O(1) buffering. The public docs note this.
- Both kinds validate against the gap&#39;s live bounds and reject out-of-order keys with `UnorderedKey`; a rejection that leaves no pending inserts still closes the run.

The fuzzer&#39;s `CursorInsert` operation gains a descending mode that walks down from the gap with `insert_after`, checking the peeks against the reference model on both sides. Tests cover descending bulk loads (including across the flush threshold), mixed-direction runs in a single gap with live-bound rejections on both sides and the implicit switch splices, and the mirror of the unordered-key tests.

### Remaining std `btree_map` cursor surface not yet in redb

- `CursorMut::next()` / `prev()` -- moving the mutable cursor over an entry (redb&#39;s `CursorMut` can currently only move by inserting; the read-only `Cursor` has both).
- `CursorMut::remove_next()` / `remove_prev()` -- removing the gap&#39;s neighbors through the cursor.
- `CursorMut::as_cursor()` -- reborrowing the mutable cursor as a read-only `Cursor` at the same gap.
- `insert_before_unchecked()` / `insert_after_unchecked()` -- std&#39;s order-check-skipping variants; probably never appropriate for a database, since an unordered key would corrupt the table rather than just break iteration order.
- `CursorMutKey` / `with_mutable_key()` -- mutable access to keys; does not map to redb&#39;s serialized keys.

`Cursor` itself (`next`/`prev`/`peek_next`/`peek_prev`) and the four `lower_bound`/`upper_bound`/`lower_bound_mut`/`upper_bound_mut` constructors are all covered.

---

### DIFF
--- a/src/tree_store/btree_base.rs
+++ b/src/tree_store/btree_base.rs
@@ -720,26 +720,45 @@ impl OwnedEntryBuffer {
 
     // Appends one pair, which must be greater than every buffered entry.
     #[cfg(feature = "experimental_cursor")]
-    pub(super) fn push(&mut self, key: &[u8], value: &[u8]) {
+    pub(super) fn push_back(&mut self, key: &[u8], value: &[u8]) {
         let pair = self.store(key, value);
         self.pairs.push_back(pair);
     }
 
-    // Copies `range` of the leaf's pairs to the back of the buffer. The pairs
-    // must all be greater than the buffered entries.
+    // Prepends one pair, which must be smaller than every buffered entry.
+    #[cfg(feature = "experimental_cursor")]
+    pub(super) fn push_front(&mut self, key: &[u8], value: &[u8]) {
+        let pair = self.store(key, value);
+        self.pairs.push_front(pair);
+    }
+
+    // Copies `range` of the leaf's pairs to one end of the buffer. The pairs
+    // must all be greater than the buffered entries when `back` is true, and
+    // entirely smaller otherwise.
     #[cfg(feature = "experimental_cursor")]
     pub(super) fn extend_from_leaf_range(
         &mut self,
         accessor: &LeafAccessor<'_>,
         range: Range<usize>,
+        back: bool,
     ) {
         self.pairs.reserve(range.len());
         self.data
             .reserve(accessor.length_of_pairs(range.start, range.end));
-        for index in range {
-            let entry = accessor.entry(index).unwrap();
-            let pair = self.store(entry.key(), entry.value());
-            self.pairs.push_back(pair);
+        if back {
+            for index in range {
+                let entry = accessor.entry(index).unwrap();
+                let pair = self.store(entry.key(), entry.value());
+                self.pairs.push_back(pair);
+            }
+        } else {
+            // Prepending in reverse leaves the leaf's entries ascending at
+            // the front.
+            for index in range.rev() {
+                let entry = accessor.entry(index).unwrap();
+                let pair = self.store(entry.key(), entry.value());
+                self.pairs.push_front(pair);
+            }
         }
     }
 
@@ -755,6 +774,13 @@ impl OwnedEntryBuffer {
     pub(super) fn back(&self) -> Option<(&[u8], &[u8])> {
         self.pairs
             .back()
+            .map(|(key, value)| (&self.data[key.clone()], &self.data[value.clone()]))
+    }
+
+    #[cfg(feature = "experimental_cursor")]
+    pub(super) fn front(&self) -> Option<(&[u8], &[u8])> {
+        self.pairs
+            .front()
             .map(|(key, value)| (&self.data[key.clone()], &self.data[value.clone()]))
     }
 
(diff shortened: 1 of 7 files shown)


src/tree_store/btree_cursor.rs
```diff
@@ -342,44 +342,62 @@ impl LeafRunRewrite {
...
+    // is at the end of the tree.
     upper_key: Option<Vec<u8>>,
-    // Greatest key at or before the gap: the last insert or, before any
-    // insert, the entry preceding the gap; None at the start of the tree.
+    // Greatest key at or before the gap: the last `insert_before` or, before
+    // any, the entry preceding the gap; None at the start of the tree.
     // Inserts must sort strictly above it.
     previous_key: Option<Vec<u8>>,
     entries: OwnedEntryBuffer,
+    after_inserts: Vec<(Vec<u8>, Vec<u8>)>,
```
cberner: This looks pretty inefficient compared to OwnedEntryBuffer. And can't we reuse the entries field?

cberner (commented):
There's some refactoring cleanup stuff in here that we should land first, like `check_lengths()`. That will make this PR smaller and easier to review. Split all of those out into a preparation PR to merge first

src/tree_store/btree_cursor.rs
cberner: Reworked in 5eec8cd: the second pile is gone and both insert kinds now go into the `entries` field. `OwnedEntryBuffer` gained an `insert(index, key, value)` (its pair list is already a `VecDeque` over append-only bytes, so an interior insert is just a shift of whichever side of the index is shorter), and the run tracks the gap as `len - after_count`. The buffer stays in ascending key order the whole time, so the flush splices it as-is with no recomposition, and pure ascending runs keep their plain-append behavior. Also split the `check_lengths()` refactor out into #1340 and stacked this on it.

---

cberner: Correction to the reply above: while rebasing onto the merged #1340 this got simpler, and the current head (91081e6) no longer has the interior `insert(index)` / `after_count` bookkeeping described there. The run is direction-tagged instead: both insert kinds share the one `entries` buffer, an ascending run appends and a descending run prepends (both end operations on the `VecDeque`), and switching direction splices the pending run first, so the buffer stays in ascending order with no index tracking. The PR description matches the final form.

---

src/tree_store/btree_base.rs
```diff
@@ -725,21 +725,40 @@ impl OwnedEntryBuffer {
         self.pairs.push_back(pair);
     }
 
-    // Copies `range` of the leaf's pairs to the back of the buffer. The pairs
-    // must all be greater than the buffered entries.
+    // Prepends one pair, which must be smaller than every buffered entry.
+    #[cfg(feature = "experimental_cursor")]
+    pub(super) fn push_front(&mut self, key: &[u8], value: &[u8]) {
```
cberner, line 730: Rename the other one push_back() so that it's clearer

src/tree_store/btree_cursor.rs
```diff
@@ -367,15 +382,29 @@ struct InsertRun {
...
     fn buffered_previous(&self) -> Option<(&[u8], &[u8])> {
-        self.entries.back()
+        match self.direction {
+            RunDirection::Ascending => self.entries.back(),
+            RunDirection::Descending => None,
+        }
+    }
+
+    // The pending insert nearest the gap on the entry-after side, if any.
+    fn pending_next(&self) -> Option<(&[u8], &[u8])> {
```
cberner: The naming here should match with the other method

cberner (commented):
Check the rest of the diff for naming issues like the ones I flagged

cberner:
Otherwise, lgtm

src/tree_store/btree_base.rs
cberner: Renamed to `push_back()` in 3d97f46, so the buffer's mutators mirror its `back()`/`front()` peeks.

---

src/tree_store/btree_cursor.rs
cberner: Renamed to `buffered_next()` to match `buffered_previous()` in 3d97f46. Sweeping the rest of the diff turned up one more off-vocabulary name: `InsertRun.upper_key` amid the run's otherwise previous/next naming — now `opening_next_key`, so `next_key()` reads as `buffered_next().or(opening_next_key)`. The remaining pairs already mirror: `front()`/`back()`, `ensure`/`open`/`flush_insert_run`, `resume_before`/`resume_after` against `Position::Before`/`After`, and the `insert_before`/`insert_after` plumbing, docs, and tests.

---

PR #1339, opened 2026-08-08
Route the range taking methods through encoded bounds

cberner:
Split out of #1337 as its base, per review: this is the internal refactoring on its own, with no public API change.

Every range taking path already reduced its range to `Bound<Vec<u8>>` before touching the tree, but each did it separately, and the `KR`-generic range type was threaded several layers deep to get there.

This encodes once at the API boundary instead:

* `encode_bounds()` and the byte-level `bounds_are_empty()` — which replaces `range_is_empty()` — now sit next to each other in `btree_iters`.
* `range_bounds()`, `extract_from_bounds()`, and `retain_in_bounds()` take the encoded bounds; `BtreeCursorRange::from_bounds()` backs the first.
* The public methods convert and call them, so the whole-table methods (`retain()`, `extract_if()`, and the `Table` side of `range()`) pass a pair of `Unbounded` instead of routing an unbounded range back through the ranged methods and naming the key type in a turbofish to do it.
* `BtreeCursorRange::new()` and the `KR`-generic `Btree::range()` stay for the internal callers that pass concrete ranges.

No public signatures change — the only added items are private `*_in_bounds` helpers — and no allocation is added, since the bounds were already encoded to `Vec<u8>` on every call.

## Testing

* `cargo test --all-features`: 368 passed, `cargo test`: 324 passed — both identical to master
* `cargo clippy -p redb --all-targets` and `--all-targets --all-features`: clean
* `cargo clippy --workspace --all-targets --all-features`: clean
* `cargo fmt --all -- --check`: clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01JWixTK4bUrh4ANWstPeGVG)_

### DIFF
--- a/src/multimap_table.rs
+++ b/src/multimap_table.rs
@@ -2,6 +2,7 @@ use crate::db::TransactionGuard;
 use crate::multimap_table::DynamicCollectionType::{Inline, SubtreeV2};
 use crate::sealed::Sealed;
 use crate::table::{OwnedAccessGuard, ReadableTableMetadata, TableStats};
+use crate::tree_store::encode_bounds;
 use crate::tree_store::{
     AllPageNumbersBtreeIter, BRANCH, Btree, BtreeCursorRange, BtreeHeader, BtreeMut,
     DynamicCollection, DynamicCollectionType, LEAF, LeafAccessor, MAX_PAIR_LENGTH,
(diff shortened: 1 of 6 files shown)


cberner (commented):
lgtm

PR #1340, opened 2026-08-08
Extract the cursor's key and value length checks into a helper

cberner:
Preparation refactors split out of #1338, per review, so `insert_after()` lands as pure new functionality. All behavior-preserving:

- `CursorMut::insert_before`'s key and value length checks move into a `check_lengths()` helper the second insert method will share.
- The insert run names the gap's neighbors: `buffered_previous()` is the buffered entry nearest the gap (today the buffer's back), and `next_key()` is the gap's live upper bound (today it never moves). `insert_after` will change both.

The positional `OwnedEntryBuffer::insert(index)` from the earlier revision is gone, per review: the buffer keeps only end operations, and #1338 now flushes when the insert direction switches instead of shifting.

The existing cursor tests pass unchanged.

---

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -1490,17 +1490,7 @@ impl<'a, K: Key + 'static, V: Value + 'static> CursorMut<'a, K, V> {
         self.check_usable()?;
         let key_bytes = K::as_bytes(key.borrow());
         let value_bytes = V::as_bytes(value.borrow());
-        let key_len = key_bytes.as_ref().len();
-        let value_len = value_bytes.as_ref().len();
-        if value_len > MAX_VALUE_LENGTH {
-            return Err(StorageError::ValueTooLarge(value_len));
-        }
-        if key_len > MAX_VALUE_LENGTH {
-            return Err(StorageError::ValueTooLarge(key_len));
-        }
-        if value_len + key_len > MAX_PAIR_LENGTH {
-            return Err(StorageError::ValueTooLarge(value_len + key_len));
-        }
+        Self::check_lengths(key_bytes.as_ref(), value_bytes.as_ref())?;
         match self
             .inner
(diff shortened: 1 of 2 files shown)


src/tree_store/btree_base.rs
```diff
@@ -718,11 +718,13 @@ impl OwnedEntryBuffer {
...
-    // Appends one pair, which must be greater than every buffered entry.
+    // Inserts one pair at `index`, which must preserve the buffer's ascending
+    // key order. Costs a shift of whichever side of `index` is shorter; the
+    // data bytes always append.
     #[cfg(feature = "experimental_cursor")]
-    pub(super) fn push(&mut self, key: &[u8], value: &[u8]) {
+    pub(super) fn insert(&mut self, index: usize, key: &[u8], value: &[u8]) {
         let pair = self.store(key, value);
-        self.pairs.push_back(pair);
+        self.pairs.insert(index, pair);
```
cberner: The performance of this looks unacceptable. We should just limit the usage to pushing front and back. And if we need to flush when switching between insert types or something, that is fine. But I don't want any linear scaling like this

cberner: Done: the positional insert is gone from this PR (only `check_lengths()` and the gap-neighbor naming remain), and #1338 (reworked in 443efe0) now uses only end operations. A run's pending inserts all fall on one side of the gap -- ascending runs append as before, descending runs prepend (`push_front`), which keeps the buffer ascending since `insert_after` arrivals descend -- and switching direction splices the pending run first, as you suggested. Mixed workloads pay one splice per switch; single-direction runs keep O(1) buffering. A descending run's buffer holds only the inserts, so the leaf's entries before the gap are prepended once at the flush, and its `peek_prev` reads the untouched leaf at the cursor's position.

---

cberner:
Lgtm

PR #1343, opened 2026-08-09
Remove the broken read-only get() accessors behind experimental-api-5

cberner:
Follow-up to #1306; companion to #1307, which adds deprecation notices for default-feature builds.

## Change

The inherent `ReadOnlyTable::get()` and `ReadOnlyMultimapTable::get()` return `'static` guards and iterators that, contrary to their documentation, do not keep the read transaction alive: holding one after dropping the `ReadTransaction` lets a concurrent writer reclaim the referenced pages, which panics the writer's `commit()` on a debug assertion.

redb 5 will drop them in favor of the `Readable*` trait accessors (zero cost, lifetime-bound) and the `_owned()` variants from #1306. This PR previews that removal under `experimental-api-5`, the same way the `KeyRange` change already removed the inherent `range()` methods under the flag: the two `get()` methods do not exist when it is enabled, and calls resolve to the trait accessor instead.

Supporting changes:

- `get_owned()` previously delegated to the removed methods; it now goes through the tree directly / a private `get_inner()` helper, matching how `range_owned` routes through the private `range_in_bounds()`.
- Tests that read through the inherent methods incidentally keep their `get()` calls, which resolve to the non-`'static` trait accessor under the feature. Files that didn't have `ReadableTable` in scope gain a cfg-gated import; the redb-derive tests, which cannot cfg on redb's features, import it under `#[allow(unused_imports)]` since the inherent method shadows it on default features. `get_arc_lifetime`, which exercises the removed behavior itself, switches to `get_owned()` under the feature like `range_arc` already does.
- Since CI builds with `--all-features`, CI no longer compiles or runs the two methods after this change; they remain in default-feature builds until redb 5.

## Testing

- `cargo test --all --all-features` passes; the arc-lifetime tests also pass under default features
- clippy (default and `--all-features`), fmt, and `cargo doc` are clean under `--deny warnings`

https://claude.ai/code/session_01YCYPuM4nb94pF36ZX7R8sa

### DIFF
--- a/crates/redb-derive/tests/derive_tests.rs
+++ b/crates/redb-derive/tests/derive_tests.rs
@@ -1,3 +1,5 @@
+#[allow(unused_imports)]
+use redb::ReadableTable;
 use redb::{Database, Key, ReadableDatabase, TableDefinition, TableError, TypeName, Value};
 use redb_derive::{Key, Value};
 use std::fmt::Debug;
--- a/src/multimap_table.rs
+++ b/src/multimap_table.rs
@@ -1133,7 +1133,15 @@ impl<K: Key + 'static, V: Key + 'static> ReadOnlyMultimapTable<K, V> {
 
     /// This method is like [`ReadableMultimapTable::get()`], but the iterator is reference counted and keeps the transaction
     /// alive until it is dropped.
+    #[cfg(not(feature = "experimental-api-5"))]
     pub fn get<'a>(&self, key: impl Borrow<K::SelfType<'a>>) -> Result<MultimapValue<'static, V>> {
+        self.get_inner(key)
+    }
+
+    fn get_inner<'a>(
+        &self,
+        key: impl Borrow<K::SelfType<'a>>,
+    ) -> Result<MultimapValue<'static, V>> {
         let iter = if let Some(collection) = self.tree.get(key.borrow())? {
             MultimapValue::from_collection(
                 collection,
@@ -1164,7 +1172,7 @@ impl<K: Key + 'static, V: Key + 'static> ReadOnlyMultimapTable<K, V> {
         key: impl Borrow<K::SelfType<'a>>,
     ) -> Result<OwnedMultimapValue<V>> {
         Ok(OwnedMultimapValue::new(
-            self.get(key)?,
+            self.get_inner(key)?,
             self.transaction_guard.clone(),
         ))
     }
--- a/tests/backward_compatibility.rs
+++ b/tests/backward_compatibility.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "experimental-api-5")]
+use redb::ReadableTable;
 use redb::{ReadableDatabase, ReadableTableMetadata};
 
 const ELEMENTS: usize = 3;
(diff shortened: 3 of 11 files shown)


src/multimap_table.rs
```diff
@@ -1133,7 +1133,18 @@ impl<K: Key + 'static, V: Key + 'static> ReadOnlyMultimapTable<K, V> {
 
     /// This method is like [`ReadableMultimapTable::get()`], but the iterator is reference counted and keeps the transaction
     /// alive until it is dropped.
+    ///
+    /// This method does not exist when the `experimental-api-5` feature is enabled, and will be
+    /// removed in redb 5. Use [`ReadableMultimapTable::get()`] or [`Self::get_owned()`] instead.
```
cberner: Don't add this

tests/check_integrity_nondurable.rs
```diff
@@ -133,7 +133,7 @@ fn check_integrity_does_not_promote_disk_corrupt_nondurable_commit() {
     let read = reopened.begin_read().unwrap();
     let table = read.open_table(TABLE).unwrap();
     assert_eq!(table.len().unwrap(), 5);
-    assert!(table.get(&999u64).unwrap().is_none());
+    assert!(table.get_owned(&999u64).unwrap().is_none());
```
cberner: A lot of these don't need to be converted -- they can just use the non-'static version

cberner (commented):
Apply my comments to all the other locations too. Also, don't we need to remove the broken range() method?

cberner:
Applied both comments to all locations in 335a211: the doc notes are gone from both `get()` methods, and the tests keep plain `get()` calls that resolve to the non-`'static` trait accessor under the feature (via a cfg-gated `ReadableTable` import; the redb-derive tests use explicit `ReadableTable::get(&table, ...)` since they can't cfg on redb's features).

On `range()`: it's already removed under the flag — the KeyRange change (76365a3) put `#[cfg(not(feature = "experimental-api-5"))]` on the inherent `ReadOnlyTable::range()` and `ReadOnlyMultimapTable::range()`. That's why this PR shrank to the `get()` half after rebasing; title and description updated to match.

---

crates/redb-derive/tests/derive_tests.rs
```diff
@@ -86,7 +88,7 @@ fn test_value_helper<V: Value + 'static>(
 
     let read_txn = db.begin_read().unwrap();
     let table = read_txn.open_table(table_def).unwrap();
-    let retrieved = table.get(1).unwrap().unwrap();
+    let retrieved = ReadableTable::get(&table, 1).unwrap().unwrap();
```
cberner: Why is this explicit syntax necessary?? I don't like it

cberner: It isn't, dropped in 4a0d673 — plain `table.get(...)` everywhere now. The reason it existed: redb-derive can't `cfg` on redb's features, and on default features the inherent `'static` `get()` shadows the trait method, leaving a plain `use redb::ReadableTable` unused (which `-Dwarnings` rejects). The import now carries `#[allow(unused_imports)]` with a comment instead, and the call sites stay untouched.

---

tests/backward_compatibility.rs
```diff
@@ -1,4 +1,8 @@
 use redb::{ReadableDatabase, ReadableTableMetadata};
+// With experimental-api-5 the inherent 'static get() does not exist, so calls resolve to the
+// trait method
```
cberner: Remove this. And in the other imports too

PR #1346, opened 2026-08-09
Add a thread-local cache for hot btree pages

cberner:
## Problem

Multithreaded random reads scale well below linearly. Every page lookup through the shared page cache performs three writes to memory shared between threads: acquiring the stripe `RwLock`, storing the LRU second-chance flag, and incrementing the page's `Arc` refcount. For the hottest pages — the top of each btree — those cache lines ping-pong between cores, and per-thread throughput degrades as threads are added.

## Change

Each thread keeps a small bounded cache (`ThreadLocalPageCache`) of private copies of the branch pages **one level below** the roots it reads, and lookups in read transactions serve those pages from it. Hits touch only thread-local memory, so the implementation needs no locks or atomics at all: a `PageNumberHashMap<Rc<[u8]>>` behind a `RefCell`, plain single-threaded code.

**Why only one level.** A level holds roughly a page's worth of children per page of the level above it, so the level below the root is small enough for a snapshot's pages to fit in the budget, while the next one down is already far too large. On the `lmdb_benchmark` shape (5M rows, 24-byte keys) the tree is height 4 with ~15 branch pages at depth 1 and ~4050 at depth 2. Caching depth 2 against a 128-page budget misses ~97% of the time, and admitting a page copies it, so it paid a 4KiB copy on nearly every lookup — far more than the shared-cache lookup it saved. An earlier revision of this PR cached two levels and **regressed** reads by 14-25%, which is what @cberner's benchmark caught. For the same reason the cache stops admitting once full rather than replacing entries: freezing is self-limiting under thrash, replacing is not.

**Correctness.** A thread-local cache outlives transactions and page numbers are recycled, so every entry is tagged with (database instance id, snapshot generation) and the tag is validated on lookup. Generation is the read transaction's id, and bytes of a page number cannot change within a generation: snapshots are copy-on-write, and a freed page is only reused by a later write transaction whose commit advances the generation. In-place reload is the exception — rolling a non-durable commit back rewinds the id, so later commits reuse ids that named different data — so `clear_cache_and_reload()` bumps the instance id, invalidating every thread's entries at once. Reading a different snapshot empties the cache, so a stale entry is never returned. Write transactions never use the cache (they can free and reallocate pages within one generation); leaf pages are never cached (`AccessGuard`s need real `PageImpl`s). Reentrant lookups (pathological `Key::compare` implementations) and thread teardown degrade to uncached reads rather than panicking.

**Memory.** At most 128 order-zero pages per thread with a live read transaction (512KiB at 4KiB pages), further bounded by the configured cache size so a small `set_cache_size` shrinks it and `set_cache_size(0)` disables it. Released when the read transaction is dropped, so a thread that reads once does not retain them for its lifetime.

**`no_std`.** `thread_local!` is std-only, so `no_std` builds compile the cache out and their lookups are unchanged. For the same reason the instance-id counter is a `Mutex<u64>` rather than an `AtomicU64`: 64-bit atomics are unavailable on `thumbv7em-none-eabihf`, and it is taken only when a database is opened or reloaded.

## Results

`lmdb_benchmark` on a 16-core 9950X3D, measured by @cberner against master `5e6ad02`:

| | master | this PR | |
|---|---|---|---|
| random reads (1 thread) | 1.18M | 1.25M | 1.06x |
| random range reads | 531K | 534K | — |
| random reads (4 threads) | 4.44M | 4.64M | 1.05x |
| random reads (8 threads) | 7.61M | 8.61M | 1.13x |
| random reads (16 threads) | 13.2M | 15.8M | **1.20x** |
| random reads (32 threads) | 17.4M | **27.4M** | **1.57x** |

The gain grows with thread count, which is the shape the design predicts. Write benchmarks are unchanged within noise; write transactions never consult or populate the cache. One caveat: individual writes read 1.13K vs 982 txn/s across the two runs, but those are fsync-bound at ~1ms per transaction and sit on a path this change does not touch, so that looks like disk variance rather than a regression.

Shared-page-cache reads per lookup drop from 3.00 to 2.00.

## Verification

- `cargo fmt --check`, `cargo clippy --all-targets --all-features` with `-D warnings`, and `cargo test --all-features` all pass (podman sandbox unavailable in this environment, so the `just test` steps were run directly; `cargo deny` not run). `cargo check --target thumbv7em-none-eabihf --no-default-features --features experimental-api-5[,logging,experimental_cursor]` also passes with `-D warnings`.
- Unit tests in `thread_local_page_cache.rs`: tag validation including cross-database, snapshot change empties the cache, the page bound holds, and `clear()` releases entries.
- Integration tests, each asserting `tree_height() > 2` so the tree actually has cacheable branch pages below its root: snapshot stability of a long-lived reader while writers rewrite the tree, alternating short reads and writes on one thread (generation invalidation), a small `set_cache_size` bounding the per-thread cache, and two databases open on one thread not sharing entries. `concurrent_reads_and_writes` in `multithreading_tests.rs` gained the same height guard.
- Fuzzing (`cargo fuzz run --sanitizer=none fuzz_redb`, debug assertions on): 196,920 runs on this design, zero failures and no crash artifacts, plus ~660k runs across earlier revisions of the branch. The harness interleaves reads with commits, savepoints and reopens on one thread whose cache persists throughout, checked against a reference map.

Tunables to sanity-check in review: `MAX_THREAD_LOCAL_CACHE_PAGES = 128` (page_manager.rs) and `MAX_CACHED_DEPTH = 1` (btree.rs). Not covered by design: `range()`/iterators, `first()`/`last()`, and single-page tables (the guard clone on a root-leaf remains).

Open question: the 128-page bound is **per thread**, so N concurrently-reading threads can hold N x 512KiB beyond the configured cache size. That is documented on `thread_local_cache_pages()` but not enforced; making it an aggregate bound would need either a guessed thread count or an atomic total, both of which cost something. Left as-is pending your call.

https://claude.ai/code/session_01NMLXF7RTMhhhp1jhL9XkNi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,9 @@
   optional, and lets container types holding that key store shorter separators.
 * Shorten separators for array keys, when the element type is variable width. Tables with such
   keys use slightly less space, and lookups are faster.
+* Improve multithreaded read scaling. Each thread now caches the level below the root of the
+  btrees it reads, removing lock and reference-count contention on those hot pages. Random reads
+  are about 1.2x faster with 16 threads and 1.6x faster with 32.
 * Fix a crash shortly after a commit being able to silently roll that commit back during
   recovery, if `check_integrity()` had previously repaired the database.
 * Fix iterators silently omitting data when iteration continues after an error. An iterator
--- a/src/db.rs
+++ b/src/db.rs
@@ -5,6 +5,7 @@ use crate::tree_store::ReadOnlyBackend;
 use crate::tree_store::{
     AllocationPolicy, BtreeHeader, InternalTableDefinition, PAGE_SIZE, PageHint, PageNumber,
     PageResolver, ShrinkPolicy, TableTree, TableType, TransactionalMemory,
+    release_thread_local_page_cache,
 };
 use crate::types::{Key, Value};
 use crate::{
@@ -341,6 +342,10 @@ impl TransactionGuard {
             }
         }
     }
+
+    pub(crate) fn is_read(&self) -> bool {
+        matches!(self, Self::Read { .. })
+    }
 }
 
 impl Drop for TransactionGuard {
@@ -349,7 +354,12 @@ impl Drop for TransactionGuard {
             Self::Read {
                 tracker,
                 transaction_id,
-            } => tracker.deallocate_read_transaction(*transaction_id),
+            } => {
+                // Clear the thread local cache, so that it doesn't leak cache space if this
+                // thread doesn't run another transaction
+                release_thread_local_page_cache();
+                tracker.deallocate_read_transaction(*transaction_id);
+            }
             Self::Write {
                 tracker,
                 transaction_id,
--- a/src/tree_store/btree.rs
+++ b/src/tree_store/btree.rs
@@ -11,13 +11,16 @@ use crate::tree_store::btree_cursor::BtreeCursorMut;
 use crate::tree_store::btree_cursor::{CursorMut, Position};
 use crate::tree_store::btree_iters::{bounds_are_empty, encode_bounds};
 use crate::tree_store::btree_mutator::MutateHelper;
-use crate::tree_store::page_store::{Page, PageImpl, PageMut};
+use crate::tree_store::page_store::{
+    CacheTag, Page, PageImpl, PageMut, ThreadLocalPageCache, with_thread_local_page_cache,
+};
 use crate::tree_store::{
     AccessGuardMutInPlace, AllPageNumbersBtreeIter, BtreeCursorRange, BtreeExtractIf,
     PageAllocator, PageHint, PageNumber, PageNumberHashMap, PageResolver, PageTracker,
 };
 use crate::types::{Key, MutInPlaceValue, Value};
 use crate::{AccessGuard, Result, StorageError};
+use alloc::rc::Rc;
 use alloc::string::ToString;
 use alloc::sync::Arc;
 use alloc::vec;
@@ -1005,11 +1008,40 @@ impl RawBtree {
     }
 }
 
+// A page visited during a lookup: the borrowed root, one from the shared page cache, or this
+// thread's private copy of a branch page
+enum NodePage<'a> {
+    Root(&'a PageImpl),
+    Shared(PageImpl),
+    Local(PageNumber, Rc<[u8]>),
+}
+
+impl Page for NodePage<'_> {
+    fn memory(&self) -> &[u8] {
+        match self {
+            Self::Root(page) => page.memory(),
+            Self::Shared(page) => page.memory(),
+            Self::Local(_, bytes) => bytes,
+        }
+    }
+
+    fn get_page_number(&self) -> PageNumber {
+        match self {
+            Self::Root(page) => page.get_page_number(),
+            Self::Shared(page) => page.get_page_number(),
+            Self::Local(page_number, _) => *page_number,
+        }
+    }
+}
+
 pub(crate) struct Btree<K: Key + 'static, V: Value + 'static> {
     mem: PageResolver,
     transaction_guard: Arc<TransactionGuard>,
     // Cache of the root page to avoid repeated lookups
     cached_root: Option<PageImpl>,
+    // None for write transactions, which can free and reallocate pages within one generation
+    cache_tag: Option<CacheTag>,
+    cache_max_pages: usize,
     root: Option<BtreeHeader>,
     hint: PageHint,
     _key_type: PhantomData<K>,
@@ -1028,10 +1060,20 @@ impl<K: Key, V: Value> Btree<K, V> {
         } else {
             None
         };
+        let cache_tag = if guard.is_read() {
+            Some(CacheTag {
+                instance: mem.cache_instance_id(),
+                generation: guard.id().raw_id(),
+            })
+        } else {
+            None
+        };
         Ok(Self {
+            cache_max_pages: mem.thread_local_cache_pages(),
             mem,
             transaction_guard: guard,
             cached_root,
+            cache_tag,
             root,
             hint,
             _key_type: PhantomData,
@@ -1084,38 +1126,88 @@ impl<K: Key, V: Value> Btree<K, V> {
     }
 
     pub(crate) fn get(&self, key: &K::SelfType<'_>) -> Result<Option<AccessGuard<'static, V>>> {
-        if let Some(ref root_page) = self.cached_root {
-            self.get_helper(root_page, K::as_bytes(key).as_ref())
+        let Some(ref root_page) = self.cached_root else {
+            return Ok(None);
+        };
+        let key_bytes = K::as_bytes(key);
+        let query = key_bytes.as_ref();
+        if let Some(tag) = self.cache_tag {
+            with_thread_local_page_cache(|cache| {
+                self.get_helper(&mut cache.map(|cache| (cache, tag)), root_page, query)
+            })
         } else {
-            Ok(None)
+            self.get_helper(&mut None, root_page, query)
         }
     }
 
     // Returns the value for the queried key, if present
-    fn get_helper(&self, page: &PageImpl, query: &[u8]) -> Result<Option<AccessGuard<'static, V>>> {
-        let mut descended: Option<PageImpl> = None;
+    fn get_helper(
+        &self,
+        cache: &mut Option<(&mut ThreadLocalPageCache, CacheTag)>,
+        root: &PageImpl,
+        query: &[u8],
+    ) -> Result<Option<AccessGuard<'static, V>>> {
+        let mut node = NodePage::Root(root);
+        // Counting the root's children as 1
+        let mut child_depth = 1;
         for _ in 0..MAX_BTREE_DEPTH {
-            let page = descended.as_ref().unwrap_or(page);
-            let child_page = match page.memory()[0] {
+            let child_number = match node.memory()[0] {
                 LEAF => {
                     let accessor =
-                        LeafAccessor::new(page.memory(), K::fixed_width(), V::fixed_width());
-                    return Ok(accessor.find_key::<K>(query).map(|entry_index| {
-                        let (start, end) = accessor.value_range(entry_index).unwrap();
-                        AccessGuard::with_page(page.clone(), start..end)
-                    }));
+                        LeafAccessor::new(node.memory(), K::fixed_width(), V::fixed_width());
+                    let Some(entry_index) = accessor.find_key::<K>(query) else {
+                        return Ok(None);
+                    };
+                    let (start, end) = accessor.value_range(entry_index).unwrap();
+                    let page = match node {
+                        NodePage::Root(page) => page.clone(),
+                        NodePage::Shared(page) => page,
+                        NodePage::Local(..) => unreachable!("only branch pages are cached"),
+                    };
+                    return Ok(Some(AccessGuard::with_page(page, start..end)));
                 }
                 BRANCH => {
-                    let accessor = BranchAccessor::new(page, K::fixed_width());
+                    let accessor = BranchAccessor::new(&node, K::fixed_width());
                     accessor.child_for_key::<K>(query).1
                 }
                 _ => unreachable!(),
             };
-            descended = Some(self.mem.get_page(child_page, self.hint)?);
+            node = self.get_page_for_read(cache, child_number, child_depth)?;
+            child_depth += 1;
         }
         Err(StorageError::Corrupted(
             "Btree exceeded maximum depth".to_string(),
         ))
+    }
+
+    // Levels below the root served from the thread local cache. Only the first, because a level
+    // holds roughly a page's worth of children per page of the level above it: the one below the
+    // root is small enough to cache whole, and the next is already too large. Caching a level
+    // that does not fit is worse than not caching it, since each miss copies a page
+    const MAX_CACHED_DEPTH: u32 = 1;
+
+    // Fetches a page during a lookup, from this thread's cache when it holds it. Only branch
+    // pages are cached, because an AccessGuard returned for a leaf holds the PageImpl
+    fn get_page_for_read<'a>(
+        &'a self,
+        cache: &mut Option<(&mut ThreadLocalPageCache, CacheTag)>,
+        page_number: PageNumber,
+        depth: u32,
+    ) -> Result<NodePage<'a>> {
+        if depth <= Self::MAX_CACHED_DEPTH
+            && page_number.page_order == 0
+            && let Some((cache, tag)) = cache
+        {
+            if let Some(bytes) = cache.get(*tag, page_number) {
+                return Ok(NodePage::Local(page_number, bytes));
+            }
+            let page = self.mem.get_page(page_number, self.hint)?;
+            if page.memory()[0] == BRANCH {
+                cache.insert(*tag, self.cache_max_pages, page_number, page.memory());
+            }
+            return Ok(NodePage::Shared(page));
+        }
+        Ok(NodePage::Shared(self.mem.get_page(page_number, self.hint)?))
     }
 
     pub(crate) fn first(
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -28,6 +28,28 @@ use core::marker::PhantomData;
 use core::mem;
 use core::sync::atomic::{AtomicBool, Ordering};
 
+// Assigns every TransactionalMemory, and every in-place reload of one, an id that no other
+// instance in this process has held, which is what lets a thread local page cache tell whose
+// pages it is holding. The id must be unique process wide, not just within one database:
+// thread local caches outlive the databases that filled them, and every database a thread
+// reads shares one cache, so an id unique only within a database would let one database's
+// pages match another's tag. Ids are never reused -- incrementing a u64 once per instance and
+// per reload cannot wrap in practice -- so a tag from a dropped database matches nothing.
+//
+// A mutex rather than an atomic, because 64 bit atomics are unavailable on some of the targets
+// redb supports, and this is taken only when a database is opened or reloaded.
+static NEXT_CACHE_INSTANCE_ID: Mutex<u64> = Mutex::new(1);
+
+fn next_cache_instance_id() -> u64 {
+    let mut next = NEXT_CACHE_INSTANCE_ID.lock().unwrap();
+    let id = *next;
+    *next += 1;
+    id
+}
+
+// Enough for the top levels of a large btree. 512KiB with 4KiB pages
+const MAX_THREAD_LOCAL_CACHE_PAGES: usize = 128;
+
 // The region header is optional in the v3 file format
 // It's an artifact of the v2 file format, so we initialize new databases without headers to save space
 const NO_HEADER: u32 = 0;
@@ -102,6 +124,14 @@ impl PageResolver {
 
     pub(crate) fn get_page(&self, page_number: PageNumber, hint: PageHint) -> Result<PageImpl> {
         self.mem.get_page(page_number, hint)
+    }
+
+    pub(crate) fn cache_instance_id(&self) -> u64 {
+        self.mem.cache_instance_id
+    }
+
+    pub(crate) fn thread_local_cache_pages(&self) -> usize {
+        self.mem.thread_local_cache_pages()
     }
 
     pub(crate) fn count_allocated_pages(&self) -> Result<u64> {
@@ -500,6 +530,10 @@ impl UnpersistedState {
 }
 
 pub(crate) struct TransactionalMemory {
+    // Identifies this instance's pages in the thread local cache. A transaction id alone is
+    // not enough, because a reload can rewind it and later commits then reuse ids that named
+    // different data
+    cache_instance_id: u64,
     unpersisted: Mutex<UnpersistedState>,
     storage: PagedCachedFile,
     state: Mutex<InMemoryState>,
@@ -650,6 +684,7 @@ impl TransactionalMemory {
         assert!(page_size >= DB_HEADER_SIZE);
 
         Ok(Self {
+            cache_instance_id: next_cache_instance_id(),
             unpersisted: Mutex::new(UnpersistedState::default()),
             storage,
             state: Mutex::new(state),
@@ -741,6 +776,9 @@ impl TransactionalMemory {
         // leave cached pages that disagree with the file.
         self.storage.discard_write_buffer();
         self.storage.invalidate_cache_all();
+        // Invalidate the thread local caches too, since the state being loaded can give a
+        // transaction id already read under this instance different contents
+        self.cache_instance_id = next_cache_instance_id();
         self.storage.sync_file()?;
 
         let header_bytes = self.storage.read_direct(0, DB_HEADER_SIZE)?;
@@ -1666,6 +1704,14 @@ impl TransactionalMemory {
 
     pub(crate) fn get_page_size(&self) -> usize {
         self.page_size.try_into().unwrap()
+    }
+
+    // Pages each thread may cache. Bounded by the configured cache size, so that a small one
+    // shrinks it and set_cache_size(0) disables it, but this is a per thread bound: threads
+    // reading concurrently can hold this many pages each, beyond the configured size
+    fn thread_local_cache_pages(&self) -> usize {
+        let budget = self.storage.max_cache_size() / (self.page_size as usize);
+        budget.min(MAX_THREAD_LOCAL_CACHE_PAGES)
     }
 
     pub(crate) fn close(&self) -> Result {
--- a/src/tree_store/page_store/thread_local_page_cache.rs
+++ b/src/tree_store/page_store/thread_local_page_cache.rs
@@ -1,0 +1,186 @@
+use crate::tree_store::page_store::base::PageNumber;
+use crate::tree_store::page_store::fast_hash::{PageNumberHashMap, Shrink};
+use alloc::rc::Rc;
+#[cfg(not(redb_no_std))]
+use core::cell::RefCell;
+
+/// The snapshot a cached page belongs to.
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub(crate) struct CacheTag {
+    /// Assigned by `TransactionalMemory`, to distinguish databases and reloads of one
+    pub(crate) instance: u64,
+    /// The read transaction's id
+    pub(crate) generation: u64,
+}
+
+/// A thread's copies of the hottest btree pages, so that lookups in read transactions can skip
+/// the shared page cache, whose lock stripe, LRU flag and `Arc` refcount are all writes to
+/// memory other cores read. Being thread local, it needs no locks or atomics.
+///
+/// Every entry belongs to the snapshot named by `tag`, and reading a different one empties the
+/// cache. That is what keeps entries from going stale: a transaction is given generation `G`
+/// only while `G` is the last committed transaction, and pins it, and releasing a page for
+/// reuse takes commits past `G` plus a `process_freed_pages()` that stops at the oldest live
+/// read.
+///
+/// Only the level below the root is cached, which is small enough that a snapshot's pages
+/// normally all fit. Once `max_pages` are held the cache stops admitting, rather than replacing
+/// entries: admitting copies a page, so a level that does not fit would otherwise pay that copy
+/// on nearly every lookup, which costs far more than the shared cache lookup it saves.
+#[derive(Default)]
+pub(crate) struct ThreadLocalPageCache {
+    tag: Option<CacheTag>,
+    max_pages: usize,
+    pages: PageNumberHashMap<Rc<[u8]>>,
+}
+
+#[cfg(not(redb_no_std))]
+thread_local! {
+    static PAGE_CACHE: RefCell<ThreadLocalPageCache> = RefCell::default();
+}
+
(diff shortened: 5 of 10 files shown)


src/tree_store/page_store/transaction_page_cache.rs
```diff
@@ -0,0 +1,199 @@
...
+/// claim slots first. A page whose slots are already taken is simply served from the
+/// shared page cache as before. Since the cache holds pages for the lifetime of the
+/// transaction, it is restricted to `SLOTS` single (order-zero) pages to bound the
+/// pinned memory.
+///
+/// Pages pinned here live outside the shared page cache's accounting: they are not
+/// counted by `CacheStats::used_bytes`, cannot be reclaimed by shared-cache eviction
+/// until the transaction ends, and hits are not counted as shared-cache reads by the
+/// `cache_metrics` feature.
+pub(crate) struct ReadTransactionPageCache {
```
cberner: This is pretty complicated. Did you benchmark a plain bounded RwLock<HashMap> type approach? How much faster is this?

cberner: I hadn't benchmarked that variant before this comment — done now. I implemented a bounded `RwLock<HashMap<PageNumber, PageImpl>>` (128 entries, hits take the read lock and clone the page, inserts take the write lock) with identical gating, and ran it against this design and against no cache on the PR-description workload. Medians of 6 interleaved runs; note this VM is noisy (±20-30% run to run) and absolute numbers are lower than in the PR description, but the runs were interleaved so the relative deltas are meaningful:

| 4 threads | no cache | RwLock\<HashMap\> | OnceLock (this PR) |
|---|---|---|---|
| one read txn per thread | 1.16 Mops/s | 1.44 | 1.48 |
| one shared txn+table | 1.39 | **1.21** | 1.47 |

Two regimes:

- **One txn per thread** (the lmdb_benchmark shape): the two designs are equivalent within noise. The `RwLock` and the cached pages' refcounts are thread-private there, so a hit costs ~4 uncontended atomics — real but too small to see under this noise.
- **One table shared across threads** (`ReadOnlyTable` is `Sync`, so the API invites this): the `RwLock` variant regresses *below no-cache*. Every hit does two RMWs on the single lock word plus an `Arc` inc+dec on the same few hot pages, which concentrates the contention onto fewer cache lines than the 131-stripe shared cache it bypasses. Hits in the `OnceLock` design perform zero writes to shared memory, which is the property the whole scaling argument rests on; the separation should widen with core count.

So the extra complexity buys the shared-table case and the zero-shared-writes guarantee. That said, the subtle part is confined to `get`/`insert` in this file (~50 lines, with unit tests covering the racing and collision paths). If you'd rather take the simpler `RwLock` version and accept the shared-table regression, I'm happy to swap it in — your call.

---

cberner: Fixed in `1d73ddb` — thanks, this was a real hole in the invariant, and the mechanism is worse than the comment describes.

It isn't limited to external replacement preserving the header id. After `clear_cache_and_reload()`, `check_integrity_inner()` re-derives the next transaction id from the *reloaded* header (db.rs, `get_last_committed_transaction_id()?.next()`). So rolling a non-durable commit back rewinds the last committed id, and subsequent commits hand the same ids to different data — an ABA on the generation tag, reachable without any external tampering. The doc comment claimed "bytes of a page number cannot change within a generation", which that makes false.

`clear_cache_and_reload()` now assigns a fresh `cache_instance_id` from the global counter, so every thread's entries for the database stop matching at once. It costs nothing on the read path (the tag is compared either way) and restores the invariant unconditionally, covering the external-replacement case in the report as well.

One caveat on the test: `reads_after_reload_rolls_back_non_durable_commit` exercises the reload-and-rollback path with a populated cache, but it does **not** fail without the fix — I could not line up a page number being reallocated with different contents *and* the id landing back on the same value. So the fix is sound-by-construction rather than backed by a reproducing test, and I'd rather say that plainly than imply otherwise.

---

src/tree_store/btree.rs
```diff
@@ -1014,10 +1046,19 @@ impl<K: Key, V: Value> Btree<K, V> {
         } else {
             None
         };
+        let cache_tag = if guard.is_read() {
+            Some(CacheTag {
+                instance: mem.cache_instance_id(),
+                generation: guard.id().raw_id(),
```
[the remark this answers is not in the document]
cberner: Good catch — fixed in `1d73ddb`. The hole is real, and the mechanism is broader than described here.

It isn't limited to an external replacement preserving the header id. After `clear_cache_and_reload()`, `check_integrity_inner()` re-derives the next transaction id from the *reloaded* header (`get_last_committed_transaction_id()?.next()` in db.rs). So rolling a non-durable commit back rewinds the last committed id, and subsequent commits hand those same ids to different data — an ABA on the generation tag, reachable with no external tampering at all. That makes the module's claim that "bytes of a page number cannot change within a generation" false as written.

`clear_cache_and_reload()` now assigns a fresh `cache_instance_id` from the global counter, so every thread's entries for that database stop matching at once. It costs nothing on the read path (the tag is compared either way) and restores the invariant unconditionally, which also covers the external-replacement case in the report.

One caveat, stated plainly: the new test `reads_after_reload_rolls_back_non_durable_commit` exercises the reload-and-rollback path with a populated cache, but it does **not** fail without the fix — I could not line up a page number being reallocated with different contents *and* the transaction id landing back on the same value. So this is sound-by-construction rather than backed by a reproducing test.

---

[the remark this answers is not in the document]
cberner: The premise is right and the conclusion doesn't follow — I don't think this one is reachable. Documented the reasoning in `68032c9` rather than changing behavior.

You're correct that the id lags the roots: `TransactionGuard::allocate_read()` registers the pin, and `ReadTransaction::new()` calls `get_data_root()` afterwards, so a writer committing in between leaves a transaction pinned at `G` reading tree `G+k`.

What closes the hole is that the pin is taken *at* `G`, which is the conservative direction:

1. A transaction is given generation `G` only while `G` is the last committed transaction — `register_read_transaction()` reads that id and inserts the pin under one `state` lock acquisition. Ids only increase (the one exception, in-place reload, now bumps the instance id), so after any commit lands, no later transaction is ever handed `G` again.
2. Reusing a page number for different bytes takes more than one commit past `G`: the page must be freed by a commit and then released by `process_freed_pages()`, whose horizon stops at `oldest_live_read_transaction()`. Every transaction holding `G` pins `G`, so nothing reachable from any tree they can read is released while one of them exists.

So an entry tagged `G` is only ever read back by transactions pinning `G`, and its page cannot have been reused by then. The scenario needs a transaction to obtain `G` *after* the reuse, which (1) forbids: the commits required for reuse have already made `G` unobtainable. A transaction pinned at `G` reading tree `G+k` is fine for the same reason it is fine today — the older pin protects strictly more pages than a pin at `G+k` would.

On the cross-thread drop in your scenario: the flush in `TransactionGuard::drop` is a memory optimization, not the correctness mechanism, and it does only clear the dropping thread's cache. A transaction created on one thread and dropped on another leaves the first thread's entries in place until they are overwritten. That is a real (minor) limitation of the flush worth knowing about, but tag validation is what makes those entries safe, and it does not depend on the flush running.

The module comment did overstate this — it claimed bytes cannot change "within a generation" without saying why a lagging id is still sound. `68032c9` states both properties above explicitly, since this is exactly the kind of reasoning a future change could break silently.

---

CHANGELOG.md
```diff
@@ -101,6 +101,9 @@
 * Add `ExtractIf::close()` to explicitly finalize an extract iterator without removing unread
   entries.
 * Optimize `Table::pop_first()` and `Table::pop_last()` to be about 2x faster.
+* Improve multithreaded read scaling. Each thread now caches the top of the btrees it reads,
```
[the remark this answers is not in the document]
cberner: This one is incorrect — the commit metadata is not what's described here.

`git log -1 --format='%an <%ae> / %cn <%ce>'` on the reviewed commit `1d73ddb`:

```
Christopher Berner <me@cberner.com> / Christopher Berner <me@cberner.com>
```

Both author and committer are the human, and the message carries the `Assisted-by: Claude Code` trailer that AGENTS.md asks for. The repository's own `commit-authors` check — the one this comment says rejects it — passed on that exact commit ([job](https://github.com/cberner/redb/actions/runs/31516452126/job/93862680984)), as did all 8 checks.

For what it's worth, an earlier revision of this branch *did* trip that check: `git rebase --continue` stamped the committer as `Claude <noreply@anthropic.com>` while leaving the author intact. That was fixed before `1d73ddb`, and the repo-local `user.name`/`user.email` are now set so it cannot recur.

---

src/tree_store/page_store/thread_local_page_cache.rs
```diff
@@ -0,0 +1,282 @@
...
+/// transaction, e.g. 512KiB per thread with 4KiB pages, plus one page-copy memcpy per
+/// fill. Only pages actually read are cached, and they are released when the thread's
+/// read transaction is dropped, so idle threads retain only the empty slot array. Write
+/// transactions must not use the cache (they can free and reallocate pages
+/// mid-transaction, within one generation), and only branch pages are cached: leaf pages
+/// must be returned as `PageImpl`s from the shared cache anyway, because `AccessGuard`s
+/// hold them.
+pub(crate) struct ThreadLocalPageCache {
+    slots: Box<[Option<CacheEntry>]>,
+}
```
cberner: Do we really need this custom cache? It's thread local. Can't we just use a HashMap or something?

cberner (commented):
Polish the code structure. It feels a bit spread out all over the place, and with overly verbose comments too.

src/tree_store/btree.rs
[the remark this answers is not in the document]
cberner: You're right, and it turned out to be both simpler *and* faster. Replaced in `154a00c`.

The custom slot array was left over from when entries had to survive across transactions and be validated individually. Once the cache is emptied when a read transaction ends, every live entry belongs to the same snapshot — so one tag for the whole cache replaces per-entry tags, and reading under a different snapshot just empties it. That deletes the slot array, the probe sequence, the collision handling and the eviction policy:

```rust
pub(crate) struct ThreadLocalPageCache {
    tag: Option<CacheTag>,
    pages: PageNumberHashMap<Rc<[u8]>>,
}
```

`get` is a tag compare and a map lookup; `insert` clears first if the tag differs, then inserts while under the size cap. The module went from ~280 lines to ~150, most of that now tests.

It also measures better, because the probe sequence was silently dropping hot pages on collisions. Shared-cache reads per lookup on the 5M-row height-5 tree:

| | reads per get |
|---|---|
| no cache | 4.00 |
| slot array | 2.21 |
| HashMap | **2.00** |

2.00 is the floor for caching two levels below the root, so the collision waste is gone entirely. Throughput at 4 threads on my (noisy) VM went 1.60 → 2.03-2.14 against master, i.e. no worse than before and probably a little better.

One behavior note: two read transactions interleaved on one thread now thrash, since each insert under the other's tag empties the cache. Correct, just not cached — the per-entry tags used to let both live. It seemed the right trade for this much less machinery, but say the word if you'd rather keep both snapshots.

Verified with the full suite, clippy `-D warnings`, and 100k fuzz runs on this version.

---

src/tree_store/page_store/thread_local_page_cache.rs
cberner: You're right — done in `154a00c`, and it turned out both simpler and faster. (I posted this on the wrong thread first, sorry; it's the one above on btree.rs.)

The custom cache was left over from when entries survived across transactions and had to be validated individually. Once the cache is emptied at the end of a read transaction, every live entry belongs to the same snapshot — so one tag for the whole cache replaces per-entry tags, and reading under a different snapshot just empties it. That deletes the slot array, the probe sequence, the collision handling and the eviction policy:

```rust
pub(crate) struct ThreadLocalPageCache {
    tag: Option<CacheTag>,
    pages: PageNumberHashMap<Rc<[u8]>>,
}
```

`get` is a tag compare plus a map lookup; `insert` empties first if the tag differs, then inserts while under the size cap. The module is ~150 lines now, over half of it tests.

It also measures better, because the probe sequence was silently dropping hot pages on collisions — shared-cache reads per lookup, 5M-row height-5 tree:

| | reads per get |
|---|---|
| no cache | 4.00 |
| slot array | 2.21 |
| HashMap | **2.00** |

2.00 is the floor for caching two levels below the root, so the collision waste is gone. Throughput at 4 threads on my (noisy) VM: 1.60 → 2.03-2.14 vs master, so no worse than the slot array and probably slightly better.

One behavior change worth knowing: two read transactions interleaved on one thread now thrash, since each insert under the other's tag empties the cache. Still correct, just uncached — per-entry tags used to let both live. Seemed the right trade for this much less machinery, but happy to revisit if you think that pattern matters.

On the structure/verbosity feedback in your other comment: I cut the module's doc block down to the invariant argument and trimmed the commentary in `btree.rs`. The change now touches `btree.rs` (traversal + tag), the cache module, and small hooks in `db.rs` (release on transaction end) and `page_manager.rs` (instance id). Happy to pull more of it together if it still reads as scattered.

---

cberner:
Audited the tests in `23cdd98`, and the audit turned up something worse than redundancy: **four of the six integration tests were not exercising the cache at all.**

Only branch pages below the root are cached, so a tree of height <= 2 (root plus leaves) leaves the cache empty. Measuring what the tests actually built:

| test | value | tree height | exercised the cache? |
|---|---|---|---|
| `read_snapshot_stable_while_writes_proceed` | 50B | 3 | yes |
| `concurrent_reads_and_writes` | 50B | 3 | yes |
| `alternating_reads_and_writes_invalidate_thread_cache` | u64 | 2 | **no** |
| `sequential_multi_table_reads` | u64 | 2 | **no** |
| `reads_after_reload_rolls_back_non_durable_commit` | u64 | 2 | **no** |
| `multithreaded_reads` | u64 | 2 | **no** |

`u32 -> u64` tables stay at height 2 even at 20k entries, so those four were passing as generic correctness tests while testing nothing about this change.

What I did:

- **Removed** `sequential_multi_table_reads` (the per-table cliff it was written for no longer exists — one cache is shared now), `multithreaded_reads` (`concurrent_reads_and_writes` covers concurrent readers and adds a writer), and `reads_after_reload_rolls_back_non_durable_commit` (vacuous, and by my own admission above it did not fail without its fix; Codex separately flagged it for opening a second writable handle to the locked file, which is a fair portability complaint).
- **Fixed** `alternating_reads_and_writes_invalidate_thread_cache` to build a height-3 tree, since generation invalidation is the property most worth an integration test.
- **Guarded** each remaining cache test with `assert!(tree_height() > 2)`, so shrinking the data can never silently make them vacuous again.
- **Trimmed** `read_snapshot_stable_while_writes_proceed` from three rewrite rounds to two: 9.9s -> 1.3s, and it was by far the most expensive test in the change.

Net: 6 integration tests -> 3, each verified to exercise the cache, and the added suite time drops from ~16s to ~7s. The four unit tests stay — they cover tag validation, the empty-on-new-snapshot path, the size bound and `clear()`, each distinct and effectively free.

Also in this push, from the Codex P2 that was worth acting on: `set_cache_size()` now bounds the thread-local cache too. It previously sat entirely outside that budget, so a database configured with a small cache — or `set_cache_size(0)` — still got 128 pages per reading thread. The per-thread cap is now `min(128, configured_bytes / page_size)`, so a zero budget disables it, with a test covering that case. Hit rate is unchanged at 2.00 shared-cache reads per lookup.

The other Codex P1 ("Rewrite the AI-authored commit") is wrong again: it cites `4e18858faaa880806da5cd597d6bb740c3923c32`, which does not exist in this repository (`git cat-file -t` -> `no such commit`). `commit-authors` has passed on every pushed revision.

---

src/tree_store/page_store/thread_local_page_cache.rs
```diff
@@ -0,0 +1,185 @@
...
+use std::cell::RefCell;
+use std::rc::Rc;
+
+/// The snapshot a cached page belongs to: the database instance it was read from, and the
+/// read transaction's id.
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub(crate) struct CacheTag {
+    pub(crate) instance: u64,
+    pub(crate) generation: u64,
+}
```
cberner, line 14: Hmm, is this sufficient? What if multiple Databases are open in a single process. It seems like this could cause correctness issues by returning cached data for a different Database

src/tree_store/page_store/thread_local_page_cache.rs
```diff
@@ -0,0 +1,185 @@
...
+///   `register_read_transaction()` reads that id and pins it under one lock acquisition, and
+///   ids only increase. Note that the pinned id can lag the roots the transaction goes on to
+///   read, which is safe because the older pin is the conservative one.
+/// * Reusing a page number for different bytes takes more than one commit past `G`, because
+///   the page must be freed and then released by `process_freed_pages()`, which stops at the
+///   oldest live read. Every transaction holding `G` pins `G`.
+///
+/// So a page cached under `G` cannot have been reused while any transaction can still read it
+/// back. In-place reload is the exception, since rolling a non-durable commit back rewinds the
+/// id; `clear_cache_and_reload()` bumps the instance id to cover it.
```
cberner: This is too verbose

src/tree_store/page_store/thread_local_page_cache.rs
```diff
@@ -0,0 +1,185 @@
...
+#[derive(Default)]
+pub(crate) struct ThreadLocalPageCache {
+    tag: Option<CacheTag>,
+    // Page budget of the database `tag` names, so that a database configured with a small
+    // cache does not gain memory back through its readers' threads
```
cberner: This is pretty obvious from the field name. Do a pass to clean up unnecessary comments

src/tree_store/page_store/thread_local_page_cache.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /// Caches a copy of `bytes`, holding at most `max_pages` of them, and first empties the
+    /// cache if it holds another snapshot's pages. Interleaving reads of two snapshots on one
+    /// thread therefore keeps refilling it, which costs a page copy per lookup but stays correct.
+    pub(crate) fn insert(
+        &mut self,
+        tag: CacheTag,
+        max_pages: usize,
+        page_number: PageNumber,
```
cberner, line 93: I think we should have an assertion that only order-0 pages get cached. We don't want it getting bloated in edge cases

src/tree_store/btree.rs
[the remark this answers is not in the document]
cberner: All four addressed in `90982b8`.

**Multiple databases in one process** — this is what `instance` is for, and it is sufficient. Every `TransactionalMemory` takes a fresh id from a process-wide `AtomicU64` at construction, so two `Database`s never share a tag even though their page numbers overlap and they share the thread's cache. A lookup against database B finds either nothing or, if B was read most recently, only B's pages.

It was only covered by a unit test, so `90982b8` adds `two_databases_do_not_share_cached_pages`: two databases holding different values at the same keys, read alternately on one thread, asserting each returns its own. It also happens to be the worst case for the single-tag design — every lookup swaps the snapshot and refills — and it passes.

**Only order-0 pages cached** — added `assert_eq!(page_number.page_order, 0)` in `insert()`. The caller already gates on it, so this pins the invariant where the memory bound is actually promised. `insert()` runs only on a miss, so it is not on the hot path.

**Comments** — did the pass. The module doc went from ~28 lines to ~10, keeping only the two facts that make the tag sufficient (a generation is only issued while it is the last committed one and is pinned; releasing a page for reuse needs commits past it plus a `process_freed_pages()` bounded by the oldest live read) and dropping the prose around them. Removed the `max_pages` field comment and the others that restated their identifier, and tightened what remains in `btree.rs` and `db.rs`. The module is now 173 lines including tests, down from 282.

Verified: full suite, clippy with `-D warnings`, fmt, and 83k fuzz runs — which also covers the `set_cache_size` bounding from the previous push, since that had not been fuzzed yet.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -26,6 +26,15 @@ use std::marker::PhantomData;
 use std::mem;
 use std::sync::Arc;
 use std::sync::Mutex;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+// Source of `TransactionalMemory::cache_instance_id`. Starts at 1 so that 0 can mean
+// "no database" in thread-local page cache entries.
```
cberner: It seems errorprone that we rely on this being process wide. I think we should move it into the TransactionMemory or something, and then use the same address check that savepoints use to avoid restoring across databases

src/tree_store/page_store/page_manager.rs
```diff
@@ -1557,6 +1585,15 @@ impl TransactionalMemory {
         self.page_size.try_into().unwrap()
     }
 
+    // Pages each thread may copy into its page cache. Bounded by the cache size the caller
+    // configured, so that `set_cache_size()` still caps memory: a small budget shrinks it and
+    // a zero one disables the cache entirely, rather than every reading thread adding to the
+    // configured total.
```
cberner: Too verbose

src/db.rs
```diff
@@ -330,6 +331,12 @@ impl TransactionGuard {
             }
         }
     }
+
+    // Write transactions may free and reallocate pages within the transaction, so readers
+    // they create must not cache pages
```
cberner: Comments should document the function not how its callers work. Either move this comment to the caller or just remove it.

src/db.rs
```diff
@@ -338,7 +345,13 @@ impl Drop for TransactionGuard {
             Self::Read {
                 tracker,
                 transaction_id,
-            } => tracker.deallocate_read_transaction(*transaction_id),
+            } => {
+                // So that a thread which reads once does not hold pages for its lifetime.
+                // Dropping on another thread, or while this thread has a second read
+                // transaction open, only costs the optimization.
```
cberner: Not a good comment. Say it clears the thread local cache so that it doesn't "leak" cache space if the thread doesn't do another tranasction

cberner (commented):
Do a careful pass over all the comments you wrote in the code. I had a fair bit of feedback on them. Match my style in the rest of the code base

cberner:
Comment pass done in `b01c6c9`, and the counter stays with real docs on it.

**On the instance id** — I looked at what savepoints do (`std::ptr::from_ref(self.transaction_tracker.as_ref()) != savepoint.db_address()`) and I don't think the bare address works here, because savepoints get their safety from something this cache deliberately lacks: a `Savepoint` holds the `Arc<TransactionTracker>`, so the address it compares can't be reused while the savepoint exists. The cache holds no reference to the database, by design — it must not keep one alive.

Without that, the address can be recycled:

- `ReadTransaction` is `Send`, so a transaction read on thread A can be dropped on thread B. The release in `TransactionGuard::drop` only clears the dropping thread's cache, so A keeps its entries.
- The database is then dropped, and a new one allocates its tracker at the same address.
- New databases start at `TransactionId::new(0)` (page_manager.rs), so the new database's generations walk straight through the ones A cached under.

Tag matches, and A gets the old database's bytes for the new database's page numbers. Narrow, but silent corruption, so it didn't seem like the right thing to ship on a maintainer's suggestion without flagging it. Pinning the address with a `Weak` would close it, at the cost of a `Weak` field plus a separate reload epoch — more machinery than the counter, which is why I asked rather than picking.

The static now documents what it has to guarantee and how, per your comment:

```rust
// Assigns every TransactionalMemory, and every in-place reload of one, an id that no other
// instance in this process has held, which is what lets a thread local page cache tell whose
// pages it is holding. The id must be unique process wide, not just within one database:
// thread local caches outlive the databases that filled them, and every database a thread
// reads shares one cache, so an id unique only within a database would let one database's
// pages match another's tag. Ids are never reused -- incrementing a u64 once per instance and
// per reload cannot wrap in practice -- so a tag from a dropped database matches nothing.
```

**Comments** — went through every one I added:

- `TransactionGuard::is_read()`: removed, it described the callers.
- The release call site now says it clears the thread local cache so it doesn't leak cache space if the thread runs no further transaction, as you wrote it.
- `thread_local_cache_pages()`: cut to two lines.
- Trimmed the rest to the same shape — `cache_instance_id`, the reload invalidation, `MAX_CACHED_DEPTH`, `get_page_for_read`, `NodePage`, `cache_tag`, and the cache module's own docs. The module doc keeps only the two facts that make the tag sufficient, since that is the part a future change could silently break.

Verified: full suite, clippy with `-D warnings`, fmt.

---

cberner:
One correction to my own comment, in `c457f5f`, plus an open question for you.

Codex is right that the `set_cache_size()` bound I added is **per thread**, not aggregate: each reading thread may hold up to `min(128, configured / page_size)` pages, so N concurrent readers can hold N times that beyond the configured size. My comment claimed it meant "`set_cache_size()` still caps memory", which overstated it. Now:

```rust
// Pages each thread may cache. Bounded by the configured cache size, so that a small one
// shrinks it and set_cache_size(0) disables it, but this is a per thread bound: threads
// reading concurrently can hold this many pages each, beyond the configured size
```

What the bound does give: `set_cache_size(0)` disables the cache outright, and a small budget shrinks the per-thread cap proportionally. What it doesn't: a total ceiling. At the default 1GiB the per-thread cap is the full 128 pages (512KiB/thread), which is the design as it stands; at, say, 1MiB with 16 readers it is ~8MiB against a 1MiB setting.

Whether that needs fixing is your call, and the options each cost something:

- **Leave it**, documented as above. Per-thread caches outside the global budget are a normal pattern, and the absolute numbers are small.
- **Divide the budget by an assumed thread count** — arbitrary, and wrong in both directions depending on the workload.
- **Track the aggregate in an atomic on `TransactionalMemory`** and stop caching past a share of the budget. Exact, but it puts an atomic back on the fill path (misses only, not hits) and re-adds machinery you asked me to strip out.

I didn't want to pick a policy for you here, so I only corrected the comment. Say which you'd like and I'll do it.

Unrelated: the `commit-authors` P1 on this thread has now cited four SHAs that do not exist in this repository (`4e18858`, `d082f32`, `a16c4f6`, and one earlier). The branch has one commit, authored and committed by you, and that job has passed on every push. I've stopped replying to it individually.

---

src/tree_store/page_store/thread_local_page_cache.rs
```diff
@@ -0,0 +1,181 @@
...
+        // Caching a multi-page value would break this cache's memory bound
+        assert_eq!(page_number.page_order, 0);
+        if self.tag != Some(tag) {
+            self.clear();
+            self.tag = Some(tag);
+            self.max_pages = max_pages;
+        }
+        if self.pages.len() < self.max_pages {
+            self.pages.insert(page_number, Rc::from(bytes));
+        }
```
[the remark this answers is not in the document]
cberner: Confirmed and fixed in `0df6351`. This was real, and worse than "suboptimal" — I measured it and the second table got *zero* benefit.

The reproduction: read table A until the cache fills, then measure only lookups in table B within the same read transaction. Shared-page-cache reads per lookup (`cache_metrics`), 5M-row height-4 table, against master `5e6ad02`:

| | one table | A then B |
|---|---|---|
| master (no cache) | 3.00 | 3.00 |
| stop inserting when full | 2.07 | **3.00** |
| clear when full | 2.28 | 2.28 |
| two halves (now) | **2.12** | **2.12** |

"A then B" at 3.00 is exactly the uncached baseline, so the cache was doing nothing for B.

I did try the literal suggestion — evict one existing entry per admission — using `pages.keys().next()` as the victim. It measured *worse* on both shapes (2.70 / 3.00): with no use ordering the sweep evicts the few hot top-level pages as readily as cold ones, and the newly inserted entry frequently becomes the next victim.

What shipped instead: entries live in two halves. Insertions fill `pages`; when it reaches half the bound the two halves swap and the older one is cleared. A page read again before its half is discarded is promoted back into `pages`, so what stays cached is what is still being read, with no per-entry bookkeeping and no use-order tracking. `|pages| + |evicting| <= max_pages` is asserted on every insertion in the `bounded` unit test, and `a_page_read_again_outlives_the_rotation` covers promotion.

This keeps essentially all of the single-table win (2.07 → 2.12) while making the multi-table case work at all.

---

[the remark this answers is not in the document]
cberner: Correct, and I had the invariant wrong. Fixed in `ed3271c`.

Promotion moved an entry from `evicting` to `pages`, which keeps the *total* constant, and I stopped reasoning there. What I missed is that it can push `|pages|` past `half` — and then the rotation swaps that oversized map into `evicting` and keeps it, so the total ratchets up by one per round. Exactly as described.

I wrote your scenario as a test before fixing anything: insert a page, read back everything still held, repeat. With `MAX_PAGES = 128` it peaked at **384** held pages, and it fails on the parent commit:

```
PEAK held = 384, bound = 128
```

The fix routes both callers through one admission point, so the rotation check cannot be bypassed:

```rust
/// Adds an entry to the live half, rotating the halves first when it is full. Both callers
/// go through here, so that promoting cannot grow the cache past its bound either
fn admit(&mut self, page_number: PageNumber, bytes: Rc<[u8]>) {
    let half = self.max_pages / 2;
    if half == 0 {
        return;
    }
    if self.pages.len() >= half {
        swap(&mut self.pages, &mut self.evicting);
        self.pages.clear();
    }
    self.pages.insert(page_number, bytes);
}
```

`|pages| <= half` now holds after every operation, and `evicting` is only ever assigned from `pages`, so `|pages| + |evicting| <= max_pages` follows. `promoting_respects_the_bound` asserts it after each round of that access pattern and is the regression test for this.

No performance cost: still 2.12 reads/lookup on both the single-table and A-then-B workloads.

---

PR #1348, opened 2026-08-09
Stop holding the shared freed-pages lock across whole btree descents

cberner:
First of a stack of three PRs fixing the multithreaded insert benchmark's TODO (multi-threaded inserts slower than single-threaded). Stack order: this PR, then the write-buffer striping PR, then the `MutateHelper` borrow PR.

Tables of a single `WriteTransaction` share one freed-pages list, and `BtreeMut::insert()`, `remove()`, `pop_first()`, `pop_last()`, and `insert_reserve()` locked its mutex before descending the tree and held it until the operation finished. Tables may be written from different threads, and this serialized every mutation in the transaction — the largest of the three contention points behind the benchmark regressing from 1.26s at 1 thread to 2.06s at 4 threads (4-core machine, 1M u128 pairs split across N tables, one thread per table).

Each operation now accumulates the pages it frees in a scratch list held on `BtreeMut`, and merges that into the shared list under a short lock once the operation is done — the discipline the cursor paths (`CursorTree::drain_freed`) and `retain_in` already follow. Keeping the list on `BtreeMut` rather than building one per operation reuses its capacity: `Vec::append` leaves it empty but allocated for the next operation, so an operation that frees a committed page no longer allocates. Merging per operation, rather than when the tree is dropped, keeps the shared list complete whenever control returns to the caller and keeps the shared mutex out of drop paths. The merge runs on the failure path too, since pages queued before an error still need to be recorded for commit.

`retain_in_bounds` keeps its own list, because `retain_in_helper` takes `&mut self` and so cannot also borrow the field; there one allocation is amortized over the whole scan.

Also adds a `multithreaded_insert_and_remove` test that mutates four tables of one transaction concurrently, inserting and removing committed entries, to exercise merging freed pages from multiple threads under the debug-build invariant assertions. The existing `multithreaded_insert` test joins its spawned thread before the main thread writes, so it never actually overlaps two writers.

With this change alone, 2 threads go from 1.3x slower than 1 thread to ~1.2x faster (1.59s → ~1.23s); 4+ threads still don't scale because the page cache's write-buffer mutex becomes the next serialization point, addressed in the next PR of the stack.

Verified with the full test suite (all features, debug + release), clippy with denied warnings, and a 4-minute fuzz run of `fuzz_redb` on the combined stack.

### DIFF
--- a/src/tree_store/btree.rs
+++ b/src/tree_store/btree.rs
@@ -334,11 +334,26 @@ impl UntypedBtreeMut {
     }
 }
 
+// Merges pages freed by a mutation into the transaction's shared freed-pages list, leaving the
+// source empty with its capacity intact. The shared mutex is held only for the merge, never across
+// a whole tree descent: tables of one WriteTransaction share it and may be written from different
+// threads. Callers must merge before propagating errors, since pages queued before a failure still
+// need to be recorded for commit.
+fn merge_freed_pages(freed_pages: &Mutex<Vec<PageNumber>>, freed: &mut Vec<PageNumber>) {
+    if !freed.is_empty() {
+        freed_pages.lock().unwrap().append(freed);
+    }
+}
+
 pub(crate) struct BtreeMut<K: Key + 'static, V: Value + 'static> {
     page_allocator: PageAllocator,
     transaction_guard: Arc<TransactionGuard>,
     root: Option<BtreeHeader>,
     freed_pages: Arc<Mutex<Vec<PageNumber>>>,
+    // Scratch list for the pages freed by a single operation, merged into `freed_pages` before
+    // that operation returns. Kept here, instead of allocated per operation, so that its capacity
+    // is reused across operations that free committed pages.
+    local_freed: Vec<PageNumber>,
     allocated_pages: Arc<Mutex<PageTrackerPolicy>>,
     _key_type: PhantomData<K>,
     _value_type: PhantomData<V>,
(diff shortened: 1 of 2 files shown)


cberner:
Done — pushed the scratch field (`BtreeMut::local_freed`), drained into the shared list at the end of each operation rather than on `Drop`.

I went with per-operation drain over `Drop` for three reasons: it keeps the shared list complete whenever control returns to the caller (other paths touch it mid-transaction, e.g. `delete_table` on another table from another thread), it keeps the shared mutex out of drop/unwind paths (a poisoned lock during unwinding would abort), and it matches `CursorTree`'s existing `freed` + `drain_freed()` shape. `Vec::append` leaves the field empty with its capacity intact, so the reuse you were after comes for free. `retain_in_bounds` still uses its own list, since `retain_in_helper` takes `&mut self` and can't also borrow the field.

One correction to the premise, since it changes how much this buys: the version you reviewed didn't allocate per operation either. `vec[]` doesn't allocate until first push, and `conditional_free` only pushes when a *committed* page is freed — so operations whose path is already dirty never allocated at all.

I measured it with a counting global allocator, overwriting 200k committed entries in one transaction:

| | allocations |
|---|---|
| local `vec[]` | 201,634 |
| `local_freed` field | 200,850 |

So it removes ~780 allocations, not ~200k: the first write to a leaf CoWs it and queues the freed page, and every later write to that now-dirty leaf takes the in-place fast path and frees nothing. The saving is one allocation per CoW'd page (~780 leaves here), not one per operation. Wall-clock on the multithreaded benchmark is unchanged within noise, as expected.

Still worth having in my view — it costs nothing, makes the no-allocation property structural rather than incidental, and converges on the cursor code's pattern — but I didn't want to leave a "removes an allocation per insert" impression that the numbers don't support. Happy to drop it back to the local `vec[]` if you'd rather keep the diff minimal.

---

cberner (commented):
lgtm

PR #1350, opened 2026-08-09
Borrow the page allocator and allocation tracker in MutateHelper

cberner:
Last of the stack. #1348 and #1349 have merged, so this is now based directly on master and contains only its own commit.

`MutateHelper` cloned the `PageAllocator` (two inner `Arc`s) and the allocation tracker's `Arc` on every mutation. Each clone and drop is a pair of atomic refcount updates on cache lines shared by every writer thread, so with tables of one `WriteTransaction` written from different threads they showed up as measurable cross-core traffic in profiles once the coarser locks from the first two PRs stopped serializing the writers — and they're wasted work single-threaded too.

`CursorMut` already borrows both; this aligns `MutateHelper` with it (`page_allocator: &'b PageAllocator`, `allocated: &'b Arc<Mutex<PageTrackerPolicy>>`), which also removes the clones from the cursor's `mutate_helper()` bridge and simplifies every construction site.

On the multithreaded insert benchmark (1M u128 pairs split across N tables, one thread per table, 4-core machine), 4 threads improve from ~0.65s to ~0.52s on top of the first two PRs, and single-threaded throughput improves a few percent since the clones were unconditional. Full series vs. the pre-stack baseline: 1 thread 1.26s → ~1.25s, 2 threads 1.59s → ~0.75s, 4 threads 2.06s → ~0.51s, 8 threads 2.23s → ~0.56s.

Verified with the full test suite (all features, debug + release), clippy with denied warnings, and a 4-minute fuzz run of `fuzz_redb`.

### DIFF
--- a/src/tree_store/btree.rs
+++ b/src/tree_store/btree.rs
@@ -451,9 +451,9 @@ impl<K: Key + 'static, V: Value + 'static> BtreeMut<K, V> {
         );
         let mut operation: MutateHelper<'_, '_, K, V> = MutateHelper::new(
             &mut self.root,
-            self.page_allocator.clone(),
+            &self.page_allocator,
             &mut self.local_freed,
-            self.allocated_pages.clone(),
+            &self.allocated_pages,
         );
         let result = operation.insert(key, value);
         merge_freed_pages(&self.freed_pages, &mut self.local_freed);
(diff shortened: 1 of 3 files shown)


cberner:
lgtm

PR #1353, opened 2026-08-10
Skip the allocation tracker's lock when no savepoint needs it

cberner:
#1352 has merged, so this is now based directly on master and contains only its own commit.

Sharding the uncommitted-pages set left the other transaction-wide allocation lock as the next one to be contended: every table of a `WriteTransaction` shares an allocation tracker, and every allocation and free locks it — builders held the guard across the allocation *and* the whole page construction (`build_split` held it across two allocations plus all the key/value copying).

**On its own this change measures as nothing** — I tried it standalone against master first and it was indistinguishable, because the uncommitted-pages set was the bottleneck. It only pays now that #1352 is in.

That set exists solely so that restoring a savepoint can free the pages a transaction allocated, so a transaction with no live savepoint tracks nothing and already sets the policy to `Ignore` — but callers still took the lock to reach it. This wraps the policy in a `PageTracker` holding an atomic flag alongside the mutex, checks the flag first, and touches the mutex only while tracking is on, which is the uncommon case.

Two things make that safe:

- **The flag can be relaxed.** Nothing is ordered against it: a writer that misses a concurrent `disable()` records a page into a set that is about to be discarded, and one that sees it early skips a page nothing will read.
- **Tracking never turns back on within a transaction.** It is disabled only while no savepoint exists; `reset()` leaves `Ignore` alone and `close()` only moves to the poisoned `Closed` state. So a page can never be recorded while disabled and then removed while enabled, which would trip the tracker's assertion that a removed page was present.

Passing the tracker rather than a guard also shortens the critical section when tracking *is* on: builders now take the lock per allocation instead of holding it across page construction. `PageTrackerPolicy` becomes an implementation detail of `PageTracker`, and `remove_if_present`/`contains` fall out as dead code.

Benchmark as in #1352 (`multithreaded_insert_benchmark`, 16-core / 32-thread 9950X3D):

| Threads | before #1352 | #1352 | this PR on top |
|---|---|---|---|
| 8 | ~145ms | ~134ms | ~132ms |
| 16 | ~211ms | ~140ms | ~122ms |
| 32 | ~240ms | ~167ms | ~132ms |

It also removes most of the run-to-run variance at 32 threads: #1352 alone gave 192/155/153ms across three runs, with this it is 131/134/132ms.

Scaling still flattens after eight threads (4.8x at 8, 5.2x at 16). What remains is the global `state` mutex in `allocate_helper`, which every allocation takes; that is a different kind of change and not attempted here.

Verified with the full test suite (all features, debug and release) and clippy with denied warnings, on the rebased commit.

### DIFF
--- a/src/multimap_table.rs
+++ b/src/multimap_table.rs
@@ -13,7 +13,7 @@ use crate::tree_store::encode_bounds;
 use crate::tree_store::{
     AllPageNumbersBtreeIter, BRANCH, Btree, BtreeCursorRange, BtreeHeader, BtreeMut,
     DynamicCollection, DynamicCollectionType, LEAF, LeafAccessor, MAX_PAIR_LENGTH,
-    MAX_VALUE_LENGTH, Page, PageAllocator, PageHint, PageNumber, PageResolver, PageTrackerPolicy,
+    MAX_VALUE_LENGTH, Page, PageAllocator, PageHint, PageNumber, PageResolver, PageTracker,
     RawBtree, RawLeafBuilder, multimap_btree_stats,
 };
 use crate::types::{Key, Value};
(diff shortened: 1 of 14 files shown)


cberner (commented):
Lgtm

PR #1355, opened 2026-08-11
Defer non-durable commits' freed-page records to the next durable commit

cberner:
Follow-up to #1318, which stopped `Durability::None` commits writing dirty pages to the file. This removes the other per-commit cost: the system-tree mutation that records freed pages.

Every commit records the data-tree pages it freed in `DATA_FREED_TABLE`, so they can be released once no reader needs them. For a non-durable commit that means allocating pages, copying nodes and recomputing checksums in the system tree, on a path that otherwise touches only the data tree and (after #1318) does no I/O at all.

Those records are now kept in memory, keyed by transaction, and written to `DATA_FREED_TABLE` at the start of the next durable commit. This mirrors what redb already does on the allocation side -- non-durable commits record allocations in `UnpersistedState::allocations` rather than `DATA_ALLOCATED_TABLE`, for the same reason: a crash that loses the records also rolls back the commits that produced them, so the two stay consistent. The new records join that state, so they are cleared by the same `clear()` (see #1356).

## What has to account for the records not being on disk

- **Allocator rebuild.** `rebuild_allocator_state()` marks the pages named by the freed tables as allocated, because they are still in use until a commit processes them. It now marks the in-memory records the same way, via `pages_pending_free()`. Without this a repair hands those pages out while a record still names them, and `check_integrity()` reports a healthy database as unclean because the rebuilt allocator no longer matches the live one.
- **Savepoint restore.** A restore discards the commits whose records these are, so the restoring transaction drops them. Tracked transaction-locally, so an abort leaves them in place.
- **Range bounds.** The non-durable path reads its records from memory rather than the table. Its two bounds come from different places -- the oldest unprocessed non-durable commit and the oldest live read -- so a live read older than that commit inverts the range. `BTreeMap::range` panics on that, unlike the table range it replaces, so the empty case is handled explicitly.
- **Lock re-entrancy.** Reclaiming a page locks the same state that holds these records, so `process_unpersisted_data_freed()` copies the records out before running its callback and writes back what was not reclaimed. Holding the lock across the callback self-deadlocks every non-durable commit that frees a page.

Note that only the data-tree records are deferred. The system tree is written by every commit regardless, so `process_freed_pages_nondurable_helper()` still reads `SYSTEM_FREED_TABLE`.

## Benchmarks

20k single-key `Durability::None` commits over a 500k row table, alternating runs on one machine (medians):

| | pre-#1318 | master (#1318 merged) | this PR |
|---|---|---|---|
| overwrites | 1116 ms | 817 ms | **467 ms** |
| fresh inserts | 1235 ms | 915 ms | **559 ms** |

So about **1.7x** on top of master, and about **2.2-2.4x** against master before #1318. The work moves to the next durable commit, which pays for it once.

(An earlier revision of this description quoted ~2.4-2.7x for this PR alone. That came from measuring the original implementation against a base that predated the write-buffer striping and other merged work, and does not hold against current master. The numbers above are same-machine, same-session.)

## Validation

- Full test suite (418 tests), clippy and fmt clean.
- `fuzz_redb` over the corpus: 46,595 runs, no artifacts. Both the range-inversion panic and the lock re-entrancy above were caught by testing rather than inspection -- the first by fuzzing, the second by the non-durable tests hanging -- so the fuzz and test passes here are load-bearing, not a formality.

## Attribution

The approach comes from an independent implementation by Codex on `hitch/redb/20260810170452-04f36b3b`, which paired it with a write-buffer change equivalent to #1318. This port keeps its design and adds the range guard; the read-path cache-pressure reclaim from #1318 is retained, which that branch does not have.

### DIFF
--- a/src/transactions.rs
+++ b/src/transactions.rs
@@ -907,6 +907,9 @@ pub struct WriteTransaction {
     shrink_policy: ShrinkPolicy,
     quick_repair: bool,
     post_commit_free: PostCommitFree,
+    // Set by a savepoint restore, to discard the in-memory freed-page records of the non-durable
+    // commits it rolled back. Kept transaction-local so that an abort leaves them in place.
+    restored_transaction: Option<TransactionId>,
     // All transaction-local savepoint lifecycle state. See
     // `SavepointTransactionState` for the commit/abort contract.
     savepoint_state: Mutex<SavepointTransactionState>,
@@ -943,6 +946,7 @@ impl WriteTransaction {
             two_phase_commit: false,
             quick_repair: false,
             post_commit_free: PostCommitFree::Enabled,
+            restored_transaction: None,
             shrink_policy: ShrinkPolicy::Default,
             savepoint_state: Mutex::new(SavepointTransactionState::default()),
         })
@@ -1405,6 +1409,8 @@ impl WriteTransaction {
             }
         }
 
+        self.restored_transaction = Some(savepoint.get_transaction_id());
+
         Ok(())
     }
 
@@ -1649,10 +1655,27 @@ impl WriteTransaction {
             self.two_phase_commit = true;
         }
 
+        // The rolled-back commits' pages are queued for freeing by restore_savepoint(), so their
+        // deferred records must not also be written out by a durable commit.
+        if let Some(transaction_id) = self.restored_transaction {
+            self.mem.drop_unpersisted_data_freed_after(transaction_id);
+        }
+
         let (user_root, allocated_pages, data_freed) =
             self.tables.lock().unwrap().table_tree.flush_and_close()?;
 
-        let stored_data_freed_pages = self.store_data_freed_pages(data_freed)?;
+        // A non-durable commit keeps its freed-page records in memory rather than writing them to
+        // DATA_FREED_TABLE, which would mutate the system tree on every commit. durable_commit()
+        // writes the accumulated records out.
+        let stored_data_freed_pages = match self.durability {
+            InternalDurability::None => {
+                let stored_pages = !data_freed.is_empty();
+                self.mem
+                    .record_unpersisted_data_freed(self.transaction_id, data_freed);
+                stored_pages
+            }
+            InternalDurability::Immediate => self.store_data_freed_pages(data_freed)?,
+        };
 
         #[cfg(feature = "logging")]
         debug!(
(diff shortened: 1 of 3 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -419,6 +424,60 @@ impl UnpersistedState {
...
+            return vec![];
+        }
+        self.data_freed
+            .range(start..end)
+            .map(|(id, pages)| (*id, pages.clone()))
+            .collect()
+    }
+
+    /// Replaces a transaction's record with the pages that were not reclaimed.
+    fn retain_data_freed(&mut self, transaction_id: TransactionId, pages: Vec<PageNumber>) {
```
cberner: I think replace would be a better name than retain

src/transactions.rs
```diff
@@ -2324,9 +2364,16 @@ impl WriteTransaction {
...
         assert_eq!(PageNumber::serialized_size(), 8);
 
-        // Handle the data freed tree
+        // Data-tree records are held in memory by non-durable commits, so process them there
+        let oldest_unprocessed = self
+            .transaction_tracker
+            .oldest_unprocessed_non_durable_commit()
+            .unwrap_or(free_until);
         let mut processed =
-            self.process_freed_pages_nondurable_helper(free_until, DATA_FREED_TABLE)?;
```
cberner, line 2329: Is this method still used? It seems like this diff should delete it

src/tree_store/page_store/page_manager.rs
cberner: Agreed — it overwrites the record rather than filtering it in place, so `replace` describes it and `retain` implies the wrong mechanism. Renamed to `replace_data_freed` in 9e10c69, and the doc now also states that an empty replacement drops the record.

---

src/transactions.rs
cberner: It is still used, just for one caller now. `process_freed_pages_nondurable()` handles two tables: this diff moves only the data-tree side to the in-memory records, and the system-tree side still reads `SYSTEM_FREED_TABLE` through this helper (transactions.rs:2380).

Only data-tree records are deferred, because only they are worth deferring: the system tree is written by every commit anyway, and its freed pages are processed on a different schedule. Deleting the helper would mean duplicating the same table walk for the system table.

If you would rather the two paths not look asymmetric, the alternative is to defer the system-tree records the same way and drop the helper entirely — but that widens the change into the system-tree commit path, so I left it out of this PR.

---

src/transactions.rs
```diff
@@ -1649,10 +1655,27 @@ impl WriteTransaction {
...
 
-        let stored_data_freed_pages = self.store_data_freed_pages(data_freed)?;
+        // A non-durable commit keeps its freed-page records in memory rather than writing them to
+        // DATA_FREED_TABLE, which would mutate the system tree on every commit. durable_commit()
+        // writes the accumulated records out.
+        let stored_data_freed_pages = match self.durability {
+            InternalDurability::None => {
+                let stored_pages = !data_freed.is_empty();
+                self.mem
+                    .record_unpersisted_data_freed(self.transaction_id, data_freed);
```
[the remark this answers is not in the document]
cberner: The entry already exists — master carries `* \`Durability::None\` commits are about 2x faster.` from #1318 — and this PR is what makes it true. Adding a second entry would double-count one user-visible change.

The wording was chosen with both changes in view. I flagged on #1318 that 2x overstated that PR alone, and cberner asked for it anyway since this follow-up was planned. Same-machine medians, 20k non-durable commits over a 500k row table:

| | pre-#1318 | master | with this PR |
|---|---|---|---|
| overwrites | 1116 ms | 817 ms | 467 ms |
| fresh inserts | 1235 ms | 915 ms | 559 ms |

#1318 alone is ~1.37x; the pair is ~2.2-2.4x. So the existing entry is accurate once this lands, and inaccurate without it.

On the cost model: there is no behavior change to document beyond the speedup. The records move from the system tree into memory, where the pages a non-durable commit allocated are already tracked, so the accumulation-until-a-durable-commit characteristic is unchanged in kind. If anything the file grows less during a long non-durable run.

---

PR #1356, opened 2026-08-11
Group the state non-durable commits keep in memory

cberner:
Preparation for #1355, which adds a fourth piece of this state. No behavior change: `TransactionalMemory`'s interface is unchanged and `transactions.rs` and `db.rs` are untouched.

`TransactionalMemory` held three separate collections describing what non-durable commits recorded in memory instead of in the file -- the pages they allocated, the per-transaction allocation records standing in for `DATA_ALLOCATED_TABLE`, and a reverse index into those records. Nothing tied them together, so their shared lifecycle was spelled out by hand in the three places that manage it (the constructor, `clear_cache_and_reload()`, `commit()`), and a fourth collection could be added and left out of one of them with no warning. That is not hypothetical: while writing #1355 I added a fourth and missed a site, which surfaced as a confusing assertion failure rather than a compile error.

They now live in one `UnpersistedState` behind a single lock:

- **One lifecycle.** Emptying the state is `clear()`, next to the fields it clears. The two places that abandon or persist the state call it instead of listing collections, so new state is added in one place rather than three.
- **The struct owns its own consistency.** The reverse index is private, and `claim()` drops a page and its allocation record together so the two cannot be left disagreeing. That absorbs `remove_unpersisted_allocation()`, which existed only to keep them in step.
- **One lock instead of three.** `record_allocations()` previously held two of these locks simultaneously; that ordering hazard is gone.

## On the TODO this replaces

The removed TODO suggested moving this state out of `TransactionalMemory` into a separate non-durable commit layer. I looked into that and did not do it; the struct's doc now records why, so the question does not have to be re-derived.

Moving the state out is mechanically possible -- `free_if_unpersisted()` already releases the set lock before freeing, so the removal acts as a claim rather than an atomic remove-and-free, and that survives the state living elsewhere. `TransactionTracker` would even be a plausible home, since it already holds the non-durable commit ids.

The problem is lifecycle. This state is emptied by `commit()` exactly when a commit succeeds, and abandoned by `clear_cache_and_reload()` atomically with the reload. Both are `TransactionalMemory` operations. Moving the state out makes callers responsible for clearing it at precisely the right point, including on error paths -- reintroducing the drift this change is meant to remove, across a layer boundary. Reclaiming a page also has to consult this state and the allocator together.

The fuller version -- a non-durable layer built on top of the durable one -- is worse: non-durable commits touch the allocator, the header slots and the page cache, so a wrapper would need all three exposed as public API. That trades a private field for a much wider public surface.

So the state stays where its lifecycle is, and the change is to name the concern and give it one owner.

## Validation

Full test suite (418 tests), clippy and fmt clean, and a `fuzz_redb` run over the corpus: 125,433 runs, no artifacts.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D6r1JSUJ54zhBzWoVNxsYx)_

### DIFF
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -335,22 +335,94 @@ impl InMemoryState {
     }
 }
 
+/// What non-durable commits record in memory instead of in the file. All of it is volatile by
+/// construction -- a crash rolls those commits back -- so it shares one lifecycle, and `clear()`
+/// is the only way it is emptied: by the durable `commit()` that persists what it stands in for,
+/// or the `clear_cache_and_reload()` that abandons it. It lives here because of that coupling,
+/// and because reclaiming a page consults it and the allocator together.
+#[derive(Default)]
+struct UnpersistedState {
+    // Pages allocated by non-durable commits. Still reclaimable while they are here, since no
+    // durable commit references them.
+    pages: PageNumberHashSet,
+    // Data-tree pages allocated per transaction: the in-memory stand-in for DATA_ALLOCATED_TABLE,
+    // kept in memory so that reclaiming a page can cheaply drop its record. Flushed to the table
+    // by durable_commit().
+    allocations: BTreeMap<TransactionId, PageNumberHashSet>,
+    // Reverse index into `allocations`
+    allocation_txn: PageNumberHashMap<TransactionId>,
+}
+
+impl UnpersistedState {
+    fn clear(&mut self) {
+        self.pages.clear();
+        self.pages.shrink_to_fit();
+        self.allocations.clear();
+        self.allocation_txn.clear();
+    }
+
+    fn contains(&self, page: PageNumber) -> bool {
+        self.pages.contains(&page)
+    }
+
+    fn extend(&mut self, pages: PageNumberHashSet) {
+        self.pages.extend(pages);
+    }
+
+    /// Claims `page` for reclamation, returning whether this call got it. The page and its
+    /// allocation record are dropped together, so a claimed page leaves nothing to write out.
+    fn claim(&mut self, page: PageNumber) -> bool {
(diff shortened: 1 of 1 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -335,22 +335,102 @@ impl InMemoryState {
...
+/// All of it is volatile by construction: a crash rolls those commits back, so whatever stands in
+/// for their on-disk records can be dropped with them. That gives the whole struct one lifecycle,
+/// which is why it is one struct: `clear()` is the only way it is emptied, so a piece of state
+/// added here cannot be left behind by a reload or a durable commit that forgot about it.
+///
+/// It is owned by `TransactionalMemory` because its lifecycle is welded to two of that type's
+/// operations -- a durable `commit()` persists what it stands in for, and `clear_cache_and_reload()`
+/// abandons it -- and because reclaiming a page has to consult it and the allocator together.
+/// It deliberately owns its own consistency: the reverse index is private, so no caller can leave
+/// it disagreeing with `allocations`.
```
cberner: This is too verbose. Trim it down. Match the style of the other parts of the codebase

cberner (commented):
One comment. Otherwise lgtm

src/tree_store/page_store/page_manager.rs
cberner: Trimmed in 3e6ce8a: 12 lines down to 5, matching the density of the `UncommittedPages` doc just above it. It now states what the state is, that its volatility is what gives it one lifecycle via `clear()`, and why it is owned here — dropping the restatement of the drift problem and the note about the private reverse index, which `claim()` already shows. Also cut `claim()`'s doc from 3 lines to 2.

The longer reasoning about not moving this out to a separate layer is in the PR description instead of the source.

---

PR #1358, opened 2026-08-12
Store minimal key separators in branch nodes

cberner:
Branch pages only route lookups: their keys are compared against the query, never deserialized. Storing a whole key there is more than routing needs, and for long keys it is most of what a branch page holds -- with 24 byte keys, a separator costs half the bytes a child entry occupies.

This adds `Key::separator()`, which returns a slice that sorts at or above one key and below the next, and stores its result in branch pages instead of the boundary key.

```rust
fn separator<'a>(left: &'a [u8], right: &'a [u8]) -> &'a [u8] { left }
```

The default returns a whole key, so existing `Key` implementations -- including any that assume branch keys are encoded values -- are unaffected. `&[u8]` overrides it to return the next key truncated one byte past the first that distinguishes them. `&str` and `String` deliberately do not: their `compare()` calls `from_bytes()`, so a truncated separator would panic on invalid UTF-8.

Shortness is the motivation rather than part of the contract -- the trait only requires `left <= result < right`, and the default returns a whole key -- so the method is named for what it produces, not for how well.

## Results

For 5M random 24 byte keys with 150 byte values, ~50% more children fit in each branch page, and the tree loses a level:

| | before | after |
|---|---|---|
| tree height | 5 | **4** |
| branch pages | 6134 | 4076 (-33.5%) |
| leaf pages | 324030 | 323774 |

That level is the whole mechanism: one fewer page fetch and binary search per lookup. Over five iterations each:

| | before | after | change |
|---|---|---|---|
| 1M random gets | 2295-2518 ms | 1847-2051 ms | **-17%** |
| 500K x 10 range scans | 1952-2086 ms | 1663-1814 ms | **-13%** |

`redb_benchmark`'s write phases are dominated by leaf writes and I/O and did not move by more than the noise between runs -- a first pair of runs showed removals -9% and compaction -20%, but a second pair did not reproduce it (baseline bulk load alone moved 49.6s to 33.6s between runs), so I would not claim those.

One phase looks consistently **slower**: `pop` by around 15% (582/566 ms to 687/632 ms across both run pairs). Denser branches would explain it -- each rebuild of a branch page now copies more children.

## Implementation notes

Every place that promotes a boundary into a branch goes through one helper, `branch_separator()`, which also enforces what a `Key` implementation cannot be trusted with: branch keys of a fixed width type are stored at that stride, so a shorter separator would corrupt the page. Fixed width types keep using a whole key regardless of what they return.

One path needs more than the two boundary keys to hand. The cursor's splice path compares a stored separator against the greatest key beneath it to decide whether an ancestor needs rebuilding -- previously `stored == separator`, which worked only because the two were the same thing. They no longer are, so that test would almost always say yes, giving up the cheap child-pointer swap for a full spine rebuild. It now asks the question the separator actually answers, whether the replacement's bound still sorts at or below it, which is both correct and true more often than the equality test it replaces.

`build_replacement_leaves()` has no successor for its last page, so that one separator stays a whole key. An earlier revision substituted the branch's stored separator there; it was dropped as not worth its complexity, since the cost is one long separator per rebuilt branch out of ~80.

## Testing

* `cargo test --all-features`: 440 tests pass. `cargo clippy --all-targets --all-features` clean under `--deny warnings`; `cargo fmt --check` clean; both `no_std` / `thumbv7em-none-eabihf` feature combinations compile clean.
* `cargo fuzz run --sanitizer=none fuzz_redb -- -max_len=10000`: 336,972 runs in 901s, no crashes. Note `fuzz_redb`'s table is keyed by `u64`, a fixed width type, so it covers the refactor but not the `&[u8]` separator path.
* A randomized differential test written for that gap: `&[u8]` keys against a `BTreeMap` reference, with keys shaped to stress separators (long shared prefixes, keys that are prefixes of other keys, empty keys, `0x00`/`0xff` runs), mixing insert/remove/`pop_first`/`pop_last`/`retain`/`extract_from_if`, verified by full forward scan, a point get of every key, random ranges, and `check_integrity()`. 140 seeds passed across several campaigns, 22 of them against the final revision, reaching 255K entries and height 4 (three branch levels), with `retain` repeatedly halving a 400K entry table to exercise merges at depth. Emitting a separator one byte too short makes it fail immediately, so it is not vacuous.

New tests cover the `&[u8]` separator's edge cases, that a leaf split promotes a separator rather than the boundary key, that a fixed width key type cannot install a shorter one, and that branch separators are never passed to `Value::from_bytes()`.

The on-disk layout is unchanged -- branch keys were already variable length for variable width key types, and only their contents differ. A whole key is still a valid separator, so existing databases keep working; `tests/backward_compatibility.rs` covers that direction.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -230,8 +230,11 @@ only after a clean shutdown; the latter is usable even after a crash.
 ## B+tree pages
 
 redb stores data in a B+tree, in which only leaf pages contain key-value pairs; branch pages
-contain only routing keys and pointers to their children. Allocated pages may therefore be of two
-types: B+tree branch pages, or B+tree leaf pages. The format of each is described below:
+contain only routing keys and pointers to their children. A routing key sorts at or above every key
+in the child before it, and below every key in the child after it. Routing keys are never
+deserialized, so they may be prefixes rather than full keys that exist in the tree. Allocated pages
+may therefore be of two types: B+tree branch pages, or B+tree leaf pages. The format of each is
+described below:
 
 ### Branch page:
 * 1 byte: type
--- a/src/tree_store/btree_mutator.rs
+++ b/src/tree_store/btree_mutator.rs
@@ -2,8 +2,8 @@
 use crate::tree_store::btree_base::RawBranchBuilder;
 use crate::tree_store::btree_base::{
     BRANCH, BranchAccessor, BranchBuilder, BranchMutator, Checksum, DEFERRED, LEAF, LeafAccessor,
-    LeafBuilder, LeafMutator, OwnedEntryBuffer, is_single_large_value, leaf_below_merge_threshold,
-    leaf_split_required, retained_after_removals,
+    LeafBuilder, LeafMutator, OwnedEntryBuffer, branch_separator, is_single_large_value,
+    leaf_below_merge_threshold, leaf_split_required, retained_after_removals,
 };
 use crate::tree_store::btree_mutator::DeletionResult::{
     DeletedBranch, DeletedSubtree, PartialBranch, PartialLeaf, Subtree,
@@ -62,9 +62,10 @@ impl DeletedPairs {
     }
 }
 
-// A page produced while splicing an insert run into the tree, with its
-// subtree's greatest key. The key is None only for the node whose subtree
-// contains the tree's original last entry, which stays last at every level.
+// A page produced while splicing an insert run into the tree, with a bound on
+// its subtree: a separator no less than the subtree's greatest key. The bound
+// is None when the parent's stored separator still covers the subtree, which
+// includes the node holding the tree's original last entry.
 #[cfg(feature = "experimental_cursor")]
 type SplicedNode = (PageNumber, Checksum, Option<Vec<u8>>);
 
@@ -255,8 +256,8 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
 
     // Replaces the contiguous `replaced_children` range of the leaf path's
     // parent branch with packed leaves built from `entries`, then rebuilds the
-    // rest of the path. Each replacement's separator is its own greatest key;
-    // separators for preserved children are reused from the original branch.
+    // rest of the path. Separators for preserved children are reused from the
+    // original branch.
     pub(super) fn replace_leaf_children(
         &mut self,
         mut path: Vec<(PageImpl, usize)>,
@@ -329,7 +330,7 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
                 );
                 // Preserved children reuse their original checksum and
                 // separator; freshly built replacements have a deferred
-                // checksum and their greatest key as separator.
+                // checksum and the separators computed above.
                 let preserved = |i: usize| {
                     (
                         accessor.child_page(i).unwrap(),
@@ -382,10 +383,10 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
     }
 
     // Packs the buffered entries into full leaves, in key order, returning
-    // each page with its greatest key as separator. No cleanup on error: any
-    // failure here has latched the storage layer's io_failed flag, which
-    // blocks every later commit, so pages already built are transient
-    // in-memory state reclaimed on rollback.
+    // each page with its separator. No cleanup on error: any failure here has
+    // latched the storage layer's io_failed flag, which blocks every later
+    // commit, so pages already built are transient in-memory state reclaimed
+    // on rollback.
     fn build_replacement_leaves(
         &self,
         buffer: &OwnedEntryBuffer,
@@ -451,14 +452,21 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
             }
             builder
         };
-        // Build the planned pages; each separator key is its page's greatest key.
+        // Build the planned pages. Every page but the last is separated from the one that follows
+        // it; the last has no successor here, so it falls back to its greatest key, which
+        // separates it from whatever the caller places after it.
         let mut leaves = vec![];
-        for range in &plan {
+        for (index, range) in plan.iter().enumerate() {
             let page = fill(range).build()?;
-            leaves.push((page.get_page_number(), entries[range.end - 1].0.to_vec()));
+            let last_key = entries[range.end - 1].0;
+            let separator = match plan.get(index + 1).or(balance_tail.as_ref()) {
+                Some(next) => branch_separator::<K>(last_key, entries[next.start].0),
+                None => last_key,
+            };
+            leaves.push((page.get_page_number(), separator.to_vec()));
         }
         if let Some(range) = balance_tail {
-            let (page1, split_key, page2) = fill(&range).build_split()?;
+            let (page1, split_key, page2) = fill(&range).build_split::<K>()?;
             leaves.push((page1.get_page_number(), split_key.to_vec()));
             leaves.push((page2.get_page_number(), entries[range.end - 1].0.to_vec()));
         }
@@ -487,7 +495,7 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
         let leaves = self.build_replacement_leaves(entries)?;
         let mut nodes: Vec<SplicedNode> = leaves
             .into_iter()
-            .map(|(page, greatest_key)| (page, DEFERRED, Some(greatest_key)))
+            .map(|(page, bound)| (page, DEFERRED, Some(bound)))
             .collect();
         // Freed only after every fallible step, so an error never leaves the
         // surviving root referencing a freed page.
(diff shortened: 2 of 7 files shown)


src/tree_store/btree_mutator.rs
```diff
@@ -449,14 +451,21 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
...
+        // child after it whose least key is not available here.
         let mut leaves = vec![];
-        for range in &plan {
+        for (index, range) in plan.iter().enumerate() {
             let page = fill(range).build()?;
-            leaves.push((page.get_page_number(), entries[range.end - 1].0.to_vec()));
+            let last_key = entries[range.end - 1].0;
+            let separator = match plan.get(index + 1).or(balance_tail.as_ref()) {
+                Some(next) => branch_separator::<K>(last_key, entries[next.start].0),
+                None => last_key,
```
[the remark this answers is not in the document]
cberner: Good catch, fixed in 4e4ccaa.

I had considered reusing the stored separator here and dismissed it, but I only weighed the bytes saved, not the direction of the change. The asymmetry is the actual problem: before this PR both sides of the rebuild were whole keys, so a deletion could not grow the page. With short separators everywhere else and a whole key at this one boundary, it can — and as you note `finalize_branch_builder()` has no split path, so an oversized branch just allocates a larger order page and stays that way.

I took the "reuse the old separator when it still routes" option rather than reading the successor's least key, since it needs no extra page fetch. `replace_leaf_children()` now overwrites the last replacement's separator with `accessor.key(replaced_children.end - 1)` when a preserved child follows. That still routes because every entry fed to `build_replacement_leaves()` came from children in `replaced_children`, so their greatest key is at or below the separator stored after that range — including the neighbor absorbed by the merge-threshold branch above, which widens `replaced_children` before the call. When the replaced range runs to the end of the branch there is no preserved child after it, so the last replacement is the branch's last child and its separator is dropped rather than written.

`build_replacement_leaves()` keeps its greatest-key fallback, since `splice_insert_run()` also calls it and has no stored separator to reuse at that point.

Added a `debug_assert` that the substituted separator sorts at or above the bound it replaces, and re-ran the randomized differential test against a `BTreeMap` with it active: 20 seeds through `retain`/`extract_from_if`/`pop` (which are what reach this path), plus 2 seeds grown to 255K entries and height 4. Full suite still 423 passing.

---

src/tree_store/btree_mutator.rs
```diff
@@ -309,7 +311,19 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
                     }
                 }
             }
-            let replacement_leaves = self.build_replacement_leaves(&entries)?;
+            let mut replacement_leaves = self.build_replacement_leaves(&entries)?;
+            // `build_replacement_leaves` has no successor to separate the last page from, so it
+            // falls back to that page's greatest key. Here there is one: the separator this branch
+            // already stores for the replaced range still routes, since every entry came from
+            // those children, and reusing it keeps a deletion from growing the page.
```
cberner: Make this less verbose

src/types.rs
```diff
@@ -211,6 +211,25 @@ pub trait Key: Value {
...
+    /// `right`, under [`compare()`](Self::compare).
+    ///
+    /// The result is only used to route lookups through the btree's internal nodes: it is passed
+    /// to [`compare()`](Self::compare), and never to [`Value::from_bytes()`], so it need not be a
+    /// valid encoding of `Self`. The shorter it is, the more children fit in each internal node,
+    /// which makes the tree shallower and cheaper to search.
+    ///
+    /// The default implementation returns `left`, which is always valid. Overriding this has no
+    /// effect for a type with a fixed width: internal nodes store their keys at that stride, so a
+    /// shorter separator could not be read back and a same width one would save nothing.
```
cberner: Remove the implementation details about fixed width types

docs/design.md
```diff
@@ -230,8 +230,15 @@ only after a clean shutdown; the latter is usable even after a crash.
...
-types: B+tree branch pages, or B+tree leaf pages. The format of each is described below:
+contain only routing keys and pointers to their children. A routing key separates the two children
+it sits between: it sorts at or above the greatest key beneath the child before it, and below the
+least key beneath the child after it, which is all a lookup needs to choose between them. Because
+they are only ever compared, and never deserialized, a routing key need not be an encoded value of
+the key type. Key types that support it therefore store only as much of the boundary as tells the
+two children apart, which for byte string keys is often a few bytes where a whole key would be far
+longer; the rest store a whole boundary key. The less of each boundary a branch page stores, the
+more children fit in it, and the shallower the tree. Allocated pages may therefore be of two types:
+B+tree branch pages, or B+tree leaf pages. The format of each is described below:
```
cberner: Match the writing of the rest of the design doc better. I think you're over explaining. Just say the routing keys sort at or above, are never deserialized, and may be prefixes rather than full keys that exist in the tree.

src/tree_store/btree_mutator.rs
cberner: Cut to three lines in 5ae6df1. Trimmed the two neighbouring comments I had inflated in the same pass -- `build_replacement_leaves()`'s header and the `SplicedNode` type comment -- since they had the same problem.

---

src/types.rs
cberner: Removed in 5ae6df1; the doc now ends at "The default implementation returns `left`, which is always valid."

Note the behaviour itself is unchanged -- `branch_separator()` still ignores what a fixed width type returns, with the reason in a comment there. Only the public doc mention is gone.

---

docs/design.md
cberner: Cut to your three points in 5ae6df1, eight lines down to three:

> A routing key sorts at or above every key in the child before it, and below every key in the child after it. Routing keys are never deserialized, so they may be prefixes rather than full keys that exist in the tree.

---

src/tree_store/btree_mutator.rs
```diff
@@ -311,7 +312,17 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
...
+            // The last replacement fell back to its greatest key, having no successor to separate
+            // it from. Reuse the separator stored for the replaced range instead: it still routes,
+            // and keeps a deletion from growing a branch that nothing here can split.
+            if replaced_children.end < old_children
+                && let Some((_, separator)) = replacement_leaves.last_mut()
+            {
+                let stored = accessor.key(replaced_children.end - 1).unwrap();
+                debug_assert!(K::compare(separator, stored).is_le());
+                *separator = stored.to_vec();
+            }
```
cberner: I'm confused about this code. Why is this change necessary? It's not clear to me how it is related to the ::separator() code being added.

cberner: It only exists because of `separator()`, and it is not needed for correctness. Happy to drop it.

The connection: before this PR every branch key was a whole key, so rebuilding a branch wrote the same kind of thing it replaced. Now every separator on the page is a few bytes -- except this one. `build_replacement_leaves()` has no successor to separate its last page from, so it falls back to that page's greatest key, and `replace_leaf_children()` writes that into the parent where a short separator used to sit. So the rebuild puts ~24 bytes where ~4 were.

What that costs: `finalize_branch_builder()` either merges or calls `build()` -- it never splits -- and `build()` rounds up to the next page order. So a *deletion* can bump a branch onto a double sized page. Not corruption, and I should correct something I repeated from the bot's comment above: it cannot fail the allocation either. `allocate_helper()` computes `ceil_log2(required_pages)` and grows the file, so it always succeeds. It is also self correcting, since the next rebuild that fits computes a smaller order.

So the real cost is one long separator per rebuilt branch, out of ~80, plus an occasional oversized page. That is small. The eight lines buy back a case where the deletion path partly undoes what the PR is for.

Your call. I'd lean toward dropping it from this PR -- the argument for it is much weaker than the amount of explaining it takes, and "deletions can grow a branch page and nothing splits it" is a pre-existing sharp edge that this change only makes slightly easier to hit. Say the word and I'll remove it, or split it into its own PR alongside a `finalize_branch_builder()` that can split.

---

cberner: ya, drop it

cberner: Dropped in 25beebe. Removed the substitution, the `mut` binding it needed, and the two comments that referred to it; `build_replacement_leaves()` keeps its greatest-key fallback for the last page, as before. Also trimmed the paragraph about it from the commit message and PR description.

Re-verified: 440 tests, clippy clean, plus the differential fuzzer over the deletion paths this touches -- 20 broad seeds and 2 grown to 255K entries at height 4. Those runs report 314/298 branch pages, matching the runs that had the substitution, so it wasn't buying measurable density at this scale either.

---

src/tree_store/btree_mutator.rs
```diff
@@ -451,14 +452,21 @@ impl<'a, 'b, K: Key + 'static, V: Value + 'static> MutateHelper<'a, 'b, K, V> {
...
+        // separates it from whatever the caller places after it.
         let mut leaves = vec![];
-        for range in &plan {
+        for (index, range) in plan.iter().enumerate() {
             let page = fill(range).build()?;
-            leaves.push((page.get_page_number(), entries[range.end - 1].0.to_vec()));
+            let last_key = entries[range.end - 1].0;
+            let separator = match plan.get(index + 1).or(balance_tail.as_ref()) {
+                Some(next) => branch_separator::<K>(last_key, entries[next.start].0),
+                None => last_key,
```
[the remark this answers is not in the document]
cberner: Its absence is deliberate. The substitution was added in 4e4ccaa in response to your earlier comment, then removed in 25beebe because the maintainer asked for it to be dropped -- see the thread at #discussion_r3798969713. Not re-adding it.

Two corrections to this comment. It cannot "fail the larger allocation": `allocate_helper()` computes `ceil_log2(required_pages)` and grows the file, so an oversized branch always allocates, just at a higher order, and the next rebuild that fits drops back down. And the cost is bounded at one long separator per rebuilt branch out of ~80 -- the differential fuzz runs report the same 314/298 branch pages with and without the substitution, so it was not buying measurable density.

The underlying sharp edge -- a deletion can grow a branch page and `finalize_branch_builder()` cannot split -- predates this PR and is better fixed there than worked around here.

---

PR #1371, opened 2026-08-13
Report malformed freed-page records instead of panicking

### LINKED ISSUE #1333: Repair panics on malformed freed-tree entries, leaving the file permanently unopenable; post-panic allocator gap
## Summary

The repair/integrity-check path itself panics on malformed freed-tree entries, which makes an affected file **permanently unopenable by redb** - `Database::open` runs `do_repair`, and `check_integrity` runs the same code, so there is no in-library way to recover. Separately, a caller that `catch_unwind`s that panic is left holding a live `Database` whose allocator state is half-rebuilt, and `begin_write` will proceed with it - a post-panic corruption gap.

Severity: **high**. Found by code review on master @ d08f1dd (release profile); the panicking lines are all reachable in release builds.

### 4a. Panic in repair

`Database::do_repair` -> `rebuild_allocator_state` -> `visit_freed_tree` (`src/db.rs:1093-1100`) iterates the on-disk freed tables and calls `mem.mark_page_allocated(page)` (`src/tree_store/page_store/page_manager.rs:628-635`) for each entry, with no validation of the page number:

```rust
let allocator = state.get_region_mut(region_index);   // Vec index: panics if region >= num_regions
allocator.record_alloc(page_number.page_index, page_number.page_order);
```

- `record_alloc` asserts `order <= self.max_order` (`buddy_allocator.rs:394`) - release-active.
- `record_alloc_inner` -> `BtreeBitmap::get` asserts `bit < self.len` (`bitmap.rs:364`) - release-active.
- Marking the same page twice recurses `record_alloc_inner` past `max_order` and indexes `self.free[order]` out of bounds.

Any of these = panic during `Database::open`. The file keeps `recovery_required` set, so every subsequent open panics the same way; `check_integrity` panics identically. The user is stuck with a file redb cannot open and cannot repair.

Related release-active panics reachable from the system tree on open (clean shutdown, quick-repair path): `AllocatorStateKey::from_bytes` has `_ => unreachable!()` for a crafted key byte (`src/transactions.rs:220`), and `is_valid_allocator_state` unwraps `try_into()` on a `TransactionId` value whose length != 8 (`src/tree_store/page_store/page_manager.rs:753`).

### 4b. Post-panic corruption gap

`rebuild_allocator_state` (`src/db.rs:1048-1107`) first installs a fresh all-free allocator state (`reset_allocator_state`, `page_manager.rs:597-604`) and then marks pages one by one. `check_integrity` invalidates the allocator state only when the check returns `Err` (`src/db.rs:641-645`) - **not when it panics**. A caller that catches the unwind (redb explicitly supports unwind panicking; see the `panic` test at `page_manager.rs:1446-1460`) is left with `allocators = Some(partially populated)`. `begin_write` only checks `allocator_state_loaded()` (is-`Some`, `src/db.rs:1277`), so write transactions proceed with an allocator in which unvisited live pages are still marked free -> double-allocation -> silent corruption.

`commit_inner` already solves this exact problem with `AllocatorStateLatch` (`src/transactions.rs:1635-1644`), which invalidates the allocator state on unwind; the repair path has no equivalent.

## Suggested fix

- Make `mark_page_allocated` / `visit_freed_tree` return `Err(StorageError::Corrupted(..))` instead of panicking (range-check region/index/order against the layout first).
- Guard `rebuild_allocator_state` with a latch that invalidates allocator state on unwind, mirroring `AllocatorStateLatch`.
- Replace the `unreachable!()` / `unwrap()`s in the allocator-state key/value parsing with `Corrupted` errors.

cberner:
Fixes #1333

Rebuilding the allocator state marks every page named by the freed tables as allocated, using page numbers read straight from the file. A malformed one panicked in four different ways, all release-active:

* an out of range **region** indexed `region_allocators` out of bounds (`page_manager.rs:318`)
* an out of range **page index** tripped `BtreeBitmap::get`'s `assert!(bit < self.len)`
* an **order** larger than the region's own `max_order` tripped `record_alloc`'s assertion -- the order check added in #1368 bounds it to `MAX_MAX_PAGE_ORDER`, but a region's maximum is `min(20, log2(capacity))`, so a small or trailing region is lower
* the **same page named twice** walked `record_alloc_inner` up past `max_order`, indexing `self.free[order]` out of bounds

This is on the untrusted side of the line: it is reached only from `Database::open` on an uncleanly shut down file, or from `check_integrity()`. Both run the same repair, so an affected file was permanently unopenable with no way to recover from inside the library.

## Changes

**Validate in `mark_page_allocated()`.** Region and extent are checked against the layout. Unlike the read path -- where this was dropped in #1368 because it needed the `state` lock on `get_page` -- this is repair-only and *already holds that lock*, so the layout is right there and the check is free.

**`record_alloc()` reports failure instead of asserting.** It bounds the order and page index at each level of the split walk, which also catches the duplicate case: marking a page that is already allocated walks up looking for a free parent to split, so running off the top is exactly how a duplicate or overlapping entry shows up. Modifications happen after the recursive call returns, so a rejected page leaves the allocator untouched. The two internal callers in `resize_to` compute their own page numbers, so they assert.

## Tests

`mark_page_allocated_rejects_corrupt_page_numbers` covers the bad region, bad index, bad order and duplicate-page cases. Verified load-bearing: with the validation stripped it panics at `page_manager.rs:318`, the region index, exactly the first site the issue names.

Full suite green, fmt and clippy clean.

## Deliberately left for a follow up

Earlier revisions of this PR also reworked the *saved* allocator state on the quick-repair open path -- replacing the `unreachable!()` in `AllocatorStateKey::from_bytes`, rejecting a partial state whose region keys had been corrupted, and discarding a half-built state if the rebuild unwound. That went through several rounds of review churn and is a distinct problem from the freed-page records this issue reports, so it has been cut back out. What was found there, for the follow up:

* `AllocatorStateKey::from_bytes` has `_ => unreachable!()`, reachable from a corrupt table because `Key::compare` parses every stored key a lookup passes.
* Simply not panicking there is not enough: an unrecognized discriminant sorts below `Region(0)` and so drops out of `load_allocator_state()`'s scan, leaving a region unaccounted for that `resize_to()` refills as free -- a silent version of the same bug. Rejecting the snapshot in `is_valid_allocator_state()` is what fixes that, and a region count alone is not sufficient; the numbering has to be checked too.
* `is_valid_allocator_state` unwraps a `try_into` on a transaction id whose length might not be 8. Every rejection there should be `Ok(false)`, not an error, so the caller falls back to the full repair -- the table is a cache the repair rebuilds without reading.
* `rebuild_allocator_state()` installs an all-free state and marks pages as it walks, so unwinding out of it leaves a half-built one in which live pages look free. Any guard for this must fire on panic only: `repair_live_state()` deliberately keeps going after a corrupt live rebuild, and invalidating on the `Err` path panics `check_integrity()`.
* `BuddyAllocator::from_bytes` slices untrusted bytes unguarded (`data[metadata..]`, `data[data_start..data_end]`) on that same path.

Thanks to the Codex reviews for finding the second and fourth of those.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NeaXwvRvuCWemNbNU9Jjaq)_

### DIFF
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -804,9 +804,31 @@ impl TransactionalMemory {
     pub(crate) fn mark_page_allocated(&self, page_number: PageNumber) -> Result<()> {
         Self::check_page_order(page_number)?;
         let mut state = self.state.lock()?;
-        let region_index = page_number.region;
-        let allocator = state.get_region_mut(region_index);
-        allocator.record_alloc(page_number.page_index, page_number.page_order);
+        // Unlike the read path, this is only reached while rebuilding the allocator state, and the
+        // state lock is already held, so validating against the layout costs nothing here
+        let layout = state.header.layout();
+        if page_number.region >= layout.num_regions() {
+            return Err(StorageError::Corrupted(format!(
+                "Page {page_number:?} is in region {}, but the database has {} region(s)",
+                page_number.region,
+                layout.num_regions()
+            )));
+        }
+        let region_pages = u64::from(layout.region_layout(page_number.region).num_pages());
+        // Cannot overflow: page_index is at most 2^32, and the order was bounded above
+        let end_page = (u64::from(page_number.page_index) + 1) << page_number.page_order;
+        if end_page > region_pages {
+            return Err(StorageError::Corrupted(format!(
+                "Page {page_number:?} extends past the end of its region, which has {region_pages} pages"
+            )));
+        }
+
+        let allocator = state.get_region_mut(page_number.region);
+        if !allocator.record_alloc(page_number.page_index, page_number.page_order) {
+            return Err(StorageError::Corrupted(format!(
+                "Page {page_number:?} overlaps a page that is already allocated"
+            )));
+        }
         #[cfg(debug_assertions)]
         assert!(self.allocated_pages.lock().unwrap().insert(page_number));
 
@@ -1683,6 +1705,56 @@ mod test {
         // The poison stays set, so later accesses fail rather than trust the state
         assert!(mem.state.is_poisoned());
         assert!(mem.get_last_committed_transaction_id().is_err());
+    }
+
+    // Rebuilding the allocator state feeds mark_page_allocated() page numbers straight out of the
+    // freed tables, which no page walk ever reads. A corrupted entry has to be reported rather than
+    // indexing the region allocators, tripping the bitmap's bounds assertion, or walking the buddy
+    // allocator past its maximum order. See https://github.com/cberner/redb/issues/1333
+    #[test]
+    fn mark_page_allocated_rejects_corrupt_page_numbers() {
+        use super::{MAX_PAGE_INDEX, TransactionalMemory};
+        use crate::StorageError;
+        use crate::tree_store::page_store::base::MAX_REGIONS;
+        use crate::tree_store::{InMemoryBackend, PageNumber};
+
+        let page_size = 4096;
+        let mem = TransactionalMemory::new(
+            Box::new(InMemoryBackend::new()),
+            true,
+            page_size,
+            Some(64 * page_size as u64),
+            0,
+            false,
+        )
+        .unwrap();
+        mem.reset_allocator_state().unwrap();
+
+        let corrupt = [
+            // Past the end of the layout, which would index the region allocators out of bounds
+            PageNumber::new(MAX_REGIONS - 1, 0, 0),
+            // Past the end of its region, which the bitmap asserts on
+            PageNumber::new(0, MAX_PAGE_INDEX, 0),
+            // An order no region can have
+            PageNumber::from_le_bytes((31u64 << 59).to_le_bytes()),
+        ];
+        for page in corrupt {
+            assert!(
+                matches!(
+                    mem.mark_page_allocated(page),
+                    Err(StorageError::Corrupted(_))
+                ),
+                "{page:?} was not rejected"
+            );
+        }
+
+        // Naming the same page twice walks the allocator up past its maximum order looking for a
+        // parent to split
+        mem.mark_page_allocated(PageNumber::new(0, 0, 0)).unwrap();
+        assert!(matches!(
+            mem.mark_page_allocated(PageNumber::new(0, 0, 0)),
+            Err(StorageError::Corrupted(_))
+        ));
     }
 
     // A page order read from a corrupted file can be far larger than any real page, which the read
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@
   could be slower than writing from a single thread; they now scale with the number of threads.
 * Fix a panic when opening a database containing a corrupted persistent savepoint record;
   `StorageError::Corrupted` is now returned instead.
+* Fix a panic while repairing a database file containing a malformed freed-page record, which left
+  the file permanently unopenable because both `Database::open` and `check_integrity()` run the same
+  repair. `StorageError::Corrupted` is now returned instead.
 * Fix a process abort when reading a database file whose branch pages form a cycle, or an
   arbitrarily long chain. Descending the tree recursed until the stack overflowed, which aborts the
   process rather than failing; `StorageError::Corrupted` is now returned instead.
--- a/src/tree_store/page_store/buddy_allocator.rs
+++ b/src/tree_store/page_store/buddy_allocator.rs
@@ -262,7 +262,7 @@ impl BuddyAllocator {
                 if processed_pages + order_size > self.len() {
                     break;
                 }
-                self.record_alloc_inner(page, order);
+                assert!(self.record_alloc_inner(page, order));
                 processed_pages += order_size;
             }
             // Allocate the remaining space, at the highest order
@@ -270,7 +270,7 @@ impl BuddyAllocator {
                 let order_size = 2u32.pow(order.into());
                 while processed_pages + order_size <= self.len() {
                     let page = processed_pages / order_size;
-                    self.record_alloc_inner(page, order);
+                    assert!(self.record_alloc_inner(page, order));
                     processed_pages += order_size;
                 }
             }
@@ -389,19 +389,33 @@ impl BuddyAllocator {
         }
     }
 
-    // `page_number` must be free
-    pub(crate) fn record_alloc(&mut self, page_number: u32, order: u8) {
-        assert!(order <= self.max_order);
+    // `page_number` must be free. Returns false if it cannot be marked allocated: the order is too
+    // large for this region, the page is out of range, or the space is already in use. Rebuilding
+    // the allocator state feeds this page numbers read from the file, where any of the three means
+    // the file is corrupt; the callers that compute page numbers themselves assert instead.
+    #[must_use]
+    pub(crate) fn record_alloc(&mut self, page_number: u32, order: u8) -> bool {
         // Split parent pages as necessary, and update the free index
-        self.record_alloc_inner(page_number, order);
+        self.record_alloc_inner(page_number, order)
     }
 
-    pub(crate) fn record_alloc_inner(&mut self, page_number: u32, order: u8) {
+    #[must_use]
+    pub(crate) fn record_alloc_inner(&mut self, page_number: u32, order: u8) -> bool {
+        // Marking a page that is already allocated walks up the orders looking for a free parent to
+        // split, so running off the top is how a duplicate or overlapping page number shows up
+        if order > self.max_order {
+            return false;
+        }
         let allocator = self.get_order_free_mut(order);
+        if page_number >= allocator.len() {
+            return false;
+        }
         if allocator.get(page_number) {
             // Need to split parent page
             let upper_page = next_higher_order(page_number);
-            self.record_alloc_inner(upper_page, order + 1);
+            if !self.record_alloc_inner(upper_page, order + 1) {
+                return false;
+            }
             let allocator = self.get_order_free_mut(order);
 
             let (free1, free2) = (upper_page * 2, upper_page * 2 + 1);
@@ -414,6 +428,8 @@ impl BuddyAllocator {
         } else {
             allocator.set(page_number);
         }
+
+        true
     }
 
     /// data must have been initialized by `Self::init_new()`
@@ -466,7 +482,7 @@ mod test {
         assert_eq!(allocator.count_allocated_pages(), 0);
 
         for page in 0..num_pages {
-            allocator.record_alloc(page, 0);
+            assert!(allocator.record_alloc(page, 0));
         }
         assert_eq!(allocator.count_allocated_pages(), num_pages);
 


src/transactions.rs
```diff
@@ -220,13 +220,21 @@ impl Value for AllocatorStateKey {
...
-            3 => Self::Region(u32::from_le_bytes(data[1..].try_into().unwrap())),
+            3 => match data[1..].try_into() {
+                Ok(region) => Self::Region(u32::from_le_bytes(region)),
+                Err(_) => Self::Deprecated,
+            },
             4 => Self::RegionTracker,
             5 => Self::TransactionId,
-            _ => unreachable!(),
+            _ => Self::Deprecated,
         }
```
cberner, line 230: How is Deprecated handled by the rest of the code base? Is this going to cause future issues by defaulting everything to it?

cberner: Checked rather than assumed. `Deprecated` is almost entirely inert: the only two references in the tree are the `as_bytes` arm and the test I just added. Nothing matches on it, and nothing constructs it to write, so that `as_bytes` arm is already unreachable in practice.

Every consumer of the table is in `page_manager.rs` and asks only for keys it names: point lookups for `Region(i)`, `RegionTracker` and `TransactionId`, plus one range scan over `Region(0)..=Region(u32::MAX)`. The derived `Ord` follows declaration order, so `Deprecated < Region < RegionTracker < TransactionId` -- it sorts strictly below the range scan and can never equal a point lookup. An unrecognized key is therefore skipped by every reader, not misread as one.

**On future issues.** Adding a discriminant later has to come with a file format version bump, since older code could not act on the new state anyway, and `header.rs:488` already rejects any file with `version > FILE_FORMAT_VERSION3` before a single key is parsed. So an old binary meeting a new key fails at the header, not here.

**The one real wart**, worth naming: `compare` is `from_bytes(a).cmp(&from_bytes(b))`, so every discriminant that maps to `Deprecated` compares equal. That was already true of 0, 1 and 2; this widens it to 6..=255. In a corrupt table holding two such keys the ordering is not injective and a lookup could pick either. That only arises in a file we are about to reject, and no reader looks up a `Deprecated` key, so I do not think it matters -- but it is the reason the catch-all is not entirely free.

If you would rather have it explicit, the alternative is a distinct `Unknown` variant ordered above `TransactionId`, which keeps it outside the `Region` range and lets `load_allocator_state` report corruption instead of silently skipping. I did not do that because no consumer currently wants to error on a stray key, so the variant would exist without a user; semantically "a discriminant this version does not write" and "a discriminant redb 2.x wrote" are the same category -- ignore it. Happy to switch if you prefer the explicit form.

---

cberner: Yes to both -- 6 and above stay available, and a future version can define them as real variants.

Nothing about "6 means deprecated" is persisted. `from_bytes` is read-side interpretation in *this* version's code only; a future version ships a `from_bytes` that maps 6 to whatever it wants, and the bytes on disk are unchanged either way. The mapping cannot leak back onto disk either: every insert into the table constructs its key explicitly (`Region(i)`, `RegionTracker`, `TransactionId` at `page_manager.rs:875-886` and `919-933`), and nothing ever re-inserts a key it read, so the `as_bytes(Deprecated)` arm can never rewrite a stray discriminant as 0.

Worth adding to what I said above: `Deprecated` is *already* unreachable in a file we will open. Discriminants 0-2 only exist in v1/v2 files, and `TransactionHeader::from_bytes` (`header.rs:482`) returns `UpgradeRequired` for those before a key is ever parsed -- the same gate that returns `Corrupted` for a version above 3. So both the existing `Deprecated` bucket and the new catch-all only ever fire on a *corrupt v3 file*, where the only goal is not to panic.

**The one thing that does change**, and it is worth deciding deliberately: if a future version adds a discriminant *without* bumping the file format version, an old binary now silently skips that key instead of panicking. The panic would have caught the mistake in testing; silence would not. The version gate is what makes that a non-issue, so this is only a hazard if someone later reasons "old versions ignore unknown keys, so no bump needed."

If you want that failure mode to stay loud without being a panic, the clean version is a distinct `Unknown` variant plus having `is_valid_allocator_state` treat its presence as "this saved state cannot be trusted" -- it already returns `false` for a stale transaction id, which falls back to a full repair. Then a corrupt table gets rebuilt, and a future-key file read by an old binary degrades to a repair instead of silently misreading. The cost is that spotting a stray key needs a scan of the table rather than the current point lookup. Say the word and I will do it; otherwise I will leave the catch-all as is.

---

cberner: Good question to push on -- checking it properly showed that most of what I wrote here was dead. Cut down to one line in 6f7de62.

**What it fixes:** exactly one reachable panic, the `_ => unreachable!()`. A btree lookup compares the query against the *stored* key bytes -- `compare` is `from_bytes(a).cmp(&from_bytes(b))` -- so `is_valid_allocator_state`'s `tree.get(&TransactionId)` runs `from_bytes` over every stored key it passes on the way down. A key whose discriminant is not 0-5 panics there. That is on the quick-repair open path, so it is a *cleanly shut down* file that panics in `Database::open` with no way past it -- the same permanently-unopenable shape as the freed-table entries this PR is really about, just a narrower target.

**What was dead, and is now gone:** the length guards. `AllocatorStateKey::fixed_width()` is `Some(5)`, and for a fixed-width key `LeafAccessor::key_end(n)` is `key_section_start() + 5 * (n + 1)` (`btree_base.rs:550`), so the slice handed to `from_bytes` is always exactly 5 bytes. Neither the empty-slice check nor the fallible `try_into` on `data[1..]` can ever fire. I have dropped both, and the two test cases that covered them.

The remaining diff in this file is the `unreachable!()` arm becoming `Deprecated`, merged with the existing `0..=2` arm since clippy correctly pointed out they are now the same body.

One caveat on the evidence: the mechanism is certain from the code, but I did not build an end-to-end corrupted file to prove a byte flip lands there -- I tried, and locating the allocator state table's leaf by scanning for the key bytes was not reliable enough to turn into a test. So this arm has no regression test, unlike the freed-table validation, which does.

---

src/db.rs
```diff
@@ -1083,6 +1083,11 @@ impl Database {
         repair_callback: &(dyn Fn(&mut RepairSession) + 'static),
     ) -> Result<[Option<BtreeHeader>; 2], DatabaseError> {
         mem.reset_allocator_state()?;
+        // From here the allocator state is a half-built description of the file: every page starts
+        // out free and is marked allocated as it is visited. Unwinding out of this without
+        // discarding it would leave live pages looking free, which a later write would hand out
+        // again. Mirrors the latch commit_inner() uses for the same reason.
+        let latch = AllocatorStateLatch::arm(mem.clone());
```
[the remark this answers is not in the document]
cberner: You're right, and it was a regression I introduced. `repair_live_state()` maps `Err(Corrupted)` to `Ok(None)` on purpose so the caller falls through to the durable state, and the latch cleared the allocators on that path, so `check_integrity_inner()`'s later `allocator_hash()` panicked on `expect("allocators have not been loaded yet")`.

Fixed by narrowing the guard to fire only while unwinding, since the gap this is closing is the panic one — every `Err` return out of `rebuild_allocator_state()` is one a caller asked for and already handles:

```rust
struct DiscardOnPanic<'a>(&'a TransactionalMemory);

impl Drop for DiscardOnPanic<'_> {
    fn drop(&mut self) {
        if crate::panicking() {
            self.0.invalidate_allocator_state();
        }
    }
}
```

`AllocatorStateLatch` is back to private and untouched.

---

src/transactions.rs
```diff
@@ -221,12 +221,14 @@ impl Value for AllocatorStateKey {
...
-            0..=2 => Self::Deprecated,
             3 => Self::Region(u32::from_le_bytes(data[1..].try_into().unwrap())),
             4 => Self::RegionTracker,
             5 => Self::TransactionId,
-            _ => unreachable!(),
+            // 0, 1, 2 were used in redb 2.x and have a different format. Any other discriminant
+            // appears only in a corrupted table, which a btree lookup still reaches, since it
+            // compares the query against the stored key bytes. Both sort below every key this
+            // version writes, so the scans that load the allocator state skip them.
+            _ => Self::Deprecated,
```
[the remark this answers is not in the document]
cberner: Correct, and it matters more than the panic this arm replaced — a panic is loud, a partial allocator state is not. Fixed in `14e9179`.

I also had the reason wrong in my own comment here. `Key::compare` is `from_bytes(a).cmp(&from_bytes(b))`, so it compares the *parsed* keys, not the stored bytes; an unrecognized discriminant becomes `Deprecated`, the smallest variant, and sorts **below** `Region(0)` rather than above it. That is exactly why the scan drops it. (It also corrects what I said further up this PR about discriminants ≥ 6 sorting above the keys this version writes — they do not, though a future version defining them is still fine, since that version ships its own `from_bytes` and the on-disk bytes are unchanged.)

The fix is in `is_valid_allocator_state`, so a rejected state returns `Ok(false)` → `get_allocator_state_table` returns `None` → full repair, rather than an error the caller has to handle:

```rust
let num_regions = self.state.lock()?.header.layout().num_regions();
// Not holding the state lock: reading the tree below faults pages in, which takes it
let mut regions = 0u32;
for entry in tree.range(&(AllocatorStateKey::Region(0)..=AllocatorStateKey::Region(u32::MAX)))? {
    if entry?.key() != AllocatorStateKey::Region(regions) {
        return Ok(false);
    }
    regions += 1;
}
Ok(regions == num_regions && tree.get(&AllocatorStateKey::RegionTracker)?.is_some())
```

Checking the numbering as well as the count catches a region lost from the middle, where the count alone would not. `try_save_allocator_state` writes `Region(0)`..`Region(num_regions - 1)` and the tracker, or nothing at all, so anything else is corrupt.

The `RegionTracker` half is the same defect one key over: `load_allocator_state` does `tree.get(&RegionTracker)?.unwrap()`, which a corrupted discriminant hides just as it does a region — that unwrap was reachable from the same crafted file.

Tested end to end in `tests/corrupted_allocator_state.rs`: build a database, close it cleanly, corrupt the `Region(0)` key's discriminant to 6, and assert the read-only open reports `RepairAborted` instead of loading. I verified it is load-bearing — with the check stubbed out to `Ok(true)` it fails. Read-only open is the probe because it turns a rejected state into exactly one error and skips the debug-only page walk that arbitrary corruption trips first. Two notes on what it does not assert: the corrupted page also fails checksum verification during the subsequent repair, so the file is not recoverable in this particular scenario and the test does not claim it is; and the search corrupts every occurrence of the key's byte pattern, since only one of them is the real key.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -928,10 +950,42 @@ impl TransactionalMemory {
         let Some(value) = tree.get(&AllocatorStateKey::TransactionId)? else {
             return Ok(false);
         };
-        let transaction_id =
-            TransactionId::new(u64::from_le_bytes(value.value().try_into().unwrap()));
+        let Ok(raw_id) = value.value().try_into() else {
+            return Err(StorageError::Corrupted(
+                "Invalid transaction id in the allocator state table".to_string(),
+            ));
```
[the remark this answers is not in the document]
cberner: Agreed — and it was inconsistent with the rest of the same function, which returns `Ok(false)` for a missing transaction id, a stale one, a missing or misnumbered region, and a missing region tracker. A malformed id was the only case I made an error. Fixed in `8307c4c`:

```rust
// A malformed one is no more usable than a stale one, and no less repairable
let Ok(raw_id) = value.value().try_into() else {
    return Ok(false);
};
```

Your reasoning holds up against the repair path: `do_repair` rebuilds the allocator state by walking the roots and the freed tables, and never reads this table, so nothing about the file is unrecoverable because of a bad byte in a cached snapshot. The read-only open also gets a more accurate answer — `RepairAborted` ("this needs a repair I cannot perform") rather than `Corrupted`.

---

PR #1374, opened 2026-08-13
Call StorageBackend::close() when opening a database fails

cberner:
Split out of #1372 at your request, so that one stays documentation only.

`close()` is documented to be called exactly once, but that only held for a database that finished opening. `TransactionalMemory::new` takes the backend and hands it to `PagedCachedFile`, and every error between there and `Database::new` constructing `db` returns with the backend dropped and never closed:

- a magic number that does not match (`page_manager.rs`, "Not a redb database")
- an empty file when `create` was not requested
- any `?` on the repair path, `begin_writable()`, or the header read
- the persistent savepoint restore below `db.rs:1213` -- this one *does* close today, since `db` is live and its `Drop` runs, which is what makes the current behaviour inconsistent rather than merely absent

A backend that releases a file lock, powers down a device, or frees a peripheral in `close()` never got the chance on any of the others.

### The fix

`CheckedBackend` owns the `Box<dyn StorageBackend>` and already tracks whether it has been closed, so closing from its `Drop` covers all of those at once:

```rust
impl Drop for CheckedBackend {
    fn drop(&mut self) {
        if !self.closed.load(Ordering::Acquire) {
            let _ = self.file.close();
        }
    }
}
```

That is one place instead of the twenty-odd early returns, and it covers any added later for free. `close()` sets the flag, so a database that shuts down normally still closes exactly once -- `close_called_exactly_once_on_shutdown_io_error` is what pins that half, and it still passes.

The error is discarded. `Drop` cannot return it, and on the paths this affects the open is already failing with an error worth more than "the close also failed".

### Testing

`close_called_exactly_once_when_open_fails` covers a backend rejected for its magic number and one whose very first `len()` fails. It fails without the `Drop` impl, so it is not vacuous.

Full suite passes with `--all-features`, `cargo clippy --all --all-targets --all-features` is clean, `cargo fmt --check` and the ASCII check pass.

Once this lands, the `close()` note in #1372 can drop its "not guaranteed if opening the database failed" caveat.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M7DhuGXp5hmdwuRSRFYGtV)_

### DIFF
--- a/src/tree_store/page_store/cached_file.rs
+++ b/src/tree_store/page_store/cached_file.rs
@@ -126,6 +126,16 @@ struct CheckedBackend {
     closed: AtomicBool,
 }
 
+// Covers the open paths that fail before there is a Database to drop. Drop cannot report the
+// error, but the open is already failing.
+impl Drop for CheckedBackend {
+    fn drop(&mut self) {
+        if !self.closed.load(Ordering::Acquire) {
+            let _ = self.file.close();
+        }
+    }
+}
+
 impl CheckedBackend {
     fn new(file: Box<dyn StorageBackend>) -> Self {
         Self {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -165,6 +165,9 @@
   leaf page, instead of splitting the full one in half and leaving space behind that no later
   insert can use. A table loaded in key order occupies about half as many pages as before, and
   loads faster.
+* `StorageBackend::close()` is now also called when opening a database fails, so that a backend
+  can release what it holds on that path. Previously it was called only for a database that
+  finished opening.
 
 ### redb-derive (unreleased)
 * Fix the derived implementations calling inherent methods named `fixed_width`, `from_bytes`,
--- a/src/db.rs
+++ b/src/db.rs
@@ -64,8 +64,8 @@ pub trait StorageBackend: 'static + Debug + Send + Sync {
     /// Release any resources held by the backend
     ///
     /// Note: redb will not access the backend after calling this method and will call it exactly
-    /// once when the database is closed: when the [`Database`] is dropped, or, if a
-    /// [`WriteTransaction`] was live at that point, when that transaction completes
+    /// once: when the [`Database`] is dropped, or, if a [`WriteTransaction`] was live at that
+    /// point, when that transaction completes, or if opening the database fails
     fn close(&self) -> core::result::Result<(), io::Error> {
         Ok(())
     }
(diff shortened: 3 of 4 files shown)


src/tree_store/page_store/cached_file.rs
```diff
@@ -126,6 +126,18 @@ struct CheckedBackend {
     closed: AtomicBool,
 }
 
+// Opening a database can fail after the backend has been handed over but before there is a
+// Database to drop, and every one of those paths would otherwise drop the backend without closing
+// it. Closing here instead of at each of them makes close() unconditional, so a backend can always
+// release what it holds there. The error has nowhere to go, but the open is already failing.
```
cberner: Too verbose

cberner (commented):
Comments need cleanup. Otherwise lgtm

src/tree_store/page_store/cached_file.rs
cberner: Cut from four lines to two in `afd36b7`:

```rust
// Covers the open paths that fail before there is a Database to drop. Drop cannot report the
// error, but the open is already failing.
```

The three lines I dropped restated what the `Drop` impl and the commit message already say. What was worth keeping is the discarded error, since `let _ =` on its own invites someone to "fix" it.

That is every comment in this PR bar three one-line labels in the test -- the two case markers and the header -- which I left since they name what each fixture is for rather than explaining the code.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -126,6 +126,18 @@ struct CheckedBackend {
...
+impl Drop for CheckedBackend {
+    fn drop(&mut self) {
+        if !self.closed.load(Ordering::Acquire) {
+            let _ = self.file.close();
```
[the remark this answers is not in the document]
cberner: Both added in `e7877f7`. You are right that the PR was changing observable behaviour while leaving the doc that describes it untouched.

The `close()` doc said the call happens "when the database is closed", which after this change is no longer the only time:

```diff
-    /// once when the database is closed: when the [`Database`] is dropped, or, if a
-    /// [`WriteTransaction`] was live at that point, when that transaction completes
+    /// once: when the [`Database`] is dropped, or, if a [`WriteTransaction`] was live at that
+    /// point, when that transaction completes, or if opening the database fails
```

`CHANGELOG.md` gets an entry under 4.2.0 rather than 5.0.0, since this is not behind `experimental-api-5`:

> `StorageBackend::close()` is now also called when opening a database fails, so that a backend can release what it holds on that path. Previously it was called only for a database that finished opening.

One caveat for whoever merges: #1372 edits the same doc block, adding a note below this paragraph that says `close()` is *not* guaranteed on a failed open -- true today, false once this lands. Whichever merges second needs that note dropped, and I will push it rather than leave the two inconsistent.

---

PR #1378, opened 2026-08-14
Update the post-commit epilogue's pages in place

### LINKED ISSUE #829: Writes don't reuse freed pages from the previous transaction
Thanks for writing redb!

A sequence of large write transactions seems to consume twice as much disk space as I was expecting.  If each transaction modifies 1 GB of existing pages (and there are no live reads), I was expecting 1 GB of overhead: the old data needs to be kept around until the transaction has committed.

But in fact this has 2 GB of overhead.  The first transaction allocates 1 GB of new pages, and adds 1 GB of old pages to the freed tree, as expected.  The second transaction correctly frees those old pages -- but they're freed at the *end* of the transaction, rather than the beginning.  So even though the pages are no longer referenced by anything and could be reused, the second transaction doesn't use them; instead, it allocates a fresh 1 GB of pages, doubling the disk usage.  The original pages don't get reused until the third transaction[*].

It's possible to work around this, at the cost of some extra latency, by adding an extra empty write transaction after each real write.  But it's kind of surprising that that's necessary, and it would be nice if it weren't needed!

This also interacts with another issue, where live reads are blocking page freeing for one transaction longer than they should.  I'll submit that one separately.

[*] A few pages do actually get reused in the second transaction, to store the updated freed tree.  But they don't get reused for data.

cberner:
`f9a348d` (#1201, which fixed #829) added a post-commit epilogue that leaves behind a system root no durable commit has written yet. The next durable commit treats that root's pages as committed, so it copies them and writes **both** versions out. Every small durable commit has paid about two extra backend writes since.

This adopts those pages instead, so the durable system-tree updates overwrite them in place rather than copying a root that was never on disk. 50 lines of source, one commit.

## Effect on #829

The issue is about large transactions doubling disk usage. Repeated full overwrite of 200 MiB of live data, file size at steady state:

| | file | overhead |
|---|---|---|
| `f9a348d~1` (the bug) | 1028 MiB | 5.14x |
| master (the fix) | 514 MiB | 2.57x |
| this PR | 514 MiB | 2.57x |

Unchanged. The epilogue still runs and still processes `DATA_FREED_TABLE`, so pages freed by one transaction remain available to the next. Only the epilogue's own allocations are treated differently.

## Effect on commit speed

**Small durable commits.** This is what the change is for. 2,000 single-key durable commits against a file backend, counted with `strace`:

| | `pwrite64` | per commit | `fdatasync` |
|---|---|---|---|
| `f9a348d~1` | 10,033 | 5.0 | 2,014 |
| master | 16,040 | 8.0 | 2,014 |
| this PR | 12,035 | 6.0 | 2,014 |

25% fewer writes, and 10-20% faster on `redb_benchmark`'s individual-write case. `fdatasync` is untouched, so commit *latency* is unchanged -- this is write amplification and device wear.

**Large transactions.** Neutral: 12 large commits interleaved against master, 12.39s total for master and 12.48s for this branch.

**Non-durable commits.** A cost rather than a win, and the one place this is not neutral-or-better. 100,000 `Durability::None` commits run at ~9.6 us/commit here against ~9.1 on master -- about 5%, reproducible across interleaved runs rather than a single sample. It is not work added to that path: adoption never runs there, and building this branch with the one line it does add to it removed (`UnpersistedState::claim()` maintaining `post_commit_allocations`) measures the same. Allocations per commit are identical at 2.00 on both, so it is where pages land rather than how many: updating the epilogue's pages in place changes the layout the following commits inherit. For scale, `f9a348d~1` runs that same workload at ~21.9 us/commit, so this gives back roughly 5% of a 2.4x improvement.

The remaining write per durable commit would need the epilogue to stop writing `SYSTEM_FREED_TABLE`, which means holding those pages in memory until both durable slots have moved past the root that names them, threaded through the quick-repair allocator snapshot, the `check_integrity()` rebuild, and `compact()`'s drain loop. Written and verified, but one write per commit is not worth that, so it is dropped rather than deferred.

## Why it is safe

From the adoption point a durable commit is irreversible: it either completes or leaves the allocator state invalidated by `AllocatorStateLatch`, so nothing rolls back what it claims. `commit()` sets `completed` before anything fallible, so `Drop` never reaches `abort_inner()`, and there is no path from adoption back to `rollback_all()`.

No durable root names the adopted pages -- they were allocated after the previous commit's header was written and synced -- so a crash mid-commit recovers the previous root intact. Read transactions capture only the data root and never descend the system tree, so no reader can observe the rewrite. Every other `get_system_root()` caller is a repair or verify path holding exclusive access.

Savepoints cannot reach an adopted page either: a `Savepoint` stores only `user_root`, and `restore_savepoint` works off `DATA_ALLOCATED_TABLE` and `unpersisted_allocations_after`, which reads the data-tree allocation map. Adopted pages live in `unpersisted.pages` and never enter it.

Adoption *claims* each page out of the unpersisted set rather than only testing that it is still in it. From that point the page is an ordinary uncommitted page of the commit, and a copy-on-write may free it -- `free_if_uncommitted` reaches `TransactionalMemory::free`, and `free_helper` does not maintain that set. An entry left behind would leave a page number free and reusable within the same commit while still marked unpersisted, which is what the subset invariant on `post_commit_allocations` says cannot happen.

## Testing

`durable_overwrite_does_not_write_post_commit_metadata` counts backend writes across 20 durable overwrites: 119 here against 156 on master. It fails without the change.

`cargo test --all-features` passes in debug and release -- release matters here, since the claim above is an `assert!` rather than a `debug_assert!` and would fire there. `cargo clippy --all --all-targets --all-features`, `cargo doc` and `cargo fmt --check` are clean under `--deny warnings`; `wasm32-unknown-unknown` and `thumbv7em-none-eabihf` check clean.

`cargo fuzz run fuzz_redb -max_len=10000`: 105,201 runs against this exact commit with no crashes, plus 137,795 and 465,976 against the two previous revisions and 310,556, 531,570 and 1,176,032 against earlier ones. The target mixes durabilities, `quick_repair`, two-phase commits, savepoints, `compact()`, `check_integrity()` and simulated crashes.

A separate stress harness verifies a shadow model after every transaction while mixing durabilities, quick-repair, two-phase commits, ephemeral savepoints held across transactions and restored, aborts, and three reader threads scanning concurrently (17k-27k scans per run), with periodic `check_integrity()`, `compact()` and reopen. Clean in debug and release across several seeds.

Compared against master directly: `compact()` reaches the same size in the same number of calls; crash-and-reopen at 101 and 503 commits over mixed durable / `Durability::None` / quick-repair workloads gives identical contents with `check_integrity()` clean; and forcing recovery through `repair_primary_corrupted()` by flipping the god byte recovers the same committed state.

## What the invariants are worth

Mutation testing shows the adoption *placement* is load-bearing: moving it somewhere `rollback_all()` can reach wedges the stress harness immediately (1s to complete becomes a hang past 200s).

Claiming rather than testing is load-bearing too, and has a signal. Adding `assert!(!self.mem.unpersisted(page))` to `free_if_uncommitted` -- which asserts exactly the invariant -- separates the three states cleanly:

| | result |
|---|---|
| master | 94,067 fuzz runs, clean |
| an earlier revision, testing instead of claiming | panics within seconds |
| this PR | 102,292 fuzz runs, clean |

I could not make it *observable* without that added assertion. The window runs from the copy to `mem.commit()`'s `unpersisted.clear()`; the durable path passes `None` to `store_system_freed_pages`, so nothing reads `unpersisted` inside it, and `process_freed_pages` runs before the window opens. So it was a latent invariant violation that held by an ordering accident, not a bug you could trigger. It now holds by construction.

The `post_commit_allocations` line in `UnpersistedState::claim()` is what lets adoption assert rather than branch: `pages` only ever shrinks in `clear()` and `claim()`, and both drop the page from `post_commit_allocations` at the same time, so nothing can reach adoption having already left the unpersisted set. I could not falsify that line directly -- instrumenting it shows the recycled-page path fires zero times in the suite, the fuzzer and the stress harness -- but it is no longer insurance against an unreachable state, since the assert depends on it.

## Not covered

On a database of a few pages, the file's floor is higher than master's, because `try_shrink` only trims trailing free pages and an in-place page never migrates downward the way a copied one does. It is a constant, not proportional to anything: after 4,000 single-key durable commits master sits at 90,112 bytes and this branch at 307,200, and that same figure holds whether the workload touches 1 key or 65,536. At 200 MiB of live data both sit at 514 MiB. Nothing is leaked -- `allocated_pages()` matches and `compact()` reaches the same size.

Same root cause as the non-durable commit cost above: in-place updating keeps pages where they are, and copying quietly migrated them downward. It is also the only user-visible change this PR introduces against v4.1.0 on its own -- everything else a 4.2.0 user would notice on these paths comes from `f9a348d` -- so it is the one candidate for a changelog entry if you want one.

## Dropped from this PR

An earlier revision also batched the allocator lock in `process_freed_pages()`. Measured against the case it was written for, it made no reliable difference -- 419k pages: 1052/1062 ms master vs 1078/1107 ms batched; 838k: 4175/4175/4355 vs 4192/4014/3993; 1.7M: 8140 vs 8445. The delete is dominated by walking the tree and writing pages, not by acquiring the allocator lock. It has been removed rather than landed unproven.

A second revision generalized adoption to all non-durable commits, with a separate `PageAllocator` for the system tree and a maintained unpersisted-system-page set, which deletes the `extract_if`/`post_commit_frees`/`free_if_unpersisted` reclaim dance in `non_durable_commit`. It measured 1-2 us/commit *slower* on every path it targets and changed no write count, because under `Durability::None` the system tree is already never written -- freed-page records are held in memory precisely so a non-durable commit does not mutate it. Details in [this comment](https://github.com/cberner/redb/pull/1378#issuecomment-5304436687).

### DIFF
--- a/src/transactions.rs
+++ b/src/transactions.rs
@@ -1672,6 +1672,14 @@ impl WriteTransaction {
         let (user_root, allocated_pages, data_freed) =
             self.tables.lock().unwrap().table_tree.flush_and_close()?;
 
+        // From here a durable commit is irreversible: it either completes or leaves the allocator
+        // state invalidated, so nothing rolls back what it claims. That is what lets it take over
+        // the epilogue's pages, which no durable root names and it is about to supersede.
+        if self.durability == InternalDurability::Immediate {
+            let pages = self.mem.take_post_commit_allocations();
+            self.page_allocator().adopt_unpersisted(pages);
+        }
+
         // A non-durable commit keeps its freed-page records in memory rather than writing them to
         // DATA_FREED_TABLE, which would mutate the system tree on every commit. durable_commit()
         // writes the accumulated records out.
@@ -2066,6 +2074,8 @@ impl WriteTransaction {
         };
 
         let epilogue_allocations = page_allocator.take_allocated_since_commit();
+        self.mem
+            .record_post_commit_allocations(epilogue_allocations.iter().copied());
         self.mem.non_durable_commit(
             user_root,
             system_root,
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -191,6 +191,17 @@ impl PageAllocator {
         self.allocated_since_commit.take_all()
     }
 
+    // Takes ownership of pages an earlier transaction allocated, so this one can update them in
+    // place. Only sound once this transaction can no longer abort, since rollback_all() frees
+    // whatever is adopted. The page leaves the unpersisted set: it is an ordinary uncommitted page
+    // from here, and the copy-on-write that may free it does not maintain that set.
+    pub(crate) fn adopt_unpersisted(&self, pages: impl IntoIterator<Item = PageNumber>) {
+        for page in pages {
+            assert!(self.mem.claim_unpersisted(page));
+            self.allocated_since_commit.insert(page);
+        }
+    }
+
     /// Reverses every allocation made since the last commit: drains the
     /// allocated-since-commit set and frees each page.
     pub(crate) fn rollback_all(&self) {
@@ -354,6 +365,10 @@ struct UnpersistedState {
     // here for the same reason as `allocations` -- the commits that freed them are volatile, so a
     // crash that loses the records also rolls back the commits they describe.
     data_freed: BTreeMap<TransactionId, Vec<PageNumber>>,
+    // System-tree pages allocated by the post-commit epilogue, which the next durable commit may
+    // take over once it is irreversible. Always a subset of `pages`, so a reused page number is
+    // never mistaken for the allocation that held it before.
+    post_commit_allocations: PageNumberHashSet,
 }
 
 impl UnpersistedState {
@@ -363,6 +378,7 @@ impl UnpersistedState {
         self.allocations.clear();
         self.allocation_txn.clear();
         self.data_freed.clear();
+        self.post_commit_allocations.clear();
     }
 
     fn contains(&self, page: PageNumber) -> bool {
@@ -379,6 +395,8 @@ impl UnpersistedState {
         if !self.pages.remove(&page) {
             return false;
         }
+        // Keeps the subset invariant; see the field.
+        self.post_commit_allocations.remove(&page);
         if let Some(txn) = self.allocation_txn.remove(&page) {
             let pages = self
                 .allocations
@@ -1279,6 +1297,11 @@ impl TransactionalMemory {
         self.storage.cancel_pending_write(address_range.start, len);
     }
 
+    // Drops the page from the unpersisted set without freeing it. Returns whether it was there.
+    pub(crate) fn claim_unpersisted(&self, page: PageNumber) -> bool {
+        self.unpersisted.lock().unwrap().claim(page)
+    }
+
     // Frees the page if no durable commit has occurred, since it was allocated. Returns true, if the page was freed
     pub(crate) fn free_if_unpersisted(&self, page: PageNumber, allocated: &PageTracker) -> bool {
         if self.unpersisted.lock().unwrap().claim(page) {
@@ -1307,6 +1330,21 @@ impl TransactionalMemory {
         &self,
     ) -> BTreeMap<TransactionId, PageNumberHashSet> {
         self.unpersisted.lock().unwrap().take_allocations()
+    }
+
+    pub(crate) fn record_post_commit_allocations(
+        &self,
+        pages: impl IntoIterator<Item = PageNumber>,
+    ) {
+        self.unpersisted
+            .lock()
+            .unwrap()
+            .post_commit_allocations
+            .extend(pages);
+    }
+
+    pub(crate) fn take_post_commit_allocations(&self) -> PageNumberHashSet {
+        mem::take(&mut self.unpersisted.lock().unwrap().post_commit_allocations)
     }
 
     // Returns all unpersisted data-tree pages allocated strictly after `transaction_id`. Used
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -177,6 +177,35 @@ impl StorageBackend for SharedInMemoryBackend {
     }
 }
 
+#[derive(Debug)]
+struct WriteCountingBackend {
+    inner: InMemoryBackend,
+    writes: Arc<AtomicU64>,
+}
+
+impl StorageBackend for WriteCountingBackend {
+    fn len(&self) -> Result<u64, std::io::Error> {
+        self.inner.len()
+    }
+
+    fn read(&self, offset: u64, out: &mut [u8]) -> Result<(), std::io::Error> {
+        self.inner.read(offset, out)
+    }
+
+    fn set_len(&self, len: u64) -> Result<(), std::io::Error> {
+        self.inner.set_len(len)
+    }
+
+    fn sync_data(&self) -> Result<(), std::io::Error> {
+        self.inner.sync_data()
+    }
+
+    fn write(&self, offset: u64, data: &[u8]) -> Result<(), std::io::Error> {
+        self.writes.fetch_add(1, Ordering::SeqCst);
+        self.inner.write(offset, data)
+    }
+}
+
 /// Returns pairs of key, value
 fn random_data(count: usize, key_size: usize, value_size: usize) -> Vec<(Vec<u8>, Vec<u8>)> {
     let mut pairs = vec![];
@@ -604,37 +633,8 @@ fn non_durable_commit_persistence() {
 // in the in-memory cache, visible to readers, and only written out by a later durable commit
 #[test]
 fn non_durable_commit_issues_no_backend_writes() {
-    #[derive(Debug)]
-    struct CountingBackend {
-        inner: InMemoryBackend,
-        writes: Arc<AtomicU64>,
-    }
-
-    impl StorageBackend for CountingBackend {
-        fn len(&self) -> Result<u64, std::io::Error> {
-            self.inner.len()
-        }
-
-        fn read(&self, offset: u64, out: &mut [u8]) -> Result<(), std::io::Error> {
-            self.inner.read(offset, out)
-        }
-
-        fn set_len(&self, len: u64) -> Result<(), std::io::Error> {
-            self.inner.set_len(len)
-        }
-
-        fn sync_data(&self) -> Result<(), std::io::Error> {
-            self.inner.sync_data()
-        }
-
-        fn write(&self, offset: u64, data: &[u8]) -> Result<(), std::io::Error> {
-            self.writes.fetch_add(1, Ordering::SeqCst);
-            self.inner.write(offset, data)
-        }
-    }
-
     let writes = Arc::new(AtomicU64::new(0));
-    let backend = CountingBackend {
+    let backend = WriteCountingBackend {
         inner: InMemoryBackend::new(),
         writes: writes.clone(),
     };
@@ -677,6 +677,47 @@ fn non_durable_commit_issues_no_backend_writes() {
     }
     txn.commit().unwrap();
     assert!(writes.load(Ordering::SeqCst) > writes_before);
+}
+
+// The system root a post-commit epilogue leaves behind is never written to disk on its own, so
+// the durable commit that supersedes it updates those pages rather than copying them. Copying
+// makes a small commit write the same metadata twice.
+#[test]
+fn durable_overwrite_does_not_write_post_commit_metadata() {
+    const COMMITS: u64 = 20;
+
+    let writes = Arc::new(AtomicU64::new(0));
+    let backend = WriteCountingBackend {
+        inner: InMemoryBackend::new(),
+        writes: writes.clone(),
+    };
+    let db = Database::builder().create_with_backend(backend).unwrap();
+
+    let txn = db.begin_write().unwrap();
+    {
+        let mut table = txn.open_table(U64_TABLE).unwrap();
+        table.insert(&0, &0).unwrap();
+    }
+    txn.commit().unwrap();
+
+    let writes_before = writes.load(Ordering::SeqCst);
+    for value in 1..=COMMITS {
+        let txn = db.begin_write().unwrap();
+        {
+            let mut table = txn.open_table(U64_TABLE).unwrap();
+            table.insert(&0, &value).unwrap();
+        }
+        txn.commit().unwrap();
+    }
+
+    // A steady-state single-key overwrite writes the pages it dirties and the header, nothing
+    // else. The bound leaves room above that, so it survives incidental changes to the commit
+    // path while still catching a commit that writes another root's worth of metadata.
+    let commit_writes = writes.load(Ordering::SeqCst) - writes_before;
+    assert!(
+        commit_writes <= 7 * COMMITS,
+        "{COMMITS} durable overwrites issued {commit_writes} backend writes"
+    );
 }
 
 fn test_persistence(durability: Durability) {


src/tree_store/page_store/page_manager.rs
```diff
@@ -1239,6 +1271,50 @@ impl TransactionalMemory {
...
+    // Frees pages that are known to predate the current write transaction. Keeping the allocator
+    // lock across the batch avoids lock traffic when a large transaction makes millions of pages
+    // eligible at once. Unlike free_helper() the lock is released before the pages are dropped
+    // from the caches, which is safe only because the single live write transaction is the only
+    // thing that allocates: nothing can claim a page from this batch and write to it in between.
+    pub(crate) fn free_committed_batch(&self, pages: &[PageNumber]) {
+        #[cfg(debug_assertions)]
+        {
+            let read_pages = self.read_page_ref_counts.lock().unwrap();
+            let dirty_pages = self.open_dirty_pages.lock().unwrap();
```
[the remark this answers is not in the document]
cberner: Good catch, this was real. `get_page()` takes `read_page_ref_counts` while holding `open_dirty_pages`, and this block held them in the other order, so a reader and a committing writer could deadlock in a debug build. `free_helper()`, which this replaced, never held two of these at once -- each `.lock()` there is a temporary dropped at the end of its statement -- so the inversion came in with the batch.

Fixed in `53e2df7` by checking one page and one guard at a time, exactly as `free_helper()` does. That costs N lock acquisitions instead of 3, but only under `debug_assertions`; the lock this commit is actually trying to batch is `state`, which is untouched.

---

cberner:
Fuzzing result, as promised in the description.

`cargo fuzz run --sanitizer=none fuzz_redb -- -max_len=10000` against `05108d9`: **310,556 runs**, no crashes, nothing written to `fuzz/artifacts`. Final coverage 11,795 with a 6,207-input corpus.

Earlier campaigns on this branch's predecessors, for context: 531,570 runs against the original version of the change and 1,176,032 against the larger one that also held the epilogue's frees in memory, both clean. The corpus carried across all of them, so this run started from the coverage those built rather than from scratch.

---

cberner (commented):
The comments in the code read too much like an explanation of this diff / PR. Instead rewrite them to be about the invariants and long term architecture of the code base

cberner:
Fair, and it is what AGENTS.md asks for. Rewritten in `fef3445` -- comments only, the code is byte-identical to `05108d9`.

The ones that were narrating the diff, and what they say now:

* **`free_committed_batch` (both)** dropped "Keeping the allocator lock across the batch avoids lock traffic when a large transaction makes millions of pages eligible at once" and "Unlike `free_helper()`...". They now state the locking discipline itself: the allocator lock is taken once for the batch, the cache eviction afterwards runs without it, and that is sound only because the single live write transaction is the only thing that allocates.
* **The debug block** no longer says "exactly as `free_helper()` does". It states the rule instead: `get_page()` takes `read_page_ref_counts` while holding `open_dirty_pages`, so code needing both follows that order or holds one at a time. That is the fact worth knowing next time someone touches either lock.
* **`claim()`** had four lines explaining why the fix exists. It is now one line pointing at the invariant, which lives on the `post_commit_allocations` field where the set is declared: a subset of `pages`, so a reused page number is never mistaken for the allocation that held it before.
* **`adopt_unpersisted`** and its call site were framed as "adopting them here lets ... rather than copying". They now give the precondition -- a durable commit is irreversible from that point, so nothing rolls back what it claims -- which is the part that decides whether a future caller may use it.
* **The test bound** cited 6-versus-7.8 write counts, which only mean anything if you know what this PR changed. It now describes the steady state a single-key overwrite should reach.

---

cberner:
Correcting the record: this PR has a space regression, and an earlier version of the description claimed the opposite. Not ready to merge.

Live file after 4,000 single-key durable commits, measured with the database open:

| keys touched | master | this PR |
|---|---|---|
| 1 | 90,112 | **557,056** |
| 4,096 | 135,168 | **557,056** |
| 65,536 | 135,168 | **557,056** |

Nothing is leaked -- `allocated_pages()` is 11 on both, `check_integrity()` is clean, `compact()` reaches 16,384 on both. The difference is the file's floor.

`try_shrink` only trims *trailing* free pages, so the floor is the highest allocated page. Copy-on-write reallocates at the lowest free index every time, so the working set migrates downward and the file self-compacts as a side effect of ordinary operation. Updating pages in place means they never migrate, and a page that landed high freezes the floor there permanently.

What I got wrong: I originally measured file size *after* a clean close, and that number is phase-dependent rather than a property of the change. On master, 4 warmup commits plus 1,000 gives 90,112, while 3 plus 1,000 gives 53,248 -- same code, same workload. I read one sample of that as evidence of an improvement. The live high-water above is stable across commit counts and key counts, and is the metric that should have been in the description from the start.

To be clear about what is *not* affected: `f9a348d`'s fix still holds. The epilogue still runs and still processes `DATA_FREED_TABLE`, so data pages freed by a commit remain available to the next one. Only the epilogue's own allocations are treated differently here.

The plausible fix is to bound adoption to pages below the current trailing-free boundary, so an adopted page can never be the one blocking a trim -- keeping most of the write win without defeating self-compaction. I have not implemented or measured that yet, and I would rather not land the write win at the cost of a 4-6x file floor until it is settled.

---

cberner:
Rescoped against what this is actually for: #829 and commit speed. Force-pushed to `d3e2394`, now a single 36-line commit, and the "do not merge" note is withdrawn.

**The space concern I raised is retracted.** I measured the #829 reproducer directly -- repeated full overwrite of 200 MiB of live data -- and this branch is identical to master at 2.57x overhead, against 5.14x before `f9a348d`. The fix is not at risk. The floor difference I flagged is a constant ~460 KB that only shows up on databases of a few pages and disappears entirely at any real size; it was the wrong thing to block on.

**Dropped the batching commit.** I had justified it as mattering when "a transaction makes millions of pages eligible at once", so I measured exactly that:

| pages deleted | master | batched |
|---|---|---|
| 419k | 1052, 1062 ms | 1078, 1107 ms |
| 838k | 4175, 4175, 4355 ms | 4192, 4014, 3993 ms |
| 1.7M | 8140 ms | 8445 ms |

Between 4% faster and 4% slower -- no reliable effect at any scale. The delete is dominated by walking the tree and writing pages, not by the allocator lock. It also added API surface on two layers, added peak memory proportional to the reclaim, and was where the lock-ordering deadlock came from. Not worth landing unproven, so it is gone.

What remains is the part that has a measured effect on the goal: 8.0 to 6.0 backend writes per small durable commit, neutral on large transactions, `fdatasync` unchanged.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -191,6 +191,18 @@ impl PageAllocator {
...
     }
 
+    // Takes ownership of pages allocated by an earlier transaction so this one can update them
+    // in place. Only sound once this transaction can no longer abort: rollback_all() frees
+    // whatever is adopted here. A page that has stopped being unpersisted was reclaimed and its
+    // number may now belong to something else, so only pages still in that set are taken.
+    pub(crate) fn adopt_unpersisted(&self, pages: impl IntoIterator<Item = PageNumber>) {
+        for page in pages {
+            if self.mem.unpersisted(page) {
+                self.allocated_since_commit.insert(page);
```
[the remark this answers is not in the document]
cberner: Correct mechanism, and independently arrived at -- this matches what I measured before you posted, and it is written up under "Not covered". It is a deliberate trade rather than an oversight.

The size of it: the floor is ~460 KB higher than master on a database of a few pages, and it is a constant, not proportional to anything. Live file after 4,000 single-key durable commits is 90,112 on master against 557,056 here, and that same 557,056 holds whether the workload touches 1 key or 65,536. At any real size it disappears -- with 200 MiB of live data both master and this branch sit at 514 MiB. Nothing leaks: `allocated_pages()` is 11 on both, `check_integrity()` is clean, and `compact()` reaches 16,384 on both.

The reason it is accepted is that this PR exists for #829 (large transactions doubling disk usage) and for commit throughput. Against those: the #829 reproducer is unchanged at 2.57x overhead on both branches, against 5.14x before `f9a348d`; small durable commits drop from 8.0 to 6.0 backend writes; large transactions are neutral. Maximally trimming a small file is explicitly not a goal here, so paying ~460 KB on tiny databases to remove two writes from every small commit is the intended direction.

The suggested fix is the right shape if this ever does start to matter -- decline adoption for any page at or above the trailing-free boundary, so an adopted page can never be the one blocking a truncation. That keeps the write win for the common case where the epilogue's pages are nowhere near the end of the file. I have not implemented it, because a bound that never fires on any workload measured so far is speculative complexity in the commit path, which is what this PR spent its last two revisions removing.

---

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 * `Durability::None` commits are about 2x faster.
 * Writes that do not split a page are faster: about 15% for single-key `Durability::None` commits,
   and about 6% for batched writes.
+* Reduce the metadata writes a small durable commit issues, by about a quarter.
```
cberner: Hmm, seems like this doesn't apply since we're fixing a regression since the last release

cberner: Agreed, removed in `e587aae`. The PR is now source and tests only.

The entry was measuring against an unreleased intermediate state. Nobody upgrading from 4.1.0 ever saw the 8.0 writes/commit that `f9a348d` introduced, so "reduce by about a quarter" describes a round trip inside the 4.2.0 development window rather than anything user-facing.

One thing worth deciding separately, since removing my entry makes it easier to miss: 4.2.0 still ships at **6.0 writes per small durable commit against 4.1.0's 5.0**, even with this PR. That extra write is the standing cost of the `f9a348d`/#829 fix -- the epilogue's `SYSTEM_FREED_TABLE` root has to be written once per commit -- and it is not a regression this PR leaves behind so much as the price of the feature. The existing `f9a348d` entry describes the benefit ("Reuse pages freed by a durable write transaction in the next write transaction...") without mentioning that cost. If you want that documented it belongs on that entry rather than on a new one; if you would rather not put a per-commit write count in front of users at all, then no changelog change is needed and this is settled.

---

cberner:
Force-pushed to `1ccf6ad`. One line of behaviour change: `adopt_unpersisted` now claims each page out of the unpersisted set instead of only testing that it is still in it.

This closes a hole the branch introduced. `adopt_unpersisted` left the page in `unpersisted.pages`, but if the commit then copies it out -- a split or a merge in the system tree -- `free_if_uncommitted` reaches `TransactionalMemory::free`, and `free_helper` does not maintain that set. The page number is free and reusable within the same commit while still marked unpersisted, which is exactly what the "reclaiming a page drops it from both" invariant on `post_commit_allocations` says cannot happen.

It is reachable. With `assert!(!self.mem.unpersisted(page))` added to `free_if_uncommitted`:

| | result |
|---|---|
| master | 94,067 fuzz runs, clean |
| `e587aae` | panics within seconds |
| `1ccf6ad` | 102,292 fuzz runs, clean |

I could not construct a way to make it *observable* on `e587aae`. The window runs from the copy to `mem.commit()`'s `unpersisted.clear()`, the durable path passes `None` to `store_system_freed_pages` so nothing checks `unpersisted` in between, and `process_freed_pages` runs before the window opens. So this is a latent invariant violation rather than a bug you can trigger -- but it held only by an ordering accident, and the fix makes it hold by construction. This also replaces the "not covered by anything I could construct" note about that guard in the description: the guard now has a mechanism behind it and a fuzz signal that distinguishes it.

Everything else is unchanged: `cargo test --all-features` passes, clippy and fmt are clean, and the 2,000-commit `strace` count is still 12,035 `pwrite64` / 2,014 `fdatasync`.

<details>
<summary>How this turned up: a prototype that generalized adoption, which I am not proposing</summary>

I built the version where the system tree gets its own `PageAllocator`, the unpersisted-system-page set is maintained across all non-durable commits rather than just the epilogue, and the commit fence runs for both durabilities -- which deletes the `extract_if`/`post_commit_frees`/`free_if_unpersisted` reclaim dance in `non_durable_commit` and the `Option<&mut Vec<PageNumber>>` threaded through `store_system_freed_pages`.

It works and it is slower. 100k commits, us/commit, interleaved:

| workload | master | this PR | prototype |
|---|---|---|---|
| pure `Durability::None` | 7.7-8.1 | 8.5 | 9.6-9.7 |
| durable every 100 | 11.7-12.7 | 11.3-12.0 | 12.9-14.1 |
| durable every 10 | 41.4-47.7 | 41.5-41.9 | 41.0-42.2 (fsync-bound) |

Durable write count, `#829` overhead and the file floor are all identical to this PR.

The reason is that allocations per non-durable commit are 2.00 on all three, and both are data-tree pages: under `Durability::None` the system tree is already never written, because freed-page records are held in memory precisely so a non-durable commit does not mutate it. There is no copy for the generalized fence to avoid. Keeping every other part of the prototype but gating the fence back to `Immediate` measures 8.0 us, so the whole regression is the fence running on non-durable commits.

Which is the answer to whether the epilogue special case is an accident of the design: it is not. The epilogue is the only non-durable commit that routinely writes the system tree.

</details>

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -191,6 +191,23 @@ impl PageAllocator {
...
+    // whatever is adopted here.
+    //
+    // An adopted page leaves the unpersisted set, because from here it is an ordinary uncommitted
+    // page of this transaction: a copy-on-write may free it, and freeing does not maintain that
+    // set, so an entry left behind would make a later reuse of the page number look reclaimable.
+    // A page that has already stopped being unpersisted was reclaimed and its number may now
+    // belong to something else, so only pages still in the set are taken.
+    pub(crate) fn adopt_unpersisted(&self, pages: impl IntoIterator<Item = PageNumber>) {
+        for page in pages {
+            if self.mem.claim_unpersisted(page) {
```
cberner: Should this assert instead of an if? Is it expected that it could return false?

cberner (commented):
Tighten the comments up a bit

src/tree_store/page_store/page_manager.rs
cberner: You're right, it can't return false. Changed to an assert in `d08772a`.

`unpersisted.pages` only ever shrinks in two places, `UnpersistedState::clear()` and `UnpersistedState::claim()`, and both drop the page from `post_commit_allocations` at the same time. So every page reaching here is still in `pages` by construction, and the `if` was hedging against a state the subset invariant already rules out -- it would have silently declined to adopt rather than reporting that the invariant broke.

That also took most of the comment with it: the paragraph explaining when a page might have "stopped being unpersisted" was describing the branch, not the code.

Tightened the rest of the comments in the same commit. Full suite passes in debug and release -- worth checking here specifically, since `assert!` is compiled in for release and would fire there.

---

src/transactions.rs
```diff
@@ -1672,6 +1672,14 @@ impl WriteTransaction {
         let (user_root, allocated_pages, data_freed) =
             self.tables.lock().unwrap().table_tree.flush_and_close()?;
 
+        // From here a durable commit is irreversible: it either completes or leaves the allocator
+        // state invalidated, so nothing rolls back what it claims. That is what lets it take over
+        // the epilogue's pages, which no durable root names and it is about to supersede.
+        if self.durability == InternalDurability::Immediate {
+            let pages = self.mem.take_post_commit_allocations();
+            self.page_allocator().adopt_unpersisted(pages);
```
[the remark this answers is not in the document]
cberner: Half of this was already decided, and the other half is a fair catch but not mine to make.

**The write-count half does not apply.** There was a CHANGELOG entry for it; @cberner asked for it to be removed ([here](https://github.com/cberner/redb/pull/1378#discussion_r3788066911)) and it went in `e587aae`. `f9a348d`, which introduced the extra writes, has never shipped -- it is not in v4.1.0 or v4.0.0. So "25% fewer backend writes" measures a round trip inside the 4.2.0 development window against an intermediate state no user has run. AGENTS.md asks for the *user-facing impact*, and against 4.1.0 the impact of this PR alone on write counts is nothing.

**The file-size half is a real point**, and narrower than you have it. Against 4.1.0, most of what a user would notice in 4.2.0 comes from `f9a348d`: large-transaction overhead 5.14x to 2.57x, and `Durability::None` commits ~21.9 to ~9.1 us. The higher small-database floor is the one item this PR introduces on its own -- 90,112 bytes to 307,200 after 4,000 single-key durable commits, constant rather than proportional, and gone at any real size (both sit at 514 MiB with 200 MiB live).

Two corrections to the premise while I am here. Small durable commits are not "faster" in the sense that matters to a user: `fdatasync` is unchanged at 2,014 either way, so latency is the same and what improves is write amplification. And non-durable commits being slower is relative to master, not to any release -- against 4.1.0 they are still ~2.3x faster.

Whether the floor belongs in the changelog is @cberner's call, not something to re-add unilaterally to a PR whose changelog entry he just removed. It is flagged alongside the other open question there: whether the standing +1 write per small durable commit that 4.2.0 ships (6.0 against 4.1.0's 5.0, the price of the #829 fix) is worth noting on `f9a348d`'s existing entry. Both are consequences of `f9a348d` rather than of this diff, so if either is documented it likely belongs on that entry.

---

PR #1382, opened 2026-08-17
Store minimal key separators for string keys

cberner:
Follow-up to #1358, which left `&str` and `String` on the default `Key::separator()`.

The reason given there was that their `compare()` deserializes, so a truncated separator could be invalid UTF-8. That rules out truncating *anywhere*, not truncating at all. `str` orders as its bytes, so the byte slice rule separates two strings correctly; only the cut has to land on a character boundary.

```rust
let mut separator_len = <common prefix> + 1;
// Step over the continuation bytes that follow, keeping whole the character the cut
// landed inside
while separator_len < right.len() && right[separator_len] & 0b1100_0000 == 0b1000_0000 {
    separator_len += 1;
}
```

Recognising a continuation byte by its leading `10` bits costs at most three byte tests. `str::is_char_boundary()` would have been clearer, but it needs `from_utf8()`, which validates the whole key -- on every leaf split. `String` delegates to `&str`, being encoded identically.

## Effect

Over 2M random pairs drawn from an alphabet of one, two, three and four byte characters (plus NUL and U+10FFFF), this shortens **79% of separators, by 10.2 bytes on average**. Keys that share long prefixes -- paths, URLs, namespaced identifiers -- are where string keys tend to live, and are exactly the case it helps.

## Which other built-in types could take one

For the record, having gone through all of them. The remaining variable width types are blocked by one shared constraint: `separator()` returns `&'a [u8]`, so it can only return a *subslice* of an input, and cannot rewrite an embedded length or offset table.

| type | verdict |
|---|---|
| `&[u8]` | done in #1358 |
| **`&str`, `String`** | **this PR** |
| `Option<T>`, T variable | possible -- the tag byte is at the front, so `[1] ++ T's separator` is a valid prefix when T shortened. Narrow: needs a byte comparison to detect whether it did |
| tuple, differing in the **last** element | possible -- varint lengths cover every element *except* the last, whose length is implicit, so a cut inside it stays structurally valid |
| tuple, differing in any earlier element | not possible -- would have to rewrite the varint length table |
| `[T; N]`, T variable | not possible -- end offsets for all N elements sit at the front, and a truncation leaves them pointing past the buffer |
| `#[derive(Key)]` structs | not possible -- the generated `compare()` runs `from_bytes()` on both sides and `Ord::cmp`, so a separator must round trip as a real value |
| `()`, `bool`, `char`, all integers, `&[u8; N]`, `Uuid`, all five chrono types, and the internal `SavepointId` / `TransactionIdWithPagination` / `AllocatorStateKey` | fixed width -- `branch_separator()` returns before `separator()` is reached |

`Vec<T>` never arises: it implements `Value` but not `Key`.

## Testing

* `cargo test --all-features`: 442 tests pass. `cargo clippy --all-targets --all-features` clean under `--deny warnings`; `cargo fmt --check` clean; both `no_std` / `thumbv7em-none-eabihf` feature combinations compile clean.
* A `&str` keyed variant of the randomized differential test from #1358, against a `BTreeMap` reference, with keys built from an alphabet spanning one, two, three and four byte characters so a naive byte cut would split one: 20 broad seeds, plus 2 grown to ~180K entries and height 4. The `&[u8]` campaign was re-run unchanged as a regression check.
* Deleting the character boundary loop makes that fuzzer panic immediately, so it is exercising mid-character truncation rather than passing vacuously.

New unit tests cover the boundary cases directly: a cut inside a two byte and a four byte character, a case where keeping the whole character would reach `right` itself, and the empty key. They also assert every separator is valid UTF-8, and that `String` agrees with `&str`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HHd4GfPNP7dELGJtck5JWn)_

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -613,11 +613,32 @@ impl Value for &str {
     }
 }
 
+// `index` advanced to the next character boundary, or the end. Only a character's trailing bytes
+// match `10xx_xxxx`, so this needs no deserialization, unlike `str::is_char_boundary()`.
+fn round_up_to_char_boundary(utf8: &[u8], mut index: usize) -> usize {
+    while index < utf8.len() && utf8[index] & 0b1100_0000 == 0b1000_0000 {
+        index += 1;
+    }
+    index
+}
+
 impl Key for &str {
     fn compare(data1: &[u8], data2: &[u8]) -> Ordering {
         let str1 = Self::from_bytes(data1);
         let str2 = Self::from_bytes(data2);
         str1.cmp(str2)
+    }
+
+    fn separator<'a>(left: &'a [u8], right: &'a [u8]) -> &'a [u8] {
+        debug_assert!(left < right);
+        // `str` orders as its bytes, but `compare()` deserializes, so the cut must keep it valid
+        let common_bytes = left.iter().zip(right).take_while(|(x, y)| x == y).count();
+        let separator_len = round_up_to_char_boundary(right, common_bytes + 1);
+        if separator_len < left.len() && separator_len < right.len() {
+            &right[..separator_len]
+        } else {
+            left
+        }
     }
 }
 
@@ -659,6 +680,11 @@ impl Key for String {
         let str1 = core::str::from_utf8(data1).unwrap();
         let str2 = core::str::from_utf8(data2).unwrap();
         str1.cmp(str2)
+    }
+
+    // Encoded the same way as `&str`, so it separates the same way
+    fn separator<'a>(left: &'a [u8], right: &'a [u8]) -> &'a [u8] {
+        <&str as Key>::separator(left, right)
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -613,12 +613,37 @@ impl Value for &str {
     }
 }
 
+// `index` advanced to the next character boundary of `utf8`, or its length. Equivalent to
+// scanning for the next index where `str::is_char_boundary()` holds, but without deserializing
+// the slice: only the trailing bytes of a multi-byte character match `10xx_xxxx`, so those are
+// exactly the bytes that are not a boundary. Validating instead would cost a pass over the whole
+// key, on a path that runs for every btree page that splits.
```
cberner: too verbose

cberner: Cut to two lines in 98c22fe:

```rust
// `index` advanced to the next character boundary, or the end. Only a character's trailing bytes
// match `10xx_xxxx`, so this needs no deserialization, unlike `str::is_char_boundary()`.
```

and the one inside `separator()` to one line. Went through the rest of what this PR adds for the same habit and collapsed four more two-liners in the tests.

---

PR #1385, opened 2026-08-18
Reserve the repair commit's transaction id in check_integrity()

cberner:
Fixes P0 finding 4 from the pre-5.0 bug audit: `check_integrity()`'s non-promote repair path commits with `last_committed + 1` without reserving that id in the `TransactionTracker`. The tracker's counter keeps issuing from where it left off, and after any write transaction it equals the last committed id — so the next `begin_write()` commits the very id the repair commit just used, leaving both on-disk commit slots holding equal transaction ids with different content.

1PC+C crash recovery resolves the "only the god byte reached disk" torn-commit case by `secondary.transaction_id > primary.transaction_id` (`docs/design.md` requires monotonically increasing ids). With a duplicated id that comparison keeps the wrong slot, and a user's earlier fsync-acknowledged commit silently disappears.

**The fix**
- `check_integrity_inner()` reserves the repair commit's id in the tracker after committing, via a new `TransactionTracker::reserve_repair_transaction_id()` — mirroring what `Database::new()` (which re-seeds the tracker after its repair commit) and the post-commit epilogue (`reserve_transaction_id()`) already do for their out-of-band commits.
- `TransactionalMemory::commit()` now asserts every commit is strictly newer than the primary slot it supersedes, so any remaining way to duplicate an id fails loudly instead of corrupting recovery. Only the primary is compared: the secondary is overwritten by the commit and may legitimately hold a torn slot's garbage before a repair.

**Test**
`repair_commit_reserves_transaction_id` reproduces the collision end to end: it leaks pages via a transaction dropped during a caught panic (whose clean close persists the leak), reopens, re-aligns the tracker with a normal commit (post-commit epilogue disabled, so no pending non-durable commit routes `check_integrity()` into the tracked promote path), runs `check_integrity()`, and asserts the next write transaction's id is newer than the repair commit's. Without the fix the ids are equal and the test fails. The test is gated on `cfg(panic = "unwind")` since its leak setup needs `catch_unwind`, which aborts the whole test process on the wasi target (this was the initial CI failure on this PR).

**Validation**: the constituent commands of `just test`, run directly because this environment lacks the rootless podman its sandbox wrapper needs — `cargo deny --workspace --all-features check licenses advisories` ("advisories ok, licenses ok"), `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features` with `--deny warnings`, and `RUST_BACKTRACE=1 cargo test --all-features` (all suites green).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@
   nodes store the result instead of a whole key, so more children fit in each node and lookups
   touch fewer pages. The default implementation returns a whole key, leaving existing `Key`
   implementations unchanged; `&[u8]`, `&str`, and `String` keys now store minimal prefixes.
+* Fix a crash shortly after a commit being able to silently roll that commit back during
+  recovery, if `check_integrity()` had previously repaired the database.
 
 ## 4.2.0 - 2026-08-17
 
--- a/src/db.rs
+++ b/src/db.rs
@@ -731,6 +731,10 @@ impl Database {
                 true,
                 ShrinkPolicy::Never,
             )?;
+            // Reserve the id, or the next write transaction would commit with the same one,
+            // which crash recovery could then resolve to the wrong slot
+            self.transaction_tracker
+                .reserve_repair_transaction_id(next_transaction_id);
         }
 
         self.mem.begin_writable()?;
--- a/src/transaction_tracker.rs
+++ b/src/transaction_tracker.rs
@@ -244,6 +244,14 @@ impl TransactionTracker {
         state.next_transaction_id = id;
     }
 
+    // Reserves a repair commit's id so it is never issued again: crash recovery orders the
+    // commit slots by transaction id, so an id must never be committed twice
+    pub(crate) fn reserve_repair_transaction_id(&self, id: TransactionId) {
+        let mut state = self.state.lock().unwrap();
+        assert!(state.live_write_transaction.is_none());
+        state.next_transaction_id = state.next_transaction_id.max(id);
+    }
+
     pub(crate) fn restore_savepoint_counter_state(&self, next_savepoint: SavepointId) {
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,10 @@
   nodes store the result instead of a whole key, so more children fit in each node and lookups
   touch fewer pages. The default implementation returns a whole key, leaving existing `Key`
   implementations unchanged; `&[u8]`, `&str`, and `String` keys now store minimal prefixes.
+* Fix `Database::check_integrity()` giving its repair commit a transaction id that the next
```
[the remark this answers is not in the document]
cberner: This doesn't match the branch: `claude/fix-repair-commit-id-reuse` contains exactly one commit (`6485a733`), authored and committed by `Christopher Berner <me@cberner.com>`, and its message ends with an `Assisted-by: Claude Code` trailer. No commit `f7ce8aac` exists on this branch — `git cat-file -t f7ce8aac2991...` reports no such object in the repository.

---

CHANGELOG.md
```diff
@@ -49,6 +49,10 @@
...
+  write transaction would commit again. Crash recovery orders the commit slots by transaction
+  id, so a duplicated id could make recovery after a torn commit keep the repair commit and
+  silently roll back a later fsync-acknowledged commit.
```
cberner: Word this around the user facing impact. No implementation details.

cberner (commented):
The comments you added in the code are too verbose too

CHANGELOG.md
cberner: Reworded around the user-facing impact: "Fix a crash shortly after a commit being able to silently roll that commit back during recovery, if `check_integrity()` had previously repaired the database." Also trimmed the code comments down to one or two lines each per the review note.

---

PR #1393, opened 2026-08-19
Avoid rewriting pages on logical no-op writes

cberner:
Fixes P2 findings 12 and 15 from the pre-5.0 bug audit — both are the same class of defect: a mutating operation that turns out to be a logical no-op still rewrote pages.

**`Table::get_mut()` of a missing key (15).** The mutating descent copy-on-wrote every page from the root down before the leaf revealed the key was absent. Per review, this is now a single-search two-pass descent instead of a double lookup: one read-only descent searches for the key while recording the child index at each branch; only after the leaf proves the key exists does a second pass copy-on-write along the recorded route — fetching the just-touched pages by number, with no further key comparisons. A miss is now a pure read. On a hit, the only added work over the old combined descent is re-fetching the hot pages from the cache by page number; every binary search happens exactly once.

**`MultimapTable::insert()` of an already-present value (12).** The subtree-backed arm copy-on-wrote the subtree and re-inserted the collection into the parent tree even when the value existed. It now checks the subtree for the value first and returns `true` untouched — the same shape as the merged `remove()` fix, whose early return this mirrors. The check adds one subtree-only read descent to the miss path, ahead of the full two-tree insert that path performs anyway.

**Tests** — both mirror the existing `multimap_remove_nonexistent_value_from_subtree_does_not_churn` pattern (steady-state baseline, the no-op, then `allocated_pages` must be unchanged), and both were verified to fail without their fix with the churn assertion:
- `multimap_insert_existing_value_into_subtree_does_not_churn`
- `get_mut_missing_key_does_not_churn`

**Validation**: `cargo fmt --check`, clippy with `--deny warnings`, and `RUST_BACKTRACE=1 cargo test --all-features` (all 20 suites green, including the `get_mut`, entry-API, and guard-resize tests that exercise the rewritten path), run directly because this environment lacks the rootless podman that `just test`'s sandbox wrapper needs.

### DIFF
--- a/src/tree_store/btree.rs
+++ b/src/tree_store/btree.rs
@@ -11,7 +11,7 @@ use crate::tree_store::btree_cursor::BtreeCursorMut;
 use crate::tree_store::btree_cursor::{CursorMut, Position};
 use crate::tree_store::btree_iters::{bounds_are_empty, encode_bounds};
 use crate::tree_store::btree_mutator::MutateHelper;
-use crate::tree_store::page_store::{Page, PageImpl, PageMut};
+use crate::tree_store::page_store::{Page, PageImpl};
 use crate::tree_store::{
     AccessGuardMutInPlace, AllPageNumbersBtreeIter, BtreeCursorRange, BtreeExtractIf,
     PageAllocator, PageHint, PageNumber, PageNumberHashMap, PageResolver, PageTracker,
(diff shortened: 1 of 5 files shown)


src/tree_store/btree.rs
```diff
@@ -593,6 +593,11 @@ impl<K: Key + 'static, V: Value + 'static> BtreeMut<K, V> {
         &mut self,
         key: &K::SelfType<'_>,
     ) -> Result<Option<AccessGuardMut<'_, V>>> {
+        // A miss must not modify the tree: the mutating descent below copy-on-writes every
+        // page on its path before the leaf reveals whether the key exists
+        if self.get(key)?.is_none() {
+            return Ok(None);
+        }
```
cberner: This looks like it's going to slow things down by doing 2x more work on the common path

cberner (commented):
It's not acceptable to do 2x lookups

src/tree_store/btree.rs
cberner: Reworked in 389c644: the pre-check is gone. `get_mut()` now does one read-only descent that searches and records the child index per branch, and only after the leaf proves the key exists does it copy-on-write along the recorded route — pages fetched by number, no second key search. A miss stays a pure read; on a hit the only work added over the old combined descent is re-fetching the just-touched pages from the cache.

---

tests/basic_tests.rs
```diff
@@ -1686,6 +1686,94 @@ fn insert_reserve() {
     );
 }
 
+#[test]
+fn get_mut_missing_key_does_not_churn() {
+    // get_mut() on a missing key is a logical no-op and must not rewrite any pages. There was
+    // a bug where the mutating descent copy-on-wrote the whole root-to-leaf path before the
+    // leaf revealed that the key was absent.
```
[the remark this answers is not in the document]
cberner: Trimmed in a15e015: both test comments now state only the invariant — "a logical no-op and must not rewrite any pages" — with the bug history left to the commit message.

---

cberner:
I don't think this is worth it. Too complicated for too small a gain

PR #1401, opened 2026-08-20
Shorten separators for array keys

cberner:
`[T; N]` is variable width exactly when `T` is, so it is one of the built-in key types whose branch keys `separator()` can shorten.

Only the element the two keys first differ in can be shortened: the elements before it are needed for the result to sort below the right key. The elements after it are replaced by `min_encoded_key()`, but only once the shortened element sorts **strictly** above the left key's — that is what makes both comparisons resolve at that element and never read what follows. When the shortened element comes back equal to the left key's, the comparison against the left key runs on into the tail, so the tail has to stay.

That strictness gate is what the earlier version of this PR was missing, and why it was closed. Discarding the tail unconditionally corrupted lookups on nested composites. #1406 supplies the other half: an element is discarded to its own smallest *encoding*, not to nothing, so `compare()` can still read it. `[Option<&str>; 2]` is the case that breaks otherwise — `Option::compare()` reads a tag byte, so an empty element leaves it indexing past the end.

The encoding heads each array with one end offset per element, so an assembled separator rewrites every offset. That synthesized result is what returning a `Cow` allows. An array of a fixed width element type is fixed width itself, and its encodings are compared at that width, so it keeps returning the whole key.

Nothing guarantees a smallest encoding is shorter than the element it replaces, so the assembled length is computed and compared against the whole key before anything is allocated. That also bounds every end offset below `left.len()`, which is itself a `u32` in a valid encoding, so writing the header cannot overflow. Stack use is constant in `N`: offsets are read from the header as needed rather than collected.

## Testing

- `array_separator` pins the exact separator bytes for `[&str; 2]` — a shortened leading element with the tail discarded, a shortened last element, and the case where nothing sorts between the two first elements so the tail stays — and checks the `[left, right)` bounds.
- `discarded_array_elements_stay_readable_by_compare` covers `[Option<&str>; 2]`: the discarded element comes out as `None`, not empty, and a lookup key that matches the separator's first element still compares correctly against it.
- `separator_is_not_grown_by_a_wide_minimum` uses a conforming `WideMinimum` key type whose smallest value encodes to 64 bytes. Without the length check, `[WideMinimum; 3]` turned a 17 byte key into a 141 byte separator.
- `fixed_width_array_separator_returns_full_key` covers `[u64; 2]`.
- `array_keys_with_shortened_separators` builds a multi-level tree of 2,000 `[&str; 2]` keys whose separators discard the trailing element, checks every lookup and the iteration order, then removes three quarters of them to exercise branch merges and rebalancing, and re-checks. The btree's own debug assertions validate every separator it stores along the way.
- `array_separators_shrink_branch_nodes` stores the same 5,000 keys twice, as `[&str; 2]` and as `[[&str; 1]; 2]` — identical bytes, differing only in whether the trailing element has a smallest encoding to discard to. 5 branch pages against 19.
- `separators_are_encodings` (already on master, from #1405) starts exercising `[&str; 2]`, `[&str; 3]` and `[Option<&str>; 2]` with this change.

`cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test --all-features` all pass, as does the `thumbv7em-none-eabihf` no_std check. `just`/podman are unavailable in this environment, so those ran directly rather than through `just test`; no dependencies changed, so `cargo deny check licenses` was not run.

Beyond the committed tests, I checked the contract over several thousand random ordered pairs per type (including nested cases such as `[Option<&str>; 2]`), and ran a model based btree test doing random inserts and deletes up to ~20k `[&str; 3]` entries at height 3, matching a `BTreeMap` exactly. That harness was scratch work and is not part of this PR.

Codecov flags two spots, both deliberate: the `Cow::Borrowed(left)` after the element loop, reached only when every element compares equal, which the method's precondition rules out; and `WideMinimum::type_name()`, which the test type has to define but nothing calls.

## Notes

#1406, which adds `min_encoded_key()`, has merged, so this targets master directly. #1402 does the same for tuples, and conflicts with this one only on the `CHANGELOG.md` entry. Both carry their own copy of the `WideMinimum` test type, since neither depends on the other.

Arrays could now supply a `min_encoded_key()` of their own — concatenating their elements' — which the associated const this started as could not express. Left out here to keep this to one change; it would let an array nested inside another composite be discarded too.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AcCEQ8N51j9NsV1dLHDnPL)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@
   `Option` keys shorten their payload, when the wrapped type is variable width.
 * Add `Key::min_encoded_key()`, the encoding of a key type's smallest value. Implementing it is
   optional, and lets container types holding that key store shorter separators.
+* Shorten separators for array keys, when the element type is variable width. Tables with such
+  keys use slightly less space, and lookups are faster.
 * Fix a crash shortly after a commit being able to silently roll that commit back during
   recovery, if `check_integrity()` had previously repaired the database.
 * Fix iterators silently omitting data when iteration continues after an error. An iterator
--- a/src/types.rs
+++ b/src/types.rs
@@ -626,6 +626,78 @@ impl<const N: usize, T: Key> Key for [T; N] {
         }
         Ordering::Equal
     }
+
+    fn separator<'a>(left: &'a [u8], right: &'a [u8]) -> Cow<'a, [u8]> {
+        // A fixed width element type makes the array fixed width too, and its encodings are
+        // compared at that width, so nothing shorter than `left` may be returned
+        if T::fixed_width().is_some() {
+            return Cow::Borrowed(left);
+        }
+        // Only the element the two first differ in can be shortened: the elements before it are
+        // needed for the result to sort below `right`. The elements after it can be replaced by
+        // the smallest encoding, but only once the shortened element sorts strictly above
+        // `left`'s, so that both comparisons resolve there and never read what follows.
+        let header = size_of::<u32>() * N;
+
+        for i in 0..N {
+            let left_element = array_element::<N>(left, i);
+            let right_element = array_element::<N>(right, i);
+            if T::compare(left_element, right_element).is_eq() {
+                continue;
+            }
+            let separator = T::separator(left_element, right_element);
+            // There is a tail to replace only when this is not the last element, and it may only
+            // be replaced once this element sorts strictly above `left`'s. `min_encoded_key()`
+            // may allocate, so it is asked only when the answer will be used.
+            let replaces_tail = i + 1 < N && T::compare(left_element, &separator).is_lt();
+            let min = replaces_tail.then(T::min_encoded_key).flatten();
+            let tail = min.as_deref();
+            let mut elements: Vec<&[u8]> = Vec::with_capacity(N);
+            elements.extend((0..i).map(|j| array_element::<N>(left, j)));
+            elements.push(&separator);
+            elements
+                .extend(((i + 1)..N).map(|j| tail.unwrap_or_else(|| array_element::<N>(left, j))));
+
+            // A separator longer than the whole key is no use, however it was assembled. A
+            // smallest encoding longer than the element it replaces can make it longer, so this
+            // is checked before building it, which also keeps every end offset within a `u32`.
+            let total = elements
+                .iter()
+                .map(|x| x.len())
+                .fold(header, usize::saturating_add);
+            if total >= left.len() {
+                return Cow::Borrowed(left);
+            }
+            let mut result = Vec::with_capacity(total);
+            let mut end = header;
+            for element in &elements {
+                end += element.len();
+                result.extend_from_slice(&u32::try_from(end).unwrap().to_le_bytes());
+            }
+            for element in &elements {
+                result.extend_from_slice(element);
+            }
+            return Cow::Owned(result);
+        }
+        // Every element compared equal, so `left` does not sort below `right`
+        Cow::Borrowed(left)
+    }
+}
+
+// The `index`th element of an array encoding of `N` variable width elements
+fn array_element<const N: usize>(data: &[u8], index: usize) -> &[u8] {
+    let start = if index == 0 {
+        size_of::<u32>() * N
+    } else {
+        array_end_offset(data, index - 1)
+    };
+    &data[start..array_end_offset(data, index)]
+}
+
+// The end offset an array encoding stores for its `index`th element
+fn array_end_offset(data: &[u8], index: usize) -> usize {
+    let range = size_of::<u32>() * index..size_of::<u32>() * (index + 1);
+    u32::from_le_bytes(data[range].try_into().unwrap()) as usize
 }
 
 impl Value for &str {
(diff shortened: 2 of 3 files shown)


cberner:
Closing this for now. The approach here only shortens the element the two keys first differ in and carries the rest of the left key, so for an array of long elements most of the key still ends up in the branch node. Dropping the trailing elements instead — keeping the end offsets and giving them zero length — is the version worth having, but it needs the element type to accept a zero-length slice in `compare()`, which `Option`, nested arrays and tuples do not. Written up with a repro in #1404.

---

cberner:
Recording the problem here rather than in a separate issue (#1404 was filed by mistake and is closed).

## What is wrong with the approach in this PR

It finds the element the two keys first differ in, replaces it with that element type's own separator, and keeps the rest of the left key. Correct, but it only shrinks one element. For `[&str; 3]` whose keys differ in the first element, two of the three are still carried in full, so most of the key ends up in the branch node anyway.

The version worth having drops the trailing elements: keep the end offsets so the encoding still parses, and give every element after the differing one zero length.

## Why that cannot be done today

It needs two conditions, not one:

1. **The shortened element must sort strictly above the left key's element.** Otherwise `compare(left, separator)` gets `Equal` there and walks on into the zeroed tail. Cheap to test with a single `T::compare`, and it fails exactly when `T::separator()` had to fall back to returning `left`.

2. **`T` must accept a zero-length slice in `compare()`.** This is the blocker. True for `&[u8]`, `&str` and `String`; false for every composite — `Option<T>` indexes byte 0 for its tag, `[T; N]` reads `4 * N` offset bytes, tuples read length varints. For a user-defined `T` it is unknowable.

Condition 2 is easy to miss because the separator's own contract does not expose it. `compare(left, separator)` and `compare(separator, right)` both resolve at the differing element and never touch the zeroed tail, so both assertions in `branch_separator()` pass. It is the **lookup** path that reaches it: a search key whose leading elements equal the separator's — easy to construct, since the separator's element is a prefix of some real key's element — compares `Equal` through the differing element and then calls `T::compare(search_element, &[])`.

For `[Option<&str>; 2]` that panics:

```
panic_bounds_check
  <Option<T> as Key>::compare        src/types.rs:387   <- data2[0], reading the tag
  <[T; N] as Key>::compare           src/types.rs:598
```

Repro: build a separator for `[Some("aaaa"), Some("zzzz")]` / `[Some("bbbb"), Some("yyyy")]` with element 0 set to `Option::<&str>::separator(...)` and element 1 given zero length, then compare `[Some("b"), Some("x")]` against it. The same construction over `[&str; 2]` is fine, which is what makes it easy to convince yourself the idea works.

## What would unlock it

A minimal encoding hook on `Key`, something like `fn smallest() -> Option<Cow<'static, [u8]>>` defaulting to `None`. `&[u8]`, `&str` and `String` return the empty slice; `Option<T>` returns its one byte `None` encoding; tuples and arrays compose theirs from their elements'. The tail then collapses to each element's minimum rather than to literally nothing, which still drops nearly all of it — and only when condition 1 holds and every trailing element type supplies a minimum.

That is a public addition to `Key`, on top of `separator()` itself being new and unreleased, so it wants to be its own change rather than bolted onto this one.

---

src/types.rs
```diff
@@ -626,6 +626,66 @@ impl<const N: usize, T: Key> Key for [T; N] {
...
+        if T::fixed_width().is_some() {
+            return Cow::Borrowed(left);
+        }
+        // Only the element the two first differ in can be shortened: the elements before it are
+        // needed for the result to sort below `right`. The elements after it can be replaced by
+        // the smallest encoding, but only once the shortened element sorts strictly above
+        // `left`'s, so that both comparisons resolve there and never read what follows.
+        let header = size_of::<u32>() * N;
+        let left_ends: [usize; N] = core::array::from_fn(|i| array_end_offset(left, i));
+        let right_ends: [usize; N] = core::array::from_fn(|i| array_end_offset(right, i));
```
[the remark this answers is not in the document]
cberner: Fixed in 2b71d2d. Both arrays are gone — they were pure caching, since each end offset is a `u32` read straight out of the header:

```rust
let element = |data: &'a [u8], i: usize| -> &'a [u8] {
    let start = if i == 0 {
        header
    } else {
        array_end_offset(data, i - 1)
    };
    &data[start..array_end_offset(data, i)]
};
```

Stack use is now constant in `N`, and the code is shorter for it. The only remaining `N`-sized allocation is the `Vec<&[u8]>` of elements, which is on the heap and no larger than what `from_bytes()` already builds.

Worth noting the premise about ordinary use was not quite right: `[T; N]::SelfType` *is* `[T::SelfType; N]`, so `from_bytes()` returns an `N`-element array by value and already puts 4.8 MB on the stack for a `[&str; 300_000]` key. Reading such a key back was never viable. But a key can be inserted without ever being deserialized, so a separator computed during insert really was reachable, and constant stack use is the better property regardless.

---

src/types.rs
```diff
@@ -626,6 +626,66 @@ impl<const N: usize, T: Key> Key for [T; N] {
...
+            let min = T::min_encoded_key().filter(|_| T::compare(left_element, &separator).is_lt());
+            let tail = min.as_deref();
+            let mut elements: Vec<&[u8]> = Vec::with_capacity(N);
+            elements.extend((0..i).map(|j| element(left, &left_ends, j)));
+            elements.push(&separator);
+            elements
+                .extend(((i + 1)..N).map(|j| tail.unwrap_or_else(|| element(left, &left_ends, j))));
+
+            let total: usize = elements.iter().map(|x| x.len()).sum();
+            let mut result = Vec::with_capacity(header + total);
```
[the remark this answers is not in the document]
cberner: Correct, and fixed in 2b71d2d. The size is now computed and checked before anything is allocated or written:

```rust
let total = elements.iter().map(|x| x.len()).fold(header, usize::saturating_add);
if total >= left.len() {
    return Cow::Borrowed(left);
}
let mut result = Vec::with_capacity(total);
```

`saturating_add` for the checked arithmetic you asked for — saturation lands on `usize::MAX`, which takes the fallback, so it cannot wrap into a small total.

That also retires the `u32::try_from()` concern rather than needing its own guard. Past the check, every `end` is at most `total`, which is strictly less than `left.len()`; and `left.len()` fits in a `u32` because it *is* the last end offset a valid array encoding stores. So the conversion cannot fail.

`separator_is_not_grown_by_a_wide_minimum` covers it with a conforming `WideMinimum` key type whose smallest value encodes to 64 bytes: for `[WideMinimum; 3]`, the pre-fix code turned a 17 byte key into a 141 byte separator, and now returns the whole key.

The same shape existed in #1402's `tuple_separator()` — it built the result before comparing lengths — so I applied the equivalent guard and test there too.

---

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@
   `Option` keys shorten their payload, when the wrapped type is variable width.
 * Add `Key::min_encoded_key()`, the encoding of a key type's smallest value. Implementing it is
   optional, and lets container types holding that key store shorter separators.
+* Shorten separators for array keys, when the element type is variable width. More children fit
+  in each internal btree node, so lookups touch fewer pages.
```
cberner: Don't talk about implementation details just say slightly less space and faster lookup

src/types.rs
```diff
@@ -626,6 +626,72 @@ impl<const N: usize, T: Key> Key for [T; N] {
...
+        // `left`'s, so that both comparisons resolve there and never read what follows.
+        let header = size_of::<u32>() * N;
+        let element = |data: &'a [u8], i: usize| -> &'a [u8] {
+            let start = if i == 0 {
+                header
+            } else {
+                array_end_offset(data, i - 1)
+            };
+            &data[start..array_end_offset(data, i)]
+        };
```
cberner: I think this would be easier to read as a free function. N can be a const generic parameter

cberner: Done in ffa4a8e. `array_element::<N>()` sits next to `array_end_offset()`, and the two read as a pair:

```rust
// The `index`th element of an array encoding of `N` variable width elements
fn array_element<const N: usize>(data: &[u8], index: usize) -> &[u8] {
    let start = if index == 0 {
        size_of::<u32>() * N
    } else {
        array_end_offset(data, index - 1)
    };
    &data[start..array_end_offset(data, index)]
}
```

---

CHANGELOG.md
cberner: Done in ffa4a8e:

> * Shorten separators for array keys, when the element type is variable width. Tables with such keys use slightly less space, and lookups are faster.

Made the same edit to #1402's tuple entry, since it read the same way.

---

src/types.rs
```diff
@@ -626,6 +626,75 @@ impl<const N: usize, T: Key> Key for [T; N] {
...
+
+        for i in 0..N {
+            let left_element = array_element::<N>(left, i);
+            let right_element = array_element::<N>(right, i);
+            if T::compare(left_element, right_element).is_eq() {
+                continue;
+            }
+            let separator = T::separator(left_element, right_element);
+            // The tail may only be replaced once this element sorts strictly above `left`'s
+            let min = T::min_encoded_key().filter(|_| T::compare(left_element, &separator).is_lt());
```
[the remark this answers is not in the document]
cberner: Right on both cases, and fixed in 61fefdf. `Option::filter` evaluates its receiver, so the hook ran even when the result was thrown away. Both conditions are now checked first:

```rust
// There is a tail to replace only when this is not the last element, and it may only
// be replaced once this element sorts strictly above `left`'s. `min_encoded_key()`
// may allocate, so it is asked only when the answer will be used.
let replaces_tail = i + 1 < N && T::compare(left_element, &separator).is_lt();
let min = replaces_tail.then(T::min_encoded_key).flatten();
```

Purely a matter of how often the hook is called — `filter` was already discarding the value in the strictness case, and the `i + 1 == N` case never read it — so no test changes.

#1402's `tuple_separator()` does not have this problem: it reaches the hook through `TupleElement::discard()`, which is called per element and only for elements the separator has decided to replace.

---

PR #1402, opened 2026-08-20
Shorten separators for tuple keys

cberner:
A tuple is variable width when any element type is, so it is one of the built-in key types whose branch keys `separator()` can shorten. This is the composite most likely to be a real key: `(u64, &str)` and `(&str, u64)` are common table keys.

`tuple_separator()` splits both encodings into their elements, finds the first pair that differ, and builds one candidate: the shared elements, then an element that sorts above the left key's — one strictly between the differing pair when the element type offers it, the right key's own otherwise — then the smallest encoding of each remaining type, or the right key's element where a type has none. The candidate always sorts above the left key, because its element at the differing position does. Whether it sorts below the right key depends on what the tail collapsed to, so that is checked with `compare_elements()` — a lexicographic comparison using the element types' own `compare()` — rather than argued case by case. `(1, "x")` against `(2, "")` is the case that fails it: collapsing produces `(2, "")`, which is the right key itself, so the whole key is kept instead.

That check is what the earlier version of this PR was missing, and why it was closed: discarding the tail unconditionally corrupted lookups on nested composites. #1406 supplies the other half — an element collapses to its type's smallest *encoding*, not to nothing, so `compare()` can still read it. `(&str, Option<&str>)` is the case that breaks otherwise: `Option::compare()` reads a tag byte, so an empty element leaves it indexing past the end. It collapses to `None` instead.

`encode_if_shorter()` then encodes the candidate, rebuilding the length varints — a replaced element's length may differ, and a 300 byte element cut to 4 bytes takes its varint from three bytes down to one — and rejects it before building when it would not be shorter than the whole key, which a smallest encoding longer than the element it replaced can cause. A fixed width element stores no varint, so a replacement must be exactly that wide; every encoding of such a type is, its smallest included, and a `debug_assert` now checks it, since a violation by a nonconforming `Key` implementation would corrupt the encoding silently.

The last element is not a special case: with the encodings split up front it is just the final index, and a candidate whose differing element is the last has no tail to check. `(T,)` delegates to `T`, which it is encoded and compared exactly like.

## Testing

- `single_element_tuple_separator`, `last_element_separator` and `leading_element_separator` pin the exact separator each strictly-between case produces. The last covers a shortened element in the middle of a 3-tuple, and a varint shrinking from three bytes to one.
- `right_element_separator` covers the two shapes that take the right key's element: a fixed width leading element, and a variable width one with nothing strictly between.
- `right_element_is_not_taken_without_a_smaller_tail` covers both ways the collapsed tail can fail to bring the candidate below the right key — the right key's tail already being the smallest encoding, and an element type with no smallest encoding at all.
- `elements_after_the_shortened_one_are_discarded` covers what each kind of trailing element collapses to: `&str` to empty, `Option<&str>` to `None`, a fixed width `Option<u64>` to its equally wide `None`, and a `u64` or `[&str; 2]` (no smallest encoding) staying the right key's.
- `separator_returns_full_key_when_nothing_is_shorter` covers the five ways nothing shorter exists.
- `separator_is_not_grown_by_a_wide_minimum` uses a conforming `WideMinimum` key type whose smallest value encodes to 64 bytes, so collapsing two of them would more than double the key.
- Every one of those goes through `check_separator`, which also checks the `[left, right)` bounds and that the separator is an encoding of the key type.
- `tuple_keys_with_shortened_separators` builds multi-level trees of 2,000 `(&str, &str)` and `(&str, Option<&str>)` keys whose separators collapse the trailing element, checks every lookup and the iteration order, then removes three quarters of them to exercise branch merges and rebalancing, and re-checks. `branch_separator()`'s debug assertions validate the bounds of every separator the tree stores along the way.
- `tuple_separators_shrink_branch_nodes` stores the same 5,000 keys twice, as `(&str, &str)` and as `(&str, [&str; 1])` — identical bytes, differing only in whether the trailing element has a smallest encoding to collapse to. 4 branch pages against 17.
- `fixed_width_leading_element_separators_shrink_branch_nodes` does the same for keys that differ in a fixed width leading element, `(u64, &str)` against `(u64, [&str; 1])`. 3 branch pages against 13; that shape produced 12 before taking the right key's element.
- `separators_are_encodings` (already on master, from #1405) starts exercising `(&str, &str)`, `(&str, u64)`, `(u64, &str)` and `(&str, Option<&str>)` with this change.

`cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test --all-features` all pass, as does the `thumbv7em-none-eabihf` no_std check. `just`/podman are unavailable in this environment, so those ran directly rather than through `just test`; no dependencies changed, so `cargo deny check licenses` was not run. The fuzz harness was not run: its keys are `u64`, which is fixed width, so it never reaches `separator()` at all.

Beyond the committed tests, two scratch harnesses (not part of this PR):

- A randomized contract check over 2,033,944 checks across 17 tuple shapes — including 12-element tuples (the maximum arity), `()` and `bool` and `[u8; 3]` elements, nested composites in every position down to `((&str, (u64, &str)), &str)`, element lengths straddling both varint width boundaries (253/254 and 65535/65536), and byte-string content containing the varint tag bytes. Each separator was checked for the `[left, right)` bounds, the length bound, the encoding round trip, and — the strongest check — against an independent oracle: three random probe keys per pair, asserting `compare(probe, separator)` equals `Ord` on the decoded values, which is exactly the comparison a lookup routed through a branch node performs.
- A model based btree differential test against a `BTreeMap`: 45,000 random inserts and removes of `(&str, &str)` keys in a debug build (every btree debug assertion active), with `check_integrity()`, full-order iteration equality, and random present-and-absent lookups after each of 30 commits, reaching tree height 3, then a full drain to empty. All matched exactly.

Codecov flags two spots in the new code, both deliberate: `WideMinimum::type_name()`, which the test type has to define but nothing calls, and the no-differing-element fallback in `tuple_separator()`, which the `left < right` precondition makes unreachable.

## Notes

Last of the four. #1405 (the `separator()` contract), #1406 (`min_encoded_key()`) and #1401 (arrays) have all merged, and this is rebased on top of them. #1401 carries its own copy of the `WideMinimum` test type, since neither PR depended on the other.

`tuple_separator()` reaches the element operations through `ElementType`, a table of function pointers, because it is generic only over the number of elements — the macro-generated impls are what can name the element types. Those calls stay indirect but are off the hot path: inserting 100k `(&str, &str)` keys makes 1,334,058 element `compare()` calls against 4,525 element `separator()` calls, at ~250ns per separator — under 0.3% of insert time. Element `compare()`, which is the hot one, is statically dispatched and fully inlined.

Tuples could also now supply a `min_encoded_key()` of their own — concatenating their elements' — which the associated const this started as could not express. Left out here to keep this to one change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AcCEQ8N51j9NsV1dLHDnPL)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@
   optional, and lets container types holding that key store shorter separators.
 * Shorten separators for array keys, when the element type is variable width. Tables with such
   keys use slightly less space, and lookups are faster.
+* Shorten separators for tuple keys, when at least one element type is variable width. Tables
+  with such keys use slightly less space, and lookups are faster.
 * Fix a crash shortly after a commit being able to silently roll that commit back during
   recovery, if `check_integrity()` had previously repaired the database.
 * Fix iterators silently omitting data when iteration continues after an error. An iterator
--- a/src/tuple_types.rs
+++ b/src/tuple_types.rs
@@ -65,6 +65,164 @@ fn not_equal<T: Key>(data1: &[u8], data2: &[u8]) -> Option<Ordering> {
     }
 }
 
+// The `Key` operations of one element's type, so that `tuple_separator()`, which is generic only
+// over the number of elements, can call them
+#[derive(Clone, Copy)]
+struct ElementType {
+    fixed_width: Option<usize>,
+    min_encoded_key: fn() -> Option<Cow<'static, [u8]>>,
+    compare: fn(&[u8], &[u8]) -> Ordering,
+    separator: for<'a> fn(&'a [u8], &'a [u8]) -> Cow<'a, [u8]>,
+}
+
+impl ElementType {
+    fn of<T: Key>() -> Self {
+        Self {
+            fixed_width: T::fixed_width(),
+            min_encoded_key: T::min_encoded_key,
+            compare: T::compare,
+            separator: T::separator,
+        }
+    }
+
+    // An encoding that sorts strictly between `left` and `right`, or `None` when the type does
+    // not offer one: `separator()` only guarantees a result in `[left, right)`
+    fn between<'a>(&self, left: &'a [u8], right: &'a [u8]) -> Option<Cow<'a, [u8]>> {
+        let separator = (self.separator)(left, right);
+        if (self.compare)(left, &separator).is_lt() {
+            Some(separator)
+        } else {
+            None
+        }
+    }
+
+    // The smallest encoding of the type, or `fallback` for a type that has none
+    fn smallest_or<'a>(&self, fallback: &'a [u8]) -> Cow<'a, [u8]> {
+        match (self.min_encoded_key)() {
+            Some(smallest) => smallest,
+            None => Cow::Borrowed(fallback),
+        }
+    }
+}
+
+// The elements of a variable width tuple encoding: the `N` leading ones, whose lengths the header
+// gives, then the last, which runs to the end
+fn split_elements<const N: usize>(fixed_width: [Option<usize>; N], data: &[u8]) -> Vec<&[u8]> {
+    let (header, lens) = parse_lens::<N>(fixed_width, data);
+    let mut elements = Vec::with_capacity(N + 1);
+    let mut offset = header;
+    for len in lens {
+        elements.push(&data[offset..(offset + len)]);
+        offset += len;
+    }
+    elements.push(&data[offset..]);
+    elements
+}
+
+// Lexicographically compares two tuples, given as their encoded elements
+fn compare_elements<A: AsRef<[u8]>, B: AsRef<[u8]>>(
+    types: &[ElementType],
+    a: &[A],
+    b: &[B],
+) -> Ordering {
+    for i in 0..types.len() {
+        let ordering = (types[i].compare)(a[i].as_ref(), b[i].as_ref());
+        if !ordering.is_eq() {
+            return ordering;
+        }
+    }
+    Ordering::Equal
+}
+
+// Encodes the tuple with these elements, or returns `None` when the encoding would be at least
+// as long as `left`. The length is checked first, so an oversized encoding -- a smallest
+// encoding can be longer than the element it replaced -- is never built.
+fn encode_if_shorter(
+    types: &[ElementType],
+    elements: &[Cow<'_, [u8]>],
+    left: &[u8],
+) -> Option<Vec<u8>> {
+    // A fixed width element's length is not recorded, so a replacement must be exactly that wide
+    debug_assert!(types.iter().zip(elements).all(|(element_type, element)| {
+        element_type
+            .fixed_width
+            .is_none_or(|width| element.len() == width)
+    }));
+    // Length varints for every variable width element but the last, whose length is implicit.
+    // They are rebuilt rather than copied, since a replaced element's length may differ.
+    let mut result = Vec::with_capacity(left.len());
+    for i in 0..(elements.len() - 1) {
+        if types[i].fixed_width.is_none() {
+            encode_varint_len(elements[i].len(), &mut result);
+        }
+    }
+    let mut total = result.len();
+    for element in elements {
+        total = total.saturating_add(element.len());
+    }
+    if total >= left.len() {
+        return None;
+    }
+    for element in elements {
+        result.extend_from_slice(element);
+    }
+    Some(result)
+}
+
+// The separator between two encodings of a variable width tuple. `leading` describes every
+// element but the last, whose length is implicit and which `last` describes.
+fn tuple_separator<'a, const N: usize>(
+    leading: [ElementType; N],
+    last: ElementType,
+    left: &'a [u8],
+    right: &'a [u8],
+) -> Cow<'a, [u8]> {
+    let fixed_width: [Option<usize>; N] = core::array::from_fn(|i| leading[i].fixed_width);
+    let types: Vec<ElementType> = leading.into_iter().chain([last]).collect();
+    let left_elements = split_elements(fixed_width, left);
+    let right_elements = split_elements(fixed_width, right);
+
+    let mut differing = None;
+    for i in 0..types.len() {
+        if !(types[i].compare)(left_elements[i], right_elements[i]).is_eq() {
+            differing = Some(i);
+            break;
+        }
+    }
+    // `left` sorts below `right`, so some element differs
+    let Some(index) = differing else {
+        return Cow::Borrowed(left);
+    };
+
+    // The candidate: the elements the two keys share, then one that sorts above `left`'s --
+    // between the differing pair when the type offers that, `right`'s own otherwise -- then the
+    // smallest encoding of each remaining type, or `right`'s element where a type has none
+    let mut elements = Vec::with_capacity(types.len());
+    for &bytes in &left_elements[..index] {
+        elements.push(Cow::Borrowed(bytes));
+    }
+    match types[index].between(left_elements[index], right_elements[index]) {
+        Some(between) => elements.push(between),
+        None => elements.push(Cow::Borrowed(right_elements[index])),
+    }
+    for (element_type, &bytes) in types[(index + 1)..]
+        .iter()
+        .zip(&right_elements[(index + 1)..])
+    {
+        elements.push(element_type.smallest_or(bytes));
+    }
+
+    // The candidate sorts above `left`, since its element at `index` does, but collapsing may
+    // have left it at or above `right`
+    if compare_elements(&types, &elements, &right_elements).is_ge() {
+        return Cow::Borrowed(left);
+    }
+    match encode_if_shorter(&types, &elements, left) {
+        Some(separator) => Cow::Owned(separator),
+        None => Cow::Borrowed(left),
+    }
+}
+
 macro_rules! fixed_width_impl {
     ( $( $t:ty ),+ ) => {
         {
@@ -276,6 +434,20 @@ macro_rules! tuple_impl {
                     compare_variable_impl!(data1, data2 $(,$t,$i)+ | $t_last, $i_last)
                 }
             }
+
+            fn separator<'a>(left: &'a [u8], right: &'a [u8]) -> Cow<'a, [u8]> {
+                // A fixed width tuple's encodings are compared at that width, so nothing shorter
+                // than `left` may be returned
+                if Self::fixed_width().is_some() {
+                    return Cow::Borrowed(left);
+                }
+                tuple_separator::<$i_last>(
+                    [$(ElementType::of::<$t>(),)+],
+                    ElementType::of::<$t_last>(),
+                    left,
+                    right,
+                )
+            }
         }
     };
 }
@@ -318,6 +490,11 @@ impl<T: Value> Value for (T,) {
 impl<T: Key> Key for (T,) {
     fn compare(data1: &[u8], data2: &[u8]) -> Ordering {
         T::compare(data1, data2)
+    }
+
+    // Encoded and compared exactly as `T`, so it separates the same way
+    fn separator<'a>(left: &'a [u8], right: &'a [u8]) -> Cow<'a, [u8]> {
+        T::separator(left, right)
     }
 
     // Encoded exactly as `T`, so its smallest value encodes the same way
@@ -438,7 +615,10 @@ tuple_impl! {
 
 #[cfg(test)]
 mod test {
-    use crate::types::Value;
+    use crate::types::{Key, TypeName, Value};
+    use alloc::borrow::Cow;
+    use alloc::format;
+    use core::cmp::Ordering;
 
     #[test]
     fn width() {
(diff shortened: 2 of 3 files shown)


cberner:
Closing this for now, alongside #1401. The approach here only shortens the element the two keys first differ in and carries the rest of the left key, so for `(&str, &str)` or `(&str, u64)` the tail still ends up in the branch node. Dropping the trailing elements instead needs the element type to accept a zero-length slice in `compare()`, which the composite types do not. Written up with a repro in #1404.

---

cberner:
Recording the problem here rather than in a separate issue (#1404 was filed by mistake and is closed).

## What is wrong with the approach in this PR

It finds the element the two keys first differ in, replaces it with that element type's own separator, and keeps the rest of the left key. Correct, but it only shrinks one element. For `(&str, &str)` or `(&str, u64)` whose keys differ in the first element, the tail is still carried in full into the branch node.

The version worth having drops the trailing elements: keep the length varints so the encoding still parses, and give every element after the differing one zero length. Tuples stand to gain more from this than arrays do — `[T; N]` with a variable width `T` is rare, but `(&str, ...)` keys are common.

## Why that cannot be done today

It needs two conditions, not one:

1. **The shortened element must sort strictly above the left key's element.** Otherwise `compare(left, separator)` gets `Equal` there and walks on into the zeroed tail. Cheap to test with a single `T::compare`, and it fails exactly when the element's `separator()` had to fall back to returning `left`.

2. **The element type must accept a zero-length slice in `compare()`.** This is the blocker. True for `&[u8]`, `&str` and `String`; false for every composite — `Option<T>` indexes byte 0 for its tag, `[T; N]` reads `4 * N` offset bytes, nested tuples read length varints. For a user-defined type it is unknowable.

Condition 2 is easy to miss because the separator's own contract does not expose it. `compare(left, separator)` and `compare(separator, right)` both resolve at the differing element and never touch the zeroed tail, so both assertions in `branch_separator()` pass. It is the **lookup** path that reaches it: a search key whose leading elements equal the separator's — easy to construct, since the separator's element is a prefix of some real key's element — compares `Equal` through the differing element and then calls `compare(search_element, &[])` on the next one.

Demonstrated on the array side in #1401, where `[Option<&str>; 2]` panics in `<Option<T> as Key>::compare` at `src/types.rs:387` reading `data2[0]`. A tuple with an `Option` or nested-tuple element after the differing one fails the same way; `(&str, &str)` and `(&str, u64)` would be fine, which is what makes it easy to convince yourself the idea works.

## What would unlock it

A minimal encoding hook on `Key`, something like `fn smallest() -> Option<Cow<'static, [u8]>>` defaulting to `None`. `&[u8]`, `&str` and `String` return the empty slice; `Option<T>` returns its one byte `None` encoding; tuples and arrays compose theirs from their elements'. The tail then collapses to each element's minimum rather than to literally nothing, which still drops nearly all of it — and only when condition 1 holds and every trailing element type supplies a minimum.

That is a public addition to `Key`, on top of `separator()` itself being new and unreleased, so it wants to be its own change rather than bolted onto this one.

---

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@
   `Option` keys shorten their payload, when the wrapped type is variable width.
 * Add `Key::min_encoded_key()`, the encoding of a key type's smallest value. Implementing it is
   optional, and lets container types holding that key store shorter separators.
+* Shorten separators for tuple keys, when at least one element type is variable width. More
```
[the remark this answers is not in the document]
cberner: This is incorrect on both counts, so there is nothing to change. Unlike the two earlier versions of this finding, the cited SHA is at least real — it is this PR's head — but its authorship is not what the finding says:

```
$ git cat-file -p c371ac8b3d20da0a9b80842918dae8e6ed3c9593
author Christopher Berner <me@cberner.com> 1787284455 +0000
committer Christopher Berner <me@cberner.com> 1787321313 +0000
...
Shorten separators for tuple keys
...
Assisted-by: Claude Code
```

Human author and committer, and the `Assisted-by:` trailer is the last line of the message. Running the repository's own check over this PR's range finds nothing:

```
$ git log --no-merges --pretty=format:'%H %an <%ae> committed by %cn <%ce>' 5e6ad02..c371ac8 \
    | grep -iE '(claude|codex)'
(no output)
```

The `commit-authors` job — which is that same script, and rejects Claude or Codex as author *or* committer — passes on this commit.

This is the third time this finding has appeared across this stack of PRs, each time asserting a Codex author on a commit that has none; the first two cited SHAs that did not exist in the repository at all. Worth flagging in case the check is reading something other than the commit under review.

---

src/tuple_types.rs
```diff
@@ -65,6 +65,130 @@ fn not_equal<T: Key>(data1: &[u8], data2: &[u8]) -> Option<Ordering> {
...
+        let bytes = &left[offset..(offset + lens[index])];
+        let right_bytes = &right[right_offset..(right_offset + right_lens[index])];
+        if (element.compare)(bytes, right_bytes).is_eq() {
+            offset += lens[index];
+            right_offset += right_lens[index];
+            continue;
+        }
+        let Some((separator, strict)) = element.shorten(bytes, right_bytes) else {
+            return Cow::Borrowed(left);
+        };
```
cberner: Instead of returning left, can't we take the right element, and then shorten all the remaining elements to their min_encoding_key()?

src/tuple_types.rs
```diff
@@ -65,6 +65,130 @@ fn not_equal<T: Key>(data1: &[u8], data2: &[u8]) -> Option<Ordering> {
...
+
+        let mut elements = Vec::with_capacity(N + 1);
```
cberner: At this point the loop is going to terminate, but the code is harder to read because its still indented in the loop. Can we break the loop and move the rest of the code out of the loop?

src/tuple_types.rs
cberner: Yes, and it's a big win — done in 89f8ade. It also subsumes the other case that gave up, where the element is variable width but has no separator above `left`'s, so `shorten()` is now just "prefer this when it works" rather than a precondition.

One condition had to come with it. When the shortened element sorts strictly above `left`'s, *both* comparisons resolve at that element, which is why the elements after it are free — nothing ever reads them. `right`'s own element only resolves the comparison against `left`; the one against `right` compares equal there and runs on into what follows. So the tail is no longer unread, and:

- every element after it has to sort at or below `right`'s — a smallest encoding does by definition, and where a type has none, `right`'s own bytes serve;
- at least one has to sort **strictly** below `right`'s, or the result is `right` itself rather than something under it. `(1, "x")` against `(2, "")` is the case: `(2, "")` *is* the right key.

`TupleElement::discard()` now reports whether what it stored sorts strictly below `right`'s, and the assembly bails to the whole key when none of them does.

Measured on 5,000 `(u64, &str)` keys that differ in the `u64` — the shape that previously got no shortening at all:

| | branch pages |
|---|---|
| before | 12 |
| after | 3 |

and against `(u64, [&str; 1])`, whose tail has no smallest encoding to collapse to, 3 against 13. That's the new `fixed_width_leading_element_separators_shrink_branch_nodes`.

New unit tests: `right_element_separator` for both shapes it now handles, and `right_element_is_not_taken_without_a_smaller_tail` for the two ways the tail can fail to sort below `right`'s. I also re-ran the randomized contract check over 1,180,258 ordered pairs across 12 tuple shapes, including nested ones like `((&str, u64), &str)` and `(&str, [&str; 2])` — `left <= separator < right`, never longer than `left`, and always a valid encoding.

---

src/tuple_types.rs
cberner: Done in 89f8ade. The loop now only locates the element, and everything else sits at the top level:

```rust
let mut differing = None;
for (index, element) in leading.iter().enumerate() {
    ...
    if !(element.compare)(bytes, right_bytes).is_eq() {
        differing = Some(index);
        break;
    }
    offset += lens[index];
    right_offset += right_lens[index];
}
let Some(index) = differing else {
    return last_element_separator(last, left, right, offset, right_offset);
};
```

The all-equal case that used to trail the loop became `last_element_separator()`, since it works differently — the last element stores no length and has nothing after it, so there is no header to rebuild.

---

src/tuple_types.rs
```diff
@@ -65,6 +65,180 @@ fn not_equal<T: Key>(data1: &[u8], data2: &[u8]) -> Option<Ordering> {
...
+        match (self.min_encoded_key)() {
+            // A smallest encoding sorts at or below every value, so it never sorts above `right`'s
+            Some(minimum) => {
+                let below = !unread && (self.compare)(&minimum, right).is_lt();
+                (minimum, below)
+            }
+            None if unread => (Cow::Borrowed(left), false),
+            None => (Cow::Borrowed(right), false),
+        }
+    }
```
cberner: All the comments on this function make me think it is doing too much. It is quite hard to understand and reason about

cberner: Agreed — it had grown to five jobs, and the worst of it was that the ordering reasoning was smeared across the whole thing by an `unread` flag threaded into `discard()`, which then returned a tuple whose second half was meaningless half the time. Split up in c92d8de, net 14 lines shorter:

```rust
let separator = (0..types.len())
    .find(|&i| !(types[i].compare)(left_elements[i], right_elements[i]).is_eq())
    .and_then(|index| {
        separator_by_shortening(&types, &left_elements, &right_elements, index)
            .or_else(|| separator_by_taking_right(&types, &left_elements, &right_elements, index))
    })
    .and_then(|elements| assemble(&types, &elements, left));

separator.map_or(Cow::Borrowed(left), Cow::Owned)
```

- `split_elements()` cuts an encoding into its element slices up front. All the offset arithmetic lives here now, and everything downstream just indexes.
- `separator_by_shortening()` and `separator_by_taking_right()` each return the separator's elements or `None`. Each is about a dozen lines, and its ordering argument is entirely local — no flag, no shared bookkeeping.
- `assemble()` rebuilds the length varints and applies the "no longer than `left`" check.

Two things fell out of it. `last_element_separator()` is gone: with the elements split up front, the last element is just `index == types.len() - 1`, and both strategies do the right thing there — shortening keeps everything before it, and taking `right`'s finds no elements after it to sort below `right`, so it correctly declines. And `TupleElement::shorten()` now returns `Option<Cow>` rather than `(Cow, bool)`, since "not strict" and "fixed width" both just mean "no shortened element".

One cost worth flagging: splitting the elements up front adds three small `Vec`s per call, which takes a separator from 98ns to 206ns. That moves all the separators in a 100k-key insert from 0.14% to 0.26% of the time. I'd take that for the clarity, but say the word and I can put the arity-sized pieces back on the stack.

No behavior change: all nine unit tests and the branch page counts are unchanged, and the randomized contract check passes over the same 1,180,258 ordered pairs across 12 tuple shapes.

---

src/tuple_types.rs
```diff
@@ -65,6 +65,166 @@ fn not_equal<T: Key>(data1: &[u8], data2: &[u8]) -> Option<Ordering> {
...
+        elements.push(&data[offset..(offset + len)]);
+        offset += len;
+    }
+    elements.push(&data[offset..]);
+    elements
+}
+
+// The separator's elements, taking a shortened element that sorts strictly above `left`'s. Both
+// comparisons resolve there, so the elements after it are never read again and each collapses to
+// the smallest encoding of its type.
```
cberner: "Both comparisons resolve at that element, so the ones after it are
// never read, and each collapses to the smallest encoding of its type, or stays `left`'s own
// where a type has none"

That reads more like commentary on this diff. Just document what this function does, what the non-obvious parameters (in this case `index`) are.

src/tuple_types.rs
```diff
@@ -65,6 +65,180 @@ fn not_equal<T: Key>(data1: &[u8], data2: &[u8]) -> Option<Ordering> {
...
+        elements.push(element_type.smallest_or(bytes));
+    }
+    Some(elements)
+}
+
+// The separator's elements, replacing element `index` with `right`'s own. That resolves the
+// comparison against `left`, but the one against `right` moves on to the elements after it, which
+// collapse to the smallest encodings of their types, or stay `right`'s own where a type has none.
+// All of those sort at or below `right`'s, so the result sorts below `right` exactly when one of
+// them sorts strictly below.
```
cberner: Same comment. The fact that we need 5 lines of commentary suggests this code is complicated and not obvious what it does

cberner (commented):
Still needs another simplification pass. And I feel like the names of the functions are not the best. Like "separator_taking_right" and "separator_between" and "assemble". Good names would make it clear what they do from the name of the function

src/tuple_types.rs
cberner: Fixed by deleting the function in ee03388. The reason those comments kept needing to carry the ordering argument is that the two strategy functions each encoded half of it, so I replaced the design rather than the wording: build one candidate, then verify it with the types' own `compare()` instead of proving it in prose.

```rust
// The candidate: the elements the two keys share, then one that sorts above `left`'s --
// between the differing pair when the type offers that, `right`'s own otherwise -- then the
// smallest encoding of each remaining type, or `right`'s element where a type has none
...
// The candidate sorts above `left`, since its element at `index` does, but collapsing may
// have left it at or above `right`
if compare_elements(&types, &elements, &right_elements).is_ge() {
    return Cow::Borrowed(left);
}
```

`separator_between` and `separator_taking_right` are both gone; the check subsumes the case analysis they existed to justify.

---

src/tuple_types.rs
cberner: Also gone in ee03388, same fix as the sibling thread: this function's five lines were the "does the tail keep the result below `right`" proof, and it is now a `compare_elements()` call against `right` instead of an argument. The strategy split, the `below_right` flag, and both multi-line justifications are deleted; what remains is one candidate built in three plain loops, one comparison, and `encode_if_shorter()`.

The remaining functions are named for what they do — `split_elements`, `compare_elements`, `encode_if_shorter`, `between`, `smallest_or` — and each carries a one-or-two-line comment saying what it returns.

Net: 21 lines smaller than the previous revision, tests updated where a no-minimum tail element now stays `right`'s instead of `left`'s (either is valid; one rule instead of two), and the randomized contract check passes over the same 1,275,658 ordered pairs across 13 shapes.

---

cberner:
lgtm

PR #1403, opened 2026-08-20
Split the benchmarks that compare against other engines into their own crate

cberner:
Building RocksDB's vendored C++ sources dominated `redb-bench`'s cold compile time, even for the benchmarks that only exercise redb.

Measured on 4 cores, `cargo bench --no-run --bench redb_benchmark` from an empty target directory:

| | before | after |
|---|---|---|
| bench profile | 10m 13s | **40s** |

Rebuilding just `librocksdb-sys` + `rocksdb` accounts for 6m 03s of that, and 2.5 GiB of artifacts.

## Cause

Cargo builds *every* `[dev-dependencies]` entry for a package whenever any of its bench targets is built -- there is no per-target dependency support ([cargo#1359](https://github.com/rust-lang/cargo/issues/1359)). `benches/common.rs` also referenced every engine unconditionally and was pulled in wholesale via `mod common;`, so the redb-only benchmark genuinely linked RocksDB.

A feature flag cannot fix this, which the first revision of this PR discovered the hard way:

- Cargo rejects optional dev-dependencies outright (`dev-dependencies are not allowed to be optional`), so `rocksdb` would have to become a regular dependency.
- Doing that fails `cargo deny check licenses`, pulling in `bindgen` (BSD-3-Clause) and `libloading` (ISC). cargo-deny prunes build-dependencies reached through a dev-dependency edge, so `librocksdb-sys`'s `bindgen` build-dep had never been audited. Verified by moving that one line between sections on master with nothing else changed.

Keeping these engines as dev-dependencies is therefore load-bearing, both for compile time and for the license audit. That means separating them by package, not by feature.

## Changes

- **`redb-bench-compare`** (new) holds the benchmarks that compare against other engines -- `lmdb_benchmark`, `int_benchmark`, `large_values_benchmark`, `syscall_benchmark` -- plus the heed/RocksDB/fjall/SQLite adapters in `benches/common/mod.rs`. It is the only crate depending on those engines, and they remain dev-dependencies. `syscall_benchmark` moved here because it measures against lmdb; that keeps `redb-bench` free of native engine dependencies entirely.
- **`redb-bench`** keeps the benchmarks that only exercise redb, and the shared harness graduates from `benches/common.rs` to `src/lib.rs`: the `BenchDatabase` trait family, the workloads, `print_results_table`, the redb adapter, and `benchmark_dir()`. It is now compiled once rather than once per benchmark.
- That split also retires two pieces of incidental overhead: `benches/common.rs` was auto-discovered by Cargo as a *tenth* bench target (`benches/*.rs` globbing), and the `#[expect(dead_code)]` / `#[allow(dead_code)]` attributes each benchmark needed because it pulled in the whole module. Bench targets drop from 10 to 9, all of them real benchmarks.
- **`just bench`** becomes one fixed command with nothing to keep in sync: `cargo bench -p redb-bench -p redb-bench-compare --bench {{bench}}`. Cargo builds only the package owning the selected target -- verified that naming a package whose targets are not selected does not build its dev-dependencies -- so `just bench` stays lean and `just bench lmdb_benchmark` still gets RocksDB. All 9 benchmark names resolve through it.
- `Dockerfile.bench` names both packages, since `bench_containerized` defaults to `lmdb_benchmark`.
- `AGENTS.md` explains the split and says to add new engine comparisons to `redb-bench-compare`.

`Cargo.lock` is untracked, so nothing to update there.

## Verification

- `cargo deny --workspace --all-features check licenses advisories`: **licenses ok, advisories ok** -- the failure that killed the first revision.
- `cargo clippy --all --all-targets -- -Dwarnings` (the default-features CI step): clean.
- `RUSTFLAGS=--deny warnings cargo clippy --all --all-targets --all-features`: clean.
- `cargo build --all --all-targets --all-features`, `cargo doc --all --no-deps`, `RUST_BACKTRACE=1 cargo test --all --all-features`: all pass, 0 failures.
- `cargo fmt --all -- --check`: clean. Source is ASCII-only.
- Cold target dir: `redb_benchmark` builds in 40s with no RocksDB, bindgen, clang-sys, SQLite, lmdb, or fjall in the graph.
- Diffed the set of top-level items in the old `common.rs` against the two new files: all 133 preserved, plus `benchmark_dir()` absorbing the deleted `benches/benchmark_dir/mod.rs`.
- `just` 1.36.0 parses the justfile and `--dry-run` renders the right command.

These ran directly rather than through `just test_all`, since rootless podman is not available in this environment.


justfile
```diff
@@ -84,7 +84,15 @@ coverage:
...
 bench bench='redb_benchmark': pre
-    ./scripts/podman-sandbox.sh 'REDB_BENCHMARK_DIR=/scratch cargo bench --frozen -p redb-bench --bench {{bench}}'
+    #!/usr/bin/env bash
+    set -euo pipefail
+    # Enable the "rocksdb" feature only for the benchmarks that declare it in required-features, so
+    # that the others do not pay for building RocksDB. Keep this list in sync with Cargo.toml.
+    features=""
+    case '{{bench}}' in
+        int_benchmark|large_values_benchmark|lmdb_benchmark) features="--features rocksdb" ;;
+    esac
```
cberner: Ugh this looks very errorprone. Now I need to remember to keep this in sync. Isn't there some better way to do this?

AGENTS.md
```diff
@@ -62,6 +62,13 @@ different bench name to compare against other engines, e.g.
 `just bench lmdb_benchmark`. `just bench_containerized` runs benchmarks inside
 a memory-limited Docker container (see `Dockerfile.bench`).
 
+Building RocksDB's vendored C++ sources dominates `redb-bench`'s compile time, so
```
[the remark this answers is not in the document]
cberner: This is incorrect. The commit's author and committer are both `Christopher Berner <me@cberner.com>`, and the agent is credited with an `Assisted-by:` trailer, exactly as AGENTS.md requires. Running the `commit-authors` job's own logic locally against the real commit range produced no offenders for either check, and the job passed in CI.

---

justfile
cberner: Agreed, and it turned out the sync list was a symptom rather than the problem. I've rewritten the PR to split the comparison benchmarks into a separate crate instead, and the recipe is now one fixed command with nothing to keep in sync:

```
cargo bench --frozen -p redb-bench -p redb-bench-compare --bench {{bench}}
```

Naming both packages is safe: Cargo builds only the package that owns the selected target, so listing `redb-bench-compare` does not build its dev-dependencies when you ask for `redb_benchmark`. I verified that separately, and that all 9 benchmark names resolve through this one command.

Two things pushed it this way rather than the feature flag. Cargo rejects optional dev-dependencies outright, so `rocksdb` had to become a regular dependency -- and that broke `cargo deny check licenses` on `bindgen` (BSD-3-Clause) and `libloading` (ISC), because cargo-deny prunes build-dependencies reached through a dev-dependency edge and had therefore never audited `librocksdb-sys`'s `bindgen`. Keeping these engines as dev-dependencies is load-bearing for the license audit as well as for compile time, so separating by package is the only thing that actually works.

Cold build of `redb_benchmark` is now 40s, down from 10m13s -- better than the 1m45s the feature approach gave, since `redb-bench` no longer builds lmdb, SQLite, or fjall either.

---

PR #1405, opened 2026-08-20
Require a separator to be an encoding of the key type

cberner:
`separator()` documented the result as needing only to sort between the two keys, and explicitly relaxed it from being a valid encoding:

> it is passed to `compare()`, and never to `Value::from_bytes()`, so it need not be a valid encoding of `Self`

That relaxation is only worth anything to a `Key` whose `compare()` reads byte strings its `from_bytes()` would reject, and it hides an obligation the two ordering bounds do not. A branch key is compared against every key of the type, not just the two it was built from, so bytes that order correctly against those two can still be unreadable against a third — and the two bounds still hold, so testing them does not catch it.

This replaces the relaxation with the plain requirement:

> The result must be an encoding of `Self` that is greater than or equal to `left`, and less than `right`, under `compare()`.

which subsumes the readability obligation, since an encoding is by definition what `compare()` handles.

Nothing in the crate uses the relaxation. `&str` already rounds its cut up to a character boundary to stay valid UTF-8 — its comment says "the cut must keep it valid", which is really the requirement asserting itself despite what the docs allowed. I checked the rest by round-tripping every separator through `from_bytes()`/`as_bytes()` across 13 key shapes, several thousand per shape: all of them already produce encodings.

Also added, from hitting it while writing composite separators: the result should never be longer than `left`, which is always available to fall back to.

## Testing

`branch_separators_are_not_deserialized` existed to pin the relaxation — its `SeparatorKey` deliberately emits invalid separators and asserts redb never deserializes them. That is a test of the promise being dropped, using a type that no longer satisfies the contract, so it is replaced by `separators_are_encodings`, which checks that separators deserialize and re-serialize to themselves across the leaf and composite key types, including a cut that would land inside a multi-byte character.

redb still does not deserialize branch keys; it just no longer promises not to. Two internal comments in `btree_base.rs` that described them as "never deserialized" / "need not be valid encodings" are updated to match.

`cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test --all-features` all pass. `just`/podman are unavailable in this environment, so those ran directly rather than through `just test`; no dependencies changed, so `cargo deny check licenses` was not run.

## Notes

No changelog entry: `separator()` is unreleased, and the 4.3.0 entry never described the relaxation.

This is the first of three. #1406 adds `Key::MIN_ENCODED_KEY`, and the tuple and array separators that use it follow. They touch adjacent lines in `src/types.rs`, so whichever merges second needs a trivial rebase.

### DIFF
--- a/src/tree_store/btree_base.rs
+++ b/src/tree_store/btree_base.rs
@@ -29,9 +29,9 @@ pub(super) type Checksum = u128;
 pub(super) const DEFERRED: Checksum = 999;
 
 // The key to store in a branch node between a child whose greatest key is `left` and the next,
-// whose least key is `right`. Branch keys only route lookups -- they are compared, never
-// deserialized -- so a `Key` implementation may shorten them, which packs more children into each
-// branch page.
+// whose least key is `right`. Branch keys only route lookups, so a `Key` implementation may
+// shorten them to any encoding that still sorts between the two, which packs more children into
+// each branch page.
 pub(super) fn branch_separator<'a, K: Key>(left: &'a [u8], right: &'a [u8]) -> Cow<'a, [u8]> {
     debug_assert!(K::compare(left, right).is_lt());
     // Branch keys of a fixed width type are stored at that stride, so a shorter separator would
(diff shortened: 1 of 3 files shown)


cberner (commented):
Lgtm

PR #1407, opened 2026-08-21
API-5 polish: io module visibility, Debug and doc fixes

cberner:
Three small fixes from the public API review, one commit each:

* **Expose the `io` module whenever `experimental-api-5` is enabled.** It was public only in the no_std configuration, so enabling `std` anywhere in the build graph removed it out from under a no_std backend crate naming `redb::io::Error`. The visibility is now gated on the feature flag; with std the module re-exports `std::io::Error`, and the 4.x surface (flag off) is unchanged.
* **Stop `InMemoryBackend`'s `Debug` from printing the database.** The derived impl printed the entire `Vec<u8>`, and opening a database formats its backend with `{:?}` when the `logging` feature is enabled. It now reports the storage length, reading through a poisoned lock so the output stays meaningful after a writer panic.
* **Fix broken doc links in the savepoint methods.** Inverted backtick/bracket syntax like `` `[SavepointError::InvalidSavepoint`] `` now renders as real intra-doc links, the reference to a nonexistent `delete_savepoint()` points at `delete_persistent_savepoint()`, and `SetDurabilityError`'s summary line no longer claims to be about transactions in general.

Verified with `cargo fmt --check`, clippy (`--all-targets`, with and without `--all-features`, under `--deny warnings`), the full `cargo test --frozen --all-features` suite, `cargo deny check licenses advisories`, rustdoc with `--deny warnings` (default and flag builds), the no_std `cargo check`, and the WASI test suite under wasmtime. Also compile-checked externally that `redb::io::Error` is nameable (and is `std::io::Error`) with the flag on, and remains private with it off.

Note: `cargo doc --no-deps --all-features` fails on master too, from an unescaped `DateTime<FixedOffset>` in chrono_v0_4.rs's module doc; not touched here.

https://claude.ai/code/session_015eKG5UgXsfZKfGRCcafVSU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
   target with atomic compare-and-swap -- Cortex-M3 and above, but not Cortex-M0. The file backend
   and everything that opens a database from a path or a `File` are unavailable in that mode, as is
   `ReadOnlyDatabase`; storage is supplied through `Builder::create_with_backend()`, and
-  `StorageBackend` reports failures as `redb::io::Error`, a module redb makes public only in that
-  mode; with std it stays private and the type is `std::io::Error`, as before. The `cache_metrics`,
+  `StorageBackend` reports failures as `redb::io::Error`; the `io` module is public whenever
+  `experimental-api-5` is enabled, and with std it re-exports `std::io::Error`. The `cache_metrics`,
   `chrono_v0_4` and `uuid` features are unavailable there: the first needs 64-bit atomics, and the
   other two link against the standard library. Leaving `experimental-api-5` off keeps the std build
   regardless of the `std` feature, as before.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,12 +107,11 @@ pub mod backends;
 mod complex_types;
 mod db;
 mod error;
-// Public only where it is needed: without std a backend author has to be able to name these types
-// to implement `StorageBackend`. With std they are re-exports of `std::io`, which the caller
-// already has, so the module stays private and redb adds no public surface.
-#[cfg(redb_no_std)]
+// Public under the redb 5 API preview, so a backend author can name the error type in both
+// modes; with std it is a re-export of `std::io`
+#[cfg(feature = "experimental-api-5")]
 pub mod io;
-#[cfg(not(redb_no_std))]
+#[cfg(not(feature = "experimental-api-5"))]
 mod io;
 #[cfg(feature = "experimental-api-5")]
 mod key_range;
--- a/src/tree_store/page_store/backends.rs
+++ b/src/tree_store/page_store/backends.rs
@@ -6,6 +6,7 @@ use crate::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 #[cfg(not(redb_no_std))]
 use alloc::boxed::Box;
 use alloc::vec::Vec;
+use core::fmt::{Debug, Formatter};
 
 #[cfg(not(redb_no_std))]
 #[derive(Debug)]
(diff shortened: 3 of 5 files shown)


src/tree_store/page_store/backends.rs
```diff
@@ -48,9 +49,21 @@ impl StorageBackend for ReadOnlyBackend {
...
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub struct InMemoryBackend(RwLock<Vec<u8>>);
 
+// Hand-written: the derived impl would print every byte of the database, and opening a database
+// formats its backend into the debug log when the "logging" feature is enabled. The length stays
+// meaningful when a writer panicked, so a poisoned lock is read through rather than branched on
+impl Debug for InMemoryBackend {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        let guard = self.0.read().unwrap_or_else(|error| error.into_inner());
```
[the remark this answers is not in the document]
cberner: Leaving this blocking, deliberately. A non-blocking read isn't portable here: the no_std spin `RwLock` in `crate::sync` has no `try_read`, and its own `Debug` impl already blocks, so `<locked>` would need new lock plumbing for a std-only behavior. It would also make the output nondeterministic under momentary contention and lose the length-after-poison reporting this impl was written for. The critical sections guarding the `Vec` are short memcpys on a test-oriented backend, so the worst case is a brief wait, and there's no re-entrancy path that could deadlock.

---

CHANGELOG.md
```diff
@@ -6,8 +6,10 @@
   target with atomic compare-and-swap -- Cortex-M3 and above, but not Cortex-M0. The file backend
   and everything that opens a database from a path or a `File` are unavailable in that mode, as is
   `ReadOnlyDatabase`; storage is supplied through `Builder::create_with_backend()`, and
-  `StorageBackend` reports failures as `redb::io::Error`, a module redb makes public only in that
-  mode; with std it stays private and the type is `std::io::Error`, as before. The `cache_metrics`,
+  `StorageBackend` reports failures as `redb::io::Error`. The `io` module is public whenever
+  `experimental-api-5` is enabled, not only without std -- a cargo feature must never remove
+  API -- and with std it re-exports `std::io::Error`, so the type is unchanged and a backend
+  written against the module builds in both modes. The `cache_metrics`,
```
cberner: Do not explain stuff about cargo. This should be like one short sentence

CHANGELOG.md
```diff
@@ -81,6 +87,9 @@
   the same file now fails with `DatabaseAlreadyOpen` where the platform supports file locks,
   and warns via the `logging` feature where it does not, instead of the lock always being
   silently skipped.
+* `InMemoryBackend`'s `Debug` output reports the storage length instead of printing every byte
+  of the database, which opening the database also formatted into the debug log when the
+  `logging` feature was enabled.
```
cberner: Remove

CHANGELOG.md
```diff
@@ -28,7 +30,11 @@
 * Add `ReadableMultimapTable::lower_bound()` and `ReadableMultimapTable::upper_bound()`, behind
   the `experimental-api-5` feature flag, returning a `MultimapCursor` pointing at a gap between
   entries. The type reserves the constructors' signatures in the trait; navigation methods will
-  be added behind the `experimental_cursor` feature flag, like the table cursors' were.
+  be added behind the `experimental_cursor` feature flag, like the table cursors' were. Under the
+  flag, these constructors and the `ReadableTable` ones are required methods: an implementation
+  of either trait must add them, alongside its `range()` signature update. `Cursor` and
+  `MultimapCursor` have no public constructor, so an implementation supplies them by delegating
+  to a table that redb provides.
```
cberner: We don't need to document this in the notes

src/lib.rs
```diff
@@ -107,12 +107,14 @@ pub mod backends;
...
-// Public only where it is needed: without std a backend author has to be able to name these types
-// to implement `StorageBackend`. With std they are re-exports of `std::io`, which the caller
-// already has, so the module stays private and redb adds no public surface.
-#[cfg(redb_no_std)]
+// Public under the redb 5 API preview, with and without std: a backend author has to be able to
+// name the error type to implement `StorageBackend`, and a cargo feature must only ever add API,
+// so the `std` feature must not take this module away from a no_std dependent elsewhere in the
+// build graph. With std the module is a re-export of `std::io`, so a backend written against it
+// builds in both modes. Private in the 4.x surface, which never had it.
+#[cfg(feature = "experimental-api-5")]
```
cberner: Way too verbose

Issue #873, opened 2024-10-04
Implement `Key` for `NonZero…`


Evrey:
Hello, there!

So, i’m modeling a bunch of relational data in redb, and i’m using all the `NonZero…` types as primary keys for their `Option<NonZero…>` optimisations where i store foreign keys. Thanks to the orphan rule, i currently have to jump through ugly wrapper type hoops to get that done. It would be nice if a built-in solution could be provided instead.

Hypothetically, it is possible that an invalid primary key 0 *somehow* makes its way into the database file, which means that reading a `NonZero…` primary key from the database is fallible. Here’s a few suggestions for what could be done in that should-not-happen case, and which one i’d favour:

- Treat reading a 0 key as »no entry found«, returning `None`. I’d prefer this behaviour.
  - Keeps the app running, delegating this case to special logic an app already has to consider.
  - However, this means that ID=0 data inside the database is a zombie data resource leak.
- Add a new error return code, for ID=0 is an example of database corruption.
  - Pretty harsh, though, for otherwise the DB seems to work fine, and it’s *just* unreachable data.

It would be good to then also specialise `Option<NonZero…>` to just map to `.unwrap_or(0)` as a `redb::Value`.

cberner:
I think the Rust `specialization` feature is required to implement it for `Option<NonZero>` because there is already an impl for `Option<T>`. You probably need to implement your own `MyKey` that wraps an `Option<NonZero>`

Evrey:
Oh that’s right. So `Option<NonZero…>` would have to wait or be put behind a `feature = "unstable"`.

0xSoftBoi:
Opening a PR for the half of this that isn't blocked, so nobody duplicates it.

The specialization blocker applies to the `Option<NonZero…>` *niche optimization* — making it encode as `unwrap_or(0)` at the same width as the bare primitive. Implementing `Value`/`Key` for the bare `NonZero…` types isn't blocked by anything, and it's the part that removes the wrapper-type hoops: `Option<NonZeroU32>` then works through the existing blanket `Option<T>` impl, just at one tag byte plus payload rather than the niche-optimized width.

[...]

PR #1409, opened 2026-08-25
Niche-encode Option<NonZero*> without specialization

0xSoftBoi:
Completes #873. **Stacked on #1408** — the first commit here is that PR; review it first, and this becomes a one-commit diff once it merges.

## The claim

The niche optimization asked for in #873 doesn't need specialization at all, so it isn't blocked on [rust-lang/rust#31844](https://github.com/rust-lang/rust/issues/31844) — which has been open since 2016 and is still labelled `I-unsound` with unresolved design concerns, so waiting on it means waiting indefinitely.

Specialization is only needed if the fix is *a second, overlapping impl* for `Option<NonZero*>`. It isn't needed if the varying behaviour moves onto `Value` instead, which is your own trait.

## The change

`Value` gains one defaulted method:

```rust
fn niche() -> Option<Cow<'static, [u8]>> { None }
```

It returns an encoding of `Self`'s width that `as_bytes()` can never produce. `NonZero*` returns all-zeros. The single blanket `impl<T: Value> Value for Option<T>` consults it and, when present, uses that pattern to mean `None` instead of prefixing a discriminant byte.

One impl. No overlap. Stable Rust. And because the method is defaulted, every existing implementor — inside the crate or out — is untouched.

Result: `Option<NonZeroU32>` is 4 bytes rather than 5, matching the layout Rust already gives it in memory.

## Compatibility

`niche()` defaults to `None`, so every type that doesn't opt in keeps the discriminant encoding byte-for-byte. There's a test asserting exactly that for `Option<u32>` and `Option<&str>`.

`Option<NonZero*>` has never existed in a released redb — `NonZero*` didn't implement `Value` — so no database can hold the old five-byte form. The niche can be introduced now at zero migration cost. Later, it couldn't.

`backward_compatibility` passes unchanged.

## Ordering

`compare()` keeps `None` sorting below every `Some`, including below negative values of the signed types, and `min_encoded_key()` returns the niche. `separator()` short-circuits since a niche implies a fixed width.

## Tests

- `option_nonzero_is_niche_encoded` — width collapses 5 → 4, `None` is all-zeros, round-trips, ordering
- `option_nonzero_ordering_in_table` — `Option<NonZeroI32>` read back in order, `None` below `i32::MIN`
- `option_non_niche_encoding_is_unchanged` — `Option<u32>` and `Option<&str>` encode exactly as before

`cargo fmt --check` and `cargo clippy --all --all-targets` clean, `--no-default-features` builds, and basic_tests 128/128, integration_tests 110/110, backward_compatibility 8/8, multimap_tests 20/20 pass.

## If you'd rather not

The trait method is the part worth arguing about — it's public API surface on `Value`, and it puts a correctness obligation on implementors (return a pattern `as_bytes` can produce and values become indistinguishable from absence; the doc comment says so). If you'd prefer to keep `Value` minimal, I'm happy to make `niche()` a sealed/internal trait instead so only redb's own types can opt in, which covers `NonZero*` and costs outside implementors nothing.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -182,6 +182,23 @@ pub trait Value: Debug {
 
     /// Globally unique identifier for this type
     fn type_name() -> TypeName;
+
+    /// An encoding of `Self`'s width that `as_bytes()` can never produce, if one exists.
+    ///
+    /// Container types may use it to represent an absent value in place of a discriminant.
+    /// `Option<Self>` uses it to encode `None`, which makes `Option<Self>` the same width as
+    /// `Self` rather than one byte wider.
+    ///
+    /// Implementations that return `Some` must ensure the returned bytes are `fixed_width()`
+    /// long, that `fixed_width()` is not `None`, and that no value of `Self` encodes to them.
+    /// Returning a pattern that `as_bytes()` can produce will make values indistinguishable
+    /// from absence.
+    ///
+    /// The default implementation returns `None`, which leaves the discriminant encoding in
+    /// place.
+    fn niche() -> Option<Cow<'static, [u8]>> {
+        None
+    }
 }
 
 /// Implementing this trait indicates that the type can be mutated in-place as a &mut [u8].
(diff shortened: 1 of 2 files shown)


cberner:
Oh ya, this is a clever way to get around the specialization issue. Thanks! I'm merging a slightly different version, but it provides the same functionality. Lemme know if you run into any issues with it

PR #1412, opened 2026-08-27
Hold the multi-process byte ranges in the existing open modes

cberner:
First PR of the multi-process implementation on the revised `docs/design.md`: the existing open modes take the byte-range locks the protocol assigns them, unconditionally -- forward compatibility with the multi-process modes, whose feature gate arrives in a later PR. No new public API.

A writable open takes an exclusive byte-range lock over every offset the file can address, and a read-only open a shared one, so a single-process holder conflicts with every lock the multi-process modes will use -- their whole namespace lies inside the full range. Filesystems without range-lock support fall back to the whole-file lock with a warning, exactly as before.

Whether the whole-file lock std takes is a namespace of its own, so that both kinds must be held, is a question about the filesystem as much as the platform: NFS and SMB mounts emulate one kind of lock with the other, and the Apple platforms conflict outright. So it is asked at runtime, on the handle that has just taken the whole-file lock, by trying to lock one byte that the ranges leave out and nothing else ever covers -- a conflict there can only be that handle's own whole-file lock, showing through a shared table. One table means the ranges exclude whole-file holders too, and could not be taken beside the whole-file lock anyway, so the whole-file lock is dropped in their favor. A file open only for reading cannot take the exclusive lock the question needs, and takes none: the shared locks it does take coexist with the whole-file lock whichever the answer is.

The platform specifics live in a new `range_lock` module, as a `RangeLock` trait implemented for `File`: the lock methods, taking ranges, plus the namespace question. Every method defaults to reporting the locks unsupported and an empty impl covers the platforms without an implementation of their own, so the file backend names no platform at all -- it handles that one error kind. The locks are open file description locks (`F_OFD_SETLK`), never POSIX record locks -- a record lock belongs to the process and is dropped when any descriptor for the file closes anywhere in it, which a library cannot rule out. That is also why the `libc` dependency returns on Linux and the Apple platforms: std has no range-lock API, and `struct flock`'s layout differs between the two.

Tested against the protocol's actual offsets -- the header bytes, the mode bytes at `2^62`, a transaction byte, and the last addressable offset -- from second file descriptions standing in for other processes: a writable open holds them all exclusively and releases them at close, read-only opens share them, a held byte refuses a new open in both directions, the byte the namespace question needs stays free, and the whole-file lock older versions of redb take is refused while any of these opens is live.

The full multi-process implementation stacks on this in the next PR.

https://claude.ai/code/session_01SJFSfcturbVcnPqY5CNzQv

---
_Generated by [Claude Code](https://claude.ai/code/session_01SJFSfcturbVcnPqY5CNzQv)_

### DIFF
--- a/src/tree_store/page_store/file_backend/optimized.rs
+++ b/src/tree_store/page_store/file_backend/optimized.rs
@@ -1,6 +1,7 @@
 use crate::{DatabaseError, Result, StorageBackend};
 use std::fs::{File, TryLockError};
 use std::io;
+use std::ops::Range;
 
 #[cfg(feature = "logging")]
 use log::warn;
@@ -11,10 +12,34 @@ use std::os::unix::fs::FileExt;
 #[cfg(windows)]
 use std::os::windows::fs::FileExt;
 
+use super::range_lock::RangeLock;
+
+/// Below the coordination bytes of the multi-process locking protocol (docs/design.md) and
+/// above every offset it assigns, and left unlocked by every participant, so that a lock
+/// found there can only be a whole-file lock the range locks conflict with
+const NAMESPACE_PROBE_BYTE: u64 = (1 << 62) - 2;
+
+/// Every offset the protocol assigns, in the pieces the probe byte punctures it into
+const PROTOCOL_RANGES: [Range<u64>; 2] =
+    [0..NAMESPACE_PROBE_BYTE, NAMESPACE_PROBE_BYTE + 1..u64::MAX];
+
+/// A lock a failing open could not release, which is logged rather than reported: the open
+/// has an error of its own, and reporting this one in its place would turn a database that
+/// is merely open elsewhere into an I/O failure.
+fn report_failed_release(result: io::Result<()>) {
+    #[cfg(feature = "logging")]
+    if let Err(err) = result {
+        warn!("Failed to release a file lock after an open failed: {err}");
+    }
+    #[cfg(not(feature = "logging"))]
+    let _ = result;
+}
+
 /// Stores a database as a file on-disk.
 #[derive(Debug)]
 pub struct FileBackend {
     lock_supported: bool,
+    range_lock_supported: bool,
     file: File,
 }
 
@@ -25,6 +50,89 @@ impl FileBackend {
     }
 
     pub(crate) fn new_internal(file: File, read_only: bool) -> Result<Self, DatabaseError> {
+        let lock_supported = Self::try_whole_file_lock(&file, read_only)?;
+
+        // Where the whole-file lock conflicts with the range locks it excludes their holders
+        // as well, and no range is taken: on Windows, where the locks are mandatory, they
+        // would block this handle's own reads and writes. Where the answer belongs to the
+        // filesystem it is asked of the lock just taken, at the byte no range covers; a
+        // question the filesystem refuses is answered by holding both.
+        let conflicts = lock_supported
+            && match File::CONFLICTS_WITH_STD_FILE_LOCK {
+                Some(known) => known,
+                None => file
+                    .query_lock(NAMESPACE_PROBE_BYTE..NAMESPACE_PROBE_BYTE + 1)
+                    .unwrap_or(false),
+            };
+
+        let range_lock_supported = if conflicts {
+            false
+        } else {
+            match Self::try_lock_protocol_ranges(&file, read_only) {
+                Ok(true) => true,
+                // A multi-process handle has the database open
+                Ok(false) => {
+                    Self::release_whole_file_lock(&file, lock_supported);
+                    return Err(DatabaseError::DatabaseAlreadyOpen);
+                }
+                Err(err) if err.kind() == io::ErrorKind::Unsupported => {
+                    #[cfg(feature = "logging")]
+                    warn!(
+                        "Byte-range locks not supported by this filesystem. You must ensure that the database file is not opened for multi-process access"
+                    );
+
+                    false
+                }
+                Err(err) => {
+                    Self::release_whole_file_lock(&file, lock_supported);
+                    return Err(err.into());
+                }
+            }
+        };
+
+        Ok(Self {
+            lock_supported,
+            range_lock_supported,
+            file,
+        })
+    }
+
+    /// Exclusively for a writable open, shared for a read-only one, matching the whole-file
+    /// lock the same open would take. Pieces taken before one conflicts are released here:
+    /// the locks belong to the open file description, which a caller holding a `try_clone()`
+    /// keeps alive past the drop.
+    fn try_lock_protocol_ranges(file: &File, read_only: bool) -> io::Result<bool> {
+        for (taken, range) in PROTOCOL_RANGES.into_iter().enumerate() {
+            let acquired = if read_only {
+                file.try_lock_shared_range(range)
+            } else {
+                file.try_lock_range(range)
+            };
+            if !matches!(acquired, Ok(true)) {
+                Self::release_protocol_ranges(file, taken);
+                return acquired;
+            }
+        }
+        Ok(true)
+    }
+
+    /// The first `taken` ranges of an open that is failing.
+    fn release_protocol_ranges(file: &File, taken: usize) {
+        for range in PROTOCOL_RANGES.into_iter().take(taken) {
+            report_failed_release(file.unlock_range(range));
+        }
+    }
+
+    /// For the open paths that fail while holding it.
+    fn release_whole_file_lock(file: &File, held: bool) {
+        if held {
+            report_failed_release(file.unlock());
+        }
+    }
+
+    /// `Ok(false)` means the platform does not support file locks at all, which is reported
+    /// once rather than failing the open.
+    fn try_whole_file_lock(file: &File, read_only: bool) -> Result<bool, DatabaseError> {
         let result = if read_only {
             file.try_lock_shared()
         } else {
@@ -32,10 +140,7 @@ impl FileBackend {
         };
 
         match result {
-            Ok(()) => Ok(Self {
-                file,
-                lock_supported: true,
-            }),
+            Ok(()) => Ok(true),
             Err(TryLockError::WouldBlock) => Err(DatabaseError::DatabaseAlreadyOpen),
             Err(TryLockError::Error(err)) if err.kind() == io::ErrorKind::Unsupported => {
                 #[cfg(feature = "logging")]
@@ -43,10 +148,7 @@ impl FileBackend {
                     "File locks not supported on this platform. You must ensure that only a single process opens the database file, at a time"
                 );
 
-                Ok(Self {
-                    file,
-                    lock_supported: false,
-                })
+                Ok(false)
             }
             Err(TryLockError::Error(err)) => Err(err.into()),
         }
@@ -127,11 +229,19 @@ impl StorageBackend for FileBackend {
     }
 
     fn close(&self) -> Result<(), io::Error> {
+        // Every lock is released even where one fails: a lock left behind outlives this
+        // backend wherever the description is shared
+        let mut result = Ok(());
+        if self.range_lock_supported {
+            for range in PROTOCOL_RANGES {
+                result = result.and(self.file.unlock_range(range));
+            }
+        }
         if self.lock_supported {
-            self.file.unlock()?;
+            result = result.and(self.file.unlock());
         }
 
-        Ok(())
+        result
     }
 }
 
@@ -204,4 +314,172 @@ fn write_all_at(file: &File, mut buf: &[u8], mut offset: u64) -> io::Result<()> 
         }
     }
     Ok(())
+}
+
+#[cfg(all(test, any(target_os = "linux", target_vendor = "apple")))]
+mod range_lock_tests {
+    use super::{FileBackend, NAMESPACE_PROBE_BYTE, RangeLock};
+    use crate::StorageBackend;
+    use std::fs::{File, OpenOptions, TryLockError};
+    use std::path::Path;
+
+    // Offsets docs/design.md assigns: the header lock, the coordination bytes at BASE, the
+    // transaction range, and the last addressable byte
+    const BASE: u64 = 1 << 62;
+    const PROTOCOL_OFFSETS: [u64; 6] = [0, 319, BASE, BASE + 2, BASE + 1024 + 12345, (1 << 63) - 1];
+
+    fn reopen(path: &Path) -> File {
+        OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open(path)
+            .unwrap()
+    }
+
+    fn byte_is_free(file: &File, offset: u64, exclusive: bool) -> bool {
+        let byte = offset..offset + 1;
+        let acquired = if exclusive {
+            file.try_lock_range(byte.clone()).unwrap()
+        } else {
+            file.try_lock_shared_range(byte.clone()).unwrap()
+        };
+        if acquired {
+            file.unlock_range(byte).unwrap();
+        }
+        acquired
+    }
+
+    #[test]
+    fn a_writable_backend_holds_every_protocol_byte_exclusively() {
+        let tmpfile = crate::create_tempfile();
+        let backend = FileBackend::new_internal(reopen(tmpfile.path()), false).unwrap();
+
+        let observer = reopen(tmpfile.path());
+        for offset in PROTOCOL_OFFSETS {
+            assert!(!byte_is_free(&observer, offset, false), "offset {offset}");
+            assert!(!byte_is_free(&observer, offset, true), "offset {offset}");
+        }
+
+        // ... and releases them all at close
+        backend.close().unwrap();
+        for offset in PROTOCOL_OFFSETS {
+            assert!(byte_is_free(&observer, offset, true), "offset {offset}");
+        }
+    }
+
+    #[test]
+    fn read_only_backends_share_the_protocol_bytes_with_each_other() {
+        let tmpfile = crate::create_tempfile();
+        let first = FileBackend::new_internal(reopen(tmpfile.path()), true).unwrap();
+
+        let observer = reopen(tmpfile.path());
+        for offset in PROTOCOL_OFFSETS {
+            assert!(byte_is_free(&observer, offset, false), "offset {offset}");
+            assert!(!byte_is_free(&observer, offset, true), "offset {offset}");
+        }
+
+        let second = FileBackend::new_internal(reopen(tmpfile.path()), true).unwrap();
+        first.close().unwrap();
+        second.close().unwrap();
+    }
+
+    #[test]
+    fn a_held_protocol_byte_reads_as_already_open() {
+        let tmpfile = crate::create_tempfile();
+        let holder = reopen(tmpfile.path());
+        assert!(holder.try_lock_range(BASE..BASE + 1).unwrap());
+
+        // A held byte anywhere in the range stands in for a multi-process handle having
+        // the database open
+        assert!(matches!(
+            FileBackend::new_internal(reopen(tmpfile.path()), false),
+            Err(crate::DatabaseError::DatabaseAlreadyOpen)
+        ));
+        assert!(matches!(
+            FileBackend::new_internal(reopen(tmpfile.path()), true),
+            Err(crate::DatabaseError::DatabaseAlreadyOpen)
+        ));
+
+        holder.unlock_range(BASE..BASE + 1).unwrap();
+        let backend = FileBackend::new_internal(reopen(tmpfile.path()), false).unwrap();
+        backend.close().unwrap();
+    }
+
+    /// Dropping the file does not suffice: a caller holding a `try_clone()` of the file it
+    /// handed over keeps the open file description, and so the locks, alive
+    #[test]
+    fn a_refused_open_releases_what_it_took() {
+        let tmpfile = crate::create_tempfile();
+        let holder = reopen(tmpfile.path());
+        // In the second piece only, so that the first is taken before the conflict
+        assert!(holder.try_lock_range(BASE..BASE + 1).unwrap());
+
+        let file = reopen(tmpfile.path());
+        let kept_by_the_caller = file.try_clone().unwrap();
+        assert!(matches!(
+            FileBackend::new_internal(file, false),
+            Err(crate::DatabaseError::DatabaseAlreadyOpen)
+        ));
+
+        holder.unlock_range(BASE..BASE + 1).unwrap();
+        let observer = reopen(tmpfile.path());
+        assert!(byte_is_free(&observer, 0, true));
+        observer.try_lock().unwrap();
+        drop(kept_by_the_caller);
+    }
+
+    /// The question the namespace answer is asked with. An implementation blind to the locks
+    /// another handle holds would report every filesystem as keeping the two kinds apart.
+    #[test]
+    fn a_query_reports_the_locks_another_handle_holds() {
+        let tmpfile = crate::create_tempfile();
+        let byte = NAMESPACE_PROBE_BYTE..NAMESPACE_PROBE_BYTE + 1;
+        let asking = reopen(tmpfile.path());
+        assert!(!asking.query_lock(byte.clone()).unwrap());
+
+        let holder = reopen(tmpfile.path());
+        assert!(holder.try_lock_range(byte.clone()).unwrap());
+        assert!(asking.query_lock(byte.clone()).unwrap());
+
+        holder.unlock_range(byte.clone()).unwrap();
+        assert!(!asking.query_lock(byte).unwrap());
+    }
+
+    /// An ordinary Linux filesystem keeps the whole-file lock out of the range locks' table,
+    /// so an open holds both, and leaves the byte it asked at free for the next one to ask
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn an_ordinary_filesystem_answers_that_the_two_kinds_are_separate() {
+        assert!(File::CONFLICTS_WITH_STD_FILE_LOCK.is_none());
+        let tmpfile = crate::create_tempfile();
+        let backend = FileBackend::new_internal(reopen(tmpfile.path()), false).unwrap();
+
+        let observer = reopen(tmpfile.path());
+        assert!(matches!(observer.try_lock(), Err(TryLockError::WouldBlock)));
+        assert!(!byte_is_free(&observer, BASE, true));
+        assert!(byte_is_free(&observer, NAMESPACE_PROBE_BYTE, true));
+
+        backend.close().unwrap();
+    }
+
+    /// The whole-file lock is all an older version of redb takes, so an open must keep
+    /// excluding it however the namespace question is answered
+    #[test]
+    fn the_whole_file_lock_is_refused_while_a_backend_is_open() {
+        let tmpfile = crate::create_tempfile();
+        let observer = reopen(tmpfile.path());
+
+        let writable = FileBackend::new_internal(reopen(tmpfile.path()), false).unwrap();
+        assert!(matches!(
+            observer.try_lock_shared(),
+            Err(TryLockError::WouldBlock)
+        ));
+        writable.close().unwrap();
+
+        let reader = FileBackend::new_internal(reopen(tmpfile.path()), true).unwrap();
+        assert!(matches!(observer.try_lock(), Err(TryLockError::WouldBlock)));
+        reader.close().unwrap();
+
+        observer.try_lock().unwrap();
+    }
 }
--- a/src/tree_store/page_store/file_backend/range_lock.rs
+++ b/src/tree_store/page_store/file_backend/range_lock.rs
@@ -1,0 +1,155 @@
+use std::fs::File;
+use std::io;
+use std::ops::Range;
+#[cfg(any(target_os = "linux", target_vendor = "apple"))]
+use std::os::unix::io::AsRawFd;
+
+/// Byte-range locks, which the multi-process locking protocol (docs/design.md) coordinates
+/// through. The methods default to reporting the locks unsupported, which is what platforms
+/// with no implementation below get.
+pub(crate) trait RangeLock {
+    /// Whether these locks and [`File::lock`]'s conflict, which decides whether a handle
+    /// excluding both kinds of holder must take one of each. `None` where the answer belongs
+    /// to the filesystem rather than the platform -- Linux keeps the two apart, but a network
+    /// filesystem may emulate one with the other -- and [`RangeLock::query_lock`] asks.
+    const CONFLICTS_WITH_STD_FILE_LOCK: Option<bool> = None;
+
+    /// `Ok(false)` means a conflicting lock is held elsewhere. An end of `u64::MAX` covers
+    /// the file however it grows, past the last offset fcntl's signed arguments can express.
+    fn try_lock_range(&self, _range: Range<u64>) -> io::Result<bool> {
+        Err(unsupported())
+    }
+
+    fn try_lock_shared_range(&self, _range: Range<u64>) -> io::Result<bool> {
+        Err(unsupported())
+    }
+
+    fn unlock_range(&self, _range: Range<u64>) -> io::Result<()> {
+        Err(unsupported())
+    }
+
+    /// Whether an exclusive lock over the range would conflict with one already held. The
+    /// locks this description holds are not conflicts, but [`File::lock`]'s is wherever it
+    /// would in fact block a range lock taken here, which is what makes this the way to
+    /// answer [`RangeLock::CONFLICTS_WITH_STD_FILE_LOCK`] at runtime.
+    fn query_lock(&self, _range: Range<u64>) -> io::Result<bool> {
+        Err(unsupported())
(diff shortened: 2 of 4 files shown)


src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -11,10 +11,110 @@ use std::os::unix::fs::FileExt;
...
+// take, so the two kinds of lock cannot see each other; a single-process or read-only
+// handle must hold the entire byte range as well, so that it and a multi-process handle
+// refuse each other. Windows needs nothing extra: std's whole-file lock is itself a range
+// lock over every byte.
+//
+// Open file description locks, never POSIX record locks: a record lock belongs to the
+// process, and is released when any descriptor for the file is closed anywhere in it --
+// which a library embedded in an arbitrary program cannot rule out.
+#[cfg(any(target_os = "linux", target_vendor = "apple"))]
+mod range_lock {
```
cberner: Let's make the encapsulation better. Create a separate range_lock module with a trait that exposes the range lock methods and also a const indicating whether there are multiple namespaces. Then implement the trait for File. This way we keep all those platform specific stuff in that file, and the rest of the codebase can just interact with that trait and File.

cberner (commented):
The comments are too verbose. Also, this behavior should be gated by the "experimental-multiprocess" feature flag and the libc dependency should be optional based on it too.

And remove the changelog entry

cberner:
actually, sorry I changed my mind.  don't gate the range locking on a feature. we want this to always be on so that it's forward compatible with the MP feature. we'll introduce the feature gate in a later PR

src/tree_store/page_store/file_backend/optimized.rs
cberner: Done in 9d91b49: range_lock is now its own module exposing a `RangeLock` trait implemented for `File` -- try/blocking/unlock methods plus a `MULTIPLE_NAMESPACES` const saying whether whole-file and range locks are separate namespaces that must both be held (Linux) or share one table where the range lock alone excludes both kinds of holder (macOS). The backend interacts only with the trait; the Windows implementation arrives in #1413, since std's whole-file lock already covers every byte here. The rest of the review is in the same commit: the behavior is gated behind `experimental-multiprocess` with `libc` optional on it, the comments are trimmed, and the changelog entry is removed.

---

cberner:
Ungated in 62609b7: the range locking is unconditional again -- platform cfgs only, `libc` back to a plain dependency, no feature in this PR. The `experimental-multiprocess` feature is declared in #1413 instead, where it gates only the multi-process modes themselves (rebased there as 1664c82).

---

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Open file description locks, never POSIX record locks: a record lock belongs to the
+// process, and is released when any descriptor for the file is closed anywhere in it,
+// which a library embedded in an arbitrary program cannot rule out.
+impl RangeLock for File {
+    // flock() and fcntl() locks cannot see each other on Linux. On the Apple platforms
+    // they share one lock table and conflict with each other, even through one file
+    // description, so a range lock alone excludes holders of either kind.
+    const MULTIPLE_NAMESPACES: bool = cfg!(target_os = "linux");
```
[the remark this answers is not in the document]
cberner: Half right, and handled in c928160. The NFS half of the claim does not hold on any kernel of the last decade: flock(2) documents the emulation ("fcntl(2) and flock() locks do interact with one another over NFS"), but the modern NFS client keys its lock state by owner alone, and a flock lock and an OFD lock on the same open file description carry the same owner -- the struct file (see __nfs4_find_lock_state in fs/nfs/nfs4state.c: "the second refers to a flock/OFD owner (struct file*)"). Same owner means the overlapping locks merge rather than conflict, so both are held and the open succeeds.

SMB mounts (kernel 5.5+, where flock is emulated per-handle with no-overlap semantics) can genuinely self-conflict, so the general point stands, and it is now handled without any filesystem-specific tables: a writable open holding the whole-file lock probes one byte that nothing but a whole-range lock ever covers. A conflict there can only be the handle's own emulated lock -- and a shared table also means the whole-file lock by itself already excludes every byte-range holder, so the open keeps it alone and takes no range lock. On separate-namespace filesystems the probe is invisible and nothing changes.

---

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,89 @@
+use std::fs::File;
+use std::io;
+use std::os::unix::io::AsRawFd;
+
+/// Byte-range locks on the database file, which the multi-process locking protocol
+/// (docs/design.md) coordinates through.
+pub(crate) trait RangeLock {
+    /// Whether whole-file locks occupy a separate namespace that range locks cannot see,
+    /// so that a handle excluding both kinds of holder must take one lock of each.
+    const MULTIPLE_NAMESPACES: bool;
```
cberner: Change this to a method, since it looks like we need to detect this at runtime on filesystems like NFS. I guess it should take an offset to a byte that is expected to be unlocked on platforms that have separate namespaces, or something like that

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Tries to lock `len` bytes from `start`, without blocking. A `len` of zero locks
+    /// everything from `start` on, and is the only way to reach the last addressable
+    /// byte, since the offsets fcntl takes are signed. `Ok(false)` means a conflicting
+    /// lock is held elsewhere.
+    fn try_lock_range(&self, start: u64, len: u64, exclusive: bool) -> io::Result<bool>;
```
cberner: Take a range instead of start and len. Have separate methods for try_lock_range() and try_lock_shared_range().

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    fn unlock_range(&self, start: u64, len: u64) -> io::Result<()>;
```
cberner: take a range

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Whether the error, from one of the lock methods, means the filesystem does not
+    /// support range locks.
+    fn lock_unsupported(err: &io::Error) -> bool;
```
cberner: No, instead rewrite the error kind to be `Unsupported`

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -11,10 +11,15 @@ use std::os::unix::fs::FileExt;
 #[cfg(windows)]
 use std::os::windows::fs::FileExt;
 
+#[cfg(any(target_os = "linux", target_vendor = "apple"))]
```
cberner: I don't want to have all these cfg in this file. Try to limit them to the range_lock file. Instead add default implementations that return something like Unsupported, and then this code should just handle that case.

src/tree_store/page_store/file_backend/range_lock.rs
cberner: Done in b9cec36: it is now `multiple_lock_namespaces(unlocked_offset)`, asked at runtime on the handle that has just taken the whole-file lock, and the separate `whole_file_lock_covers_ranges` helper in the backend is gone. The offset is a byte the ordinary opens now leave out of their ranges, so a conflict there can only be this handle's own whole-file lock showing through a shared table; one table means the ranges exclude whole-file holders too, so the whole-file lock is dropped in their favor, and separate namespaces mean both are held.

Two consequences worth flagging. The ranges are punctured at that byte, which is why nothing else can answer the probe -- there is a test that it stays free while a backend holds everything else. And read-only opens ask the question too, which they did not before: a file open only for reading cannot take the exclusive lock the question needs (EBADF), and it takes none either, so that answers "separate" -- the shared locks it does take coexist with the whole-file lock whichever the answer really is.

#1413 is rebased on top of this.

---

src/tree_store/page_store/file_backend/range_lock.rs
cberner: Done: `try_lock_range(range)` and `try_lock_shared_range(range)`, both taking a `Range<u64>`. An end of `u64::MAX` means the end of the file however it grows, which is the only way to reach the last addressable byte, since the offsets fcntl takes are signed; #1413's Windows implementation reads it as a length of `u64::MAX - start`.

---

src/tree_store/page_store/file_backend/range_lock.rs
cberner: Done -- `unlock_range(range)`, and the blocking locks and the probe that #1413 adds take ranges too, so the whole trait speaks one vocabulary.

---

src/tree_store/page_store/file_backend/range_lock.rs
cberner: Done: `lock_unsupported` is gone, and the implementation rewrites the kind instead -- EINVAL, ENOTSUP and EOPNOTSUPP from any of the lock calls come back as `io::ErrorKind::Unsupported`, with the original error kept as the source so its message survives. The backend matches on `err.kind()`.

---

src/tree_store/page_store/file_backend/optimized.rs
cberner: Done: every trait method has a default returning `Unsupported`, and an empty `impl RangeLock for File {}` covers the platforms with no implementation of their own, so the locking code in optimized.rs has no platform cfgs left -- one `new_internal`, one `close`, one struct, and the `Unsupported` kind handled where the ranges are taken (and again where the namespace question is asked, which just leaves the reporting to the acquisition). That also collapsed the two `new_internal` bodies into one.

The cfgs still in the file are the pre-existing positional-I/O ones: the `FileExt` imports and the unix/wasi/windows `read` and `write` bodies.

---

docs/design.md
```diff
@@ -487,18 +487,28 @@ The following concurrency modes are supported, and their locking protocol is des
...
+namespaces on some platforms and one lock table on others -- and filesystems that emulate one kind
+of lock with the other, such as NFS and SMB mounts, make it a question about the mount rather than
+about the platform. So it is answered at runtime: holding the whole-file lock, the process tries to
+lock one byte that the ranges always leave unlocked, where a conflict can only be its own
+whole-file lock showing through a shared table. One table means the ranges exclude whole-file
+holders as well, so the whole-file lock is dropped in their favor; separate namespaces mean both
+are held. A file opened only for reading cannot take the exclusive lock the question needs, and
+takes none either: the shared locks it does take coexist with the whole-file lock whichever the
+answer is.
+
```
cberner: Remove this

cberner: In fact, revert all your edits to this file. I don't like them

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,150 @@
+use std::fs::File;
+use std::io;
+use std::ops::Range;
+#[cfg(any(target_os = "linux", target_vendor = "apple"))]
+use std::os::unix::io::AsRawFd;
+
+/// Byte-range locks on the database file, which the multi-process locking protocol
+/// (docs/design.md) coordinates through. Every method defaults to reporting the locks
+/// unsupported, which is what platforms with no implementation below get.
```
cberner: Don't document obvious stuff. Massively trim back your comments. READ AGENTS.md ESPECIALLY THE COMMENT STYLE GUIDE!!!

docs/design.md
cberner: Reverted -- docs/design.md is untouched in this PR now, and I dropped my edits to it from #1413 as well.

---

src/tree_store/page_store/file_backend/range_lock.rs
cberner: Trimmed in f797567. range_lock.rs keeps only what is not visible from the code: OFD locks rather than POSIX record locks, what the namespace question needs of its offset, the `u64::MAX` end bound, the EBADF case, and why `struct flock` is zeroed rather than filled in. The restatements of method names are gone. optimized.rs got the same pass, and so did #1413's copies of both files and its multiprocess.rs.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +39,91 @@ impl FileBackend {
...
+    /// locked when a later one conflicts is released with the file, which the caller drops.
+    fn try_lock_protocol_ranges(file: &File, read_only: bool) -> io::Result<bool> {
+        for range in PROTOCOL_RANGES {
+            let acquired = if read_only {
+                file.try_lock_shared_range(range)?
+            } else {
+                file.try_lock_range(range)?
+            };
+            if !acquired {
+                return Ok(false);
```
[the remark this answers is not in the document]
cberner: Confirmed and fixed in f797567. `try_clone()` dups the descriptor, so both handles share one open file description and the locks live on it: dropping the failed open's `File` releases nothing while the caller keeps its clone. The acquisition loop now releases the pieces it took before the conflict, and the failure paths that hold the whole-file lock release that too.

`a_refused_open_releases_what_it_took` covers it, built around exactly this: the caller keeps a `try_clone()` alive across the refused open, then a third description checks that the first piece and the whole-file lock are both free. It fails without the fix.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +39,114 @@ impl FileBackend {
...
+    /// a read-only one, matching the whole-file lock the same open would take. Pieces taken
+    /// before one conflicts are released, since dropping this file need not release them:
+    /// the locks belong to its open file description, which a caller that kept a
+    /// `try_clone()` of the file still holds.
+    fn try_lock_protocol_ranges(file: &File, read_only: bool) -> io::Result<bool> {
+        for (taken, range) in PROTOCOL_RANGES.into_iter().enumerate() {
+            let acquired = if read_only {
+                file.try_lock_shared_range(range)?
+            } else {
+                file.try_lock_range(range)?
```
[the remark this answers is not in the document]
cberner: Right, the `?` skipped the cleanup. Fixed in 9c66783: the loop matches on the result and releases the earlier pieces for anything that is not `Ok(true)`, then returns that result unchanged -- `Ok(false)` and every `Err` alike.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -127,6 +221,11 @@ impl StorageBackend for FileBackend {
     }
 
     fn close(&self) -> Result<(), io::Error> {
+        if self.range_lock_supported {
+            for range in PROTOCOL_RANGES {
+                self.file.unlock_range(range)?;
```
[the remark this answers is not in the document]
cberner: Fixed in 9c66783: `close()` now attempts every release and keeps the first error, rather than returning on the first failure.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +39,114 @@ impl FileBackend {
...
+            }
+        }
+
+        // A conflict means a multi-process handle has the database open, or another
+        // ordinary handle does where these ranges stand in for the whole-file lock
+        let range_lock_supported = match Self::try_lock_protocol_ranges(&file, read_only) {
+            Ok(true) => true,
+            Ok(false) => {
+                Self::release_whole_file_lock(&file, lock_supported);
+                return Err(DatabaseError::DatabaseAlreadyOpen);
```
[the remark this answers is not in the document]
cberner: Not taking this one: the changelog entry was in an earlier revision of this PR and the review asked for it to be removed, so it stays out. The user-facing note belongs with the feature that makes the locks reachable, in #1413.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +38,117 @@ impl FileBackend {
...
+                return acquired;
+            }
+        }
+        Ok(true)
+    }
+
+    /// The first `taken` ranges of an open that is failing.
+    fn release_protocol_ranges(file: &File, taken: usize) {
+        for range in PROTOCOL_RANGES.into_iter().take(taken) {
+            let _ = file.unlock_range(range);
```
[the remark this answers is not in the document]
cberner: Half taken, in 51cf91c. Not preserved as the returned error: these paths already have one, and it is the one the caller needs -- returning a cleanup failure in its place would turn a database that is merely open elsewhere into an I/O error, and `DatabaseAlreadyOpen` is what callers match on to back off and retry. `close()` is the other case and does propagate, since nothing competes with it there.

What the report is right about is that discarding it silently loses the one thing worth knowing -- that a lock outlived the failed open, which only matters where the caller kept a `try_clone()` and goes on holding the description. Both helpers now route through `report_failed_release`, which logs it under the `logging` feature; that is also where the later "why is this database still refusing me" answer comes from.

---

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,155 @@
...
+                self.unlock_range(byte)?;
+                Ok(true)
+            }
+            Ok(false) => Ok(false),
+            // A file open for reading only cannot take an exclusive lock at all, so it asks
+            // by probing: a shared table reports this handle's own whole-file lock
+            Err(err) if err.raw_os_error() == Some(libc::EBADF) => range_is_free(self, byte),
+            Err(err) => Err(err),
+        }
+    }
```
cberner: I think the multiple_lock_namespaces() function is the wrong abstraction. Let's replace it with const: CONFLICTS_WITH_STD_FILE_LOCK: Option<bool> = None. None means unknown and that's the value Linux will use, Mac and Windows will use Some(true) meaning there is only a single lock namespace. Then on the optimized.rs side we will do the multiple namespace checking if it is None. To avoid any re-locking semantics that might be bad, I think we want to have two separate OFDs: one that does the File.lock() (i.e. flock()) and one that tries the range_lock.

and if you have a much better name for the constant that's fine.

cberner (commented):
There are kind of a lot of tests. Check that they are all testing important functionality and not optimizing code coverage for its own sake. Also do a polish and trimming pass on the comments.

src/tree_store/page_store/file_backend/range_lock.rs
cberner: Done in 1981cc4, keeping your name: `CONFLICTS_WITH_STD_FILE_LOCK: Option<bool>`, `None` by default and on Linux, `Some(true)` on the Apple platforms and Windows. The backend asks only where it is `None`, and where it is `Some(true)` no whole-file lock is taken at all -- the ranges are the exclusion, since either kind excludes both.

The second description turned out to matter more than as hygiene. Asking through the handle that holds the whole-file lock gives the wrong answer on the filesystem this exists for: the NFS client emulates flock() with a whole-file POSIX lock whose owner is the same `struct file *` an OFD lock uses, so for one description the two merge instead of conflicting and the handle concludes "separate" -- exactly backwards. Through a second description they are different owners and conflict as they should. So the check opens one (`independent_handle()`: `/proc/self/fd/N` on Linux, `Unsupported` elsewhere, which is only reachable where the const is `None`) and tries the exclusive range lock through it while this handle holds the whole-file lock. Anything unanswerable -- no second description, no write access on it -- reads as separate, which is the answer that keeps both kinds held.

Two things fell out. The EBADF path is gone: it existed because a file open for reading only cannot take the exclusive lock the old question needed, and the platforms where a read-only open must know the answer now answer statically. And last round's macOS failure cannot recur, since there is no whole-file lock left there to shadow the byte.

On the tests: dropped the read-only-file one, whose subject was that EBADF path; merged `a_read_only_backend_shares_the_protocol_bytes` into `read_only_backends_share_the_protocol_bytes_with_each_other`, since the second reader's open is the assertion that matters and the first test was its setup; and added one for the new mechanism. Six now, each for a distinct thing: what a writable open holds and releases, what read-only opens hold and share, an outside byte refusing an open, the whole-file lock refusing an older binary, a failed open releasing what it took, and the namespace question itself. Comments trimmed again throughout, and #1413 is rebased on it.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +50,125 @@ impl FileBackend {
     }
 
     pub(crate) fn new_internal(file: File, read_only: bool) -> Result<Self, DatabaseError> {
+        // Taken first: the question below needs a whole-file lock to look for
+        let mut lock_supported = Self::try_whole_file_lock(&file, read_only)?;
+
+        // One table: the ranges below exclude whole-file holders too, and could not be taken
+        // beside the whole-file lock anyway, so they stand in for it and it is dropped
+        if lock_supported && Self::one_lock_namespace(&file) {
```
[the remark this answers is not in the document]
cberner: Confirmed, and fixed in 82d57dd. The hole is real: any conflict at that byte read as "one table", and another opener's momentary exclusive lock there is such a conflict, so a read-only open could drop the whole-file lock an older writer is excluded by.

Read-only opens no longer ask at all, which is the shape you suggest and removes the race rather than narrowing it: a read-only open holds shared locks of both kinds, and those coexist whichever the answer is, so it never needed to know. That leaves writable opens as the only askers -- and a writable open holds the whole-file lock *exclusively* while it asks, so no second asker can exist and the only lock that can be at that byte is the one it is looking for.

`a_read_only_open_keeps_the_whole_file_lock_while_another_asks` covers it deterministically: an outside handle holds the probe byte exclusively, a read-only backend opens beside it, and an older binary's writer must still be refused. It fails without the fix, on exactly that last assertion.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +50,125 @@ impl FileBackend {
...
+            Ok(true) => true,
+            Ok(false) => {
+                Self::release_whole_file_lock(&file, lock_supported);
+                return Err(DatabaseError::DatabaseAlreadyOpen);
+            }
+            Err(err) if err.kind() == io::ErrorKind::Unsupported => {
+                #[cfg(feature = "logging")]
+                warn!(
+                    "Byte-range locks not supported by this filesystem. You must ensure that the database file is not opened for multi-process access"
+                );
```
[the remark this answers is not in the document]
cberner: Right on both counts, fixed in 82d57dd: Windows now declares `CONFLICTS_WITH_STD_FILE_LOCK = Some(true)` here, which is true whether or not the range locks are implemented, and the warning is skipped where the const says so -- there the whole-file lock excludes range-lock holders as well, so nothing is missing to warn about.

The const also spares Windows the whole-file lock's take-and-drop: an open that knows the two share a namespace does not take it at first, and reaches for it only if the ranges turn out to be unsupported, which on Windows is until #1413 adds the implementation.

---

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -0,0 +1,154 @@
...
+/// A second open file description on the same file, for the question a handle cannot answer
+/// about itself: locks taken through one description convert rather than conflict.
+/// `Unsupported` where a file cannot be reopened from its descriptor alone.
+pub(crate) fn independent_handle(file: &File) -> io::Result<File> {
+    #[cfg(target_os = "linux")]
+    {
+        File::options()
+            .read(true)
+            .write(true)
+            .open(format!("/proc/self/fd/{}", file.as_raw_fd()))
```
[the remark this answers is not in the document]
cberner: Right: an SMB mount self-conflicts, so with the whole-file lock still held the ranges fail against it, and reading an unanswerable question as "separate" turned that into `DatabaseAlreadyOpen` with nobody else there.

Fixed in e1967a8 by letting the acquisition answer what the probe could not. `one_lock_namespace()` now returns `Option<bool>` -- `None` for a question it could not ask -- and where the answer is unknown, a conflict on the ranges is treated as possibly this handle's own whole-file lock: it drops that lock and tries the ranges once more. A conflict that outlives the whole-file lock belongs to someone else and refuses the open as before, so the fallback cannot admit a second opener. On a mount where the two really are separate, the retry just fails the same way the first attempt did.

`an_unanswered_question_retries_the_ranges_without_the_whole_file_lock` drives the branch that Linux CI can otherwise never reach: an unanswered question, a conflict held by someone else, and the whole-file lock dropped on the way to refusing the open.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +50,169 @@ impl FileBackend {
...
+    ) -> io::Result<bool> {
+        if Self::try_lock_protocol_ranges(file, read_only)? {
+            return Ok(true);
+        }
+        if answered || !*whole_file_lock {
+            return Ok(false);
+        }
+        file.unlock()?;
+        *whole_file_lock = false;
+        Self::try_lock_protocol_ranges(file, read_only)
```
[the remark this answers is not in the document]
cberner: Confirmed, and fixed in 49b91e3. The retry dropped the whole-file lock on a guess, and where the namespaces really are separate that lock is the only thing an older binary respects, so the window you describe is real.

It is now dropped only where it is demonstrably the blocker: on a conflict, the open tries a range of its own at the byte no protocol range covers and nothing else ever locks. A conflict there is this handle's own whole-file lock showing through a shared table, which is exactly the filesystem where dropping it is safe -- the ranges taken instead exclude the older binaries it was held against, and an older binary that slipped into the window would conflict with those ranges rather than be invisible to them. Where that byte is free the conflict belongs to someone else, and the open is refused with the whole-file lock still held.

That also covers the NFS shape, where the two merge for one description instead of conflicting: the byte comes back free, so the original conflict really was another holder's and the refusal is right.

The test from the last round now asserts the other half of it -- `an_unanswered_question_keeps_the_whole_file_lock_for_another_holders_conflict`: unanswered question, someone else's conflict, and afterwards both the whole-file lock still held and an older binary's writer still refused.

---

src/tree_store/page_store/file_backend/range_lock.rs
cberner: I don't like this. It relies on procfs, and who knows if that is enabled or not. However, I think we can take a different approach to detecting the lock namespaces:
1) acquire the whole-file lock (flock())
2) call F_GETLK on the probe byte
3) determine if they collided

Research that carefully to see if it correctly will determine the behavior on NFS...etc.

This page (https://manpages.ubuntu.com/manpages/stonking/man2/F_OFD_SETLK.2const.html) implies yes: "Conflicting lock combinations (i.e., a read lock and a write lock or two write locks) where one lock is an open file description lock and the other is a traditional record lock conflict even when they are acquired by the same process on the same file descriptor."

If that works then add a query_lock() to RangeLock

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +50,190 @@ impl FileBackend {
...
+                    lock_supported = Self::try_whole_file_lock(&file, read_only)?;
+                }
+
+                false
+            }
+            Err(err) => {
+                Self::release_whole_file_lock(&file, lock_supported);
+                return Err(err.into());
+            }
+        };
```
cberner: This is surprisingly complicated. I haven't read it carefully yet, but can't it be structured more like:
```pseudocode
let lock_supported = try_whole_file_lock()?;
let conflicts = if CONFLICTS.is_none() {
  // probe the byte
  // true or false as appropriate
} else let Some(conflicts) = CONFLICTS {
  conflicts
};
let range_lock_acquired = if !conflicts {
  let result = try_protocol_range_lock();
  if result.is_locked() {
     // someone else has a multi-process lock
     whole_file_unlock();
     return Err(DatabaseAlreadyOpen);
  }
  true
} else {
  false
};
```

src/tree_store/page_store/file_backend/range_lock.rs
cberner: Researched it, and it works. Done in c7b4d71: `independent_handle` and the procfs reopen are gone, replaced by `query_lock(range)` on `RangeLock` -- `F_OFD_GETLK` on Unix, and on Windows (in #1413) an acquire-and-release, since there is no query operation there.

One correction to the man page quote, because it matters for why this works. That sentence is about OFD locks vs *traditional POSIX record locks*; it does not cover flock(). flock() locks live in a different list on the inode (`flc_flock`) from POSIX/OFD locks (`flc_posix`), and `posix_test_lock()` only walks the latter -- so on a local Linux filesystem `F_OFD_GETLK` structurally cannot see an flock. That happens to be the right answer there, but it is not the reason the scheme works.

The property that actually holds is narrower and better: **the query reports the whole-file lock exactly where that lock would in fact block a range lock taken through the same description.** That is the question the backend needs answered, and it is answered correctly in all three cases:

- **Local Linux.** Separate lists, so neither `F_OFD_GETLK` nor `F_OFD_SETLK` sees the flock. Query says "no conflict", and there is none: both are held.
- **NFS.** `nfs_flock()` routes into `do_setlk()`, which sends the lock to the server and records it locally through `locks_lock_file_wait()` -- as an `FL_FLOCK` lock, so the local half of `do_getlk()` (`posix_test_lock`) misses it. The remote half asks the server, where flock and OFD locks through one description share a lock owner: `__nfs4_find_lock_state` in `fs/nfs/nfs4state.c` -- "the first refers to a Posix-lock owner (`current->files`) and the second refers to a flock/OFD owner (`struct file*`)". Same owner, no conflict reported -- and none exists, so both are held here too. Another process's lock of either kind still conflicts with ours, which is all the exclusion needs.
- **SMB/CIFS**, the case that motivated this. `cifs_flock()` goes through the same `cifs_setlk()` as `cifs_lock()` and records the emulated lock in the per-fid byte-range list with `FL_FLOCK` in `li->flags`. `cifs_getlk()` -> `cifs_lock_test()` -> `cifs_find_fid_lock_conflict()`, whose same-fid exemption requires `FL_OFDLCK` on *both* the query and the recorded lock (`fs/smb/client/file.c`), so an emulated flock is reported as a conflict. And it is the same predicate `cifs_lock_add_if()` uses to grant a lock, so the query and the acquisition cannot disagree here -- which is what lets the retry go away.

macOS and Windows keep `Some(true)` and never ask. Where the answer is "no conflict" holding both is safe; where it is "conflict" the whole-file lock alone already excludes range-lock holders.

Two things fell out. Read-only opens can ask now -- the question is passive, so last round's race between two openers cannot exist -- and the retry in `take_protocol_ranges` plus `whole_file_lock_blocks_ranges` are both gone, since the acquisition no longer has to answer what the probe could not.

---

src/tree_store/page_store/file_backend/optimized.rs
cberner: Yes -- restructured to that shape in c7b4d71, which the `query_lock()` change above is what made possible. `new_internal` is now:

```rust
let lock_supported = Self::try_whole_file_lock(&file, read_only)?;

let conflicts = lock_supported
    && match File::CONFLICTS_WITH_STD_FILE_LOCK {
        Some(known) => known,
        None => file
            .query_lock(NAMESPACE_PROBE_BYTE..NAMESPACE_PROBE_BYTE + 1)
            .unwrap_or(false),
    };

let range_lock_supported = if conflicts {
    false
} else {
    match Self::try_lock_protocol_ranges(&file, read_only) { ... }
};
```

The `Unsupported` arm is the only thing beyond your sketch: a filesystem with no byte-range locks is left with the whole-file lock and a warning. That warning no longer needs its Windows suppression, since it is only reachable where the whole-file lock does *not* stand in for the ranges.

Two consequences worth flagging.

Windows and macOS now keep the whole-file lock and take no range at all, which is what your `else { false }` branch says. On Windows that makes this PR a no-op -- the lock std already takes is a `LockFileEx` over every byte -- and it removes the mandatory-lock hazard of a handle holding both.

That does not survive into #1413 unchanged, though, for read-only opens: a shared whole-file lock neither conflicts with the shared mode bytes a multi-writer cohort holds nor can be told apart from them when the open probes for one, and on Windows it would block the cohort's I/O for as long as it was held. So #1413 (ebf1a4f) adds one exception -- a read-only open holds the ranges rather than the whole-file lock wherever the two conflict -- and nothing else about the structure changes. A writable open's lock is exclusive and excludes a cohort by itself, so it needs no exception.

Tests: dropped the two that covered mechanisms that no longer exist (the retry, and read-only opens skipping the question), and added one for `query_lock` itself -- an implementation blind to other handles' locks would report every filesystem as keeping the two kinds apart, which is the failure mode that matters. Seven now.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -25,28 +50,105 @@ impl FileBackend {
...
+                // A multi-process handle has the database open
+                Ok(false) => {
+                    Self::release_whole_file_lock(&file, lock_supported);
+                    return Err(DatabaseError::DatabaseAlreadyOpen);
+                }
+                Err(err) if err.kind() == io::ErrorKind::Unsupported => {
+                    #[cfg(feature = "logging")]
+                    warn!(
+                        "Byte-range locks not supported by this filesystem. You must ensure that the database file is not opened for multi-process access"
+                    );
```
cberner: No, don't log a warning. If range locks are not supported then there will be no multi-process access, because when we implement that it will fail to open.

PR #1414, opened 2026-08-27
Publish the grown layout when a transaction is aborted

jaideeppyne:
Growing the database file extends it and updates the layout in memory, but only a header write publishes that layout. A commit writes the header itself, so the two stay in step. An abort does not, and rolling the transaction's allocations back does not shrink the file, so the header on disk is left describing fewer regions than the file holds.

Reloading the header rebuilds the layout from the file length and reports the database as unclean when the stored counts disagree, so `check_integrity()` returned `Ok(false)` — "failed but was repaired" — for a database that was never damaged, after repairing it. Any write transaction that grew the file and was then aborted or dropped triggered it, on a default `Database::create()`:

```rust
let mut db = Database::create(path).unwrap();
// ... one small committed row; check_integrity() == true
let txn = db.begin_write().unwrap();
{ let mut t = txn.open_table(TABLE).unwrap();
  for k in 0..3000u64 { t.insert(&k, vec![0xAA; 2000].as_slice()).unwrap(); } }
txn.abort().unwrap();
assert!(db.check_integrity().unwrap()); // fails before this change
```

Abort now writes the header, as a commit would. A pending `Durability::None` commit is left alone: its slot names pages that have not been flushed, so the header must not reach the disk ahead of them, and a reload verifies the live state it describes rather than the stored layout.

`tests/check_integrity_abort.rs` covers both the explicit `abort()` and the drop-abort path; both fail on master. Found with a model-based differential fuzzer using variable-width keys, which the in-tree harness does not exercise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,9 @@
   with such keys use slightly less space, and lookups are faster.
 * Fix a crash shortly after a commit being able to silently roll that commit back during
   recovery, if `check_integrity()` had previously repaired the database.
+* Fix `check_integrity()` reporting a healthy database as corrupt, and repairing it, after a
+  write transaction that grew the database file was aborted or dropped. The abort left the
+  header on disk describing fewer regions than the file held, which the check read as damage.
 * Fix iterators silently omitting data when iteration continues after an error. An iterator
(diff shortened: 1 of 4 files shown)


cberner:
Thanks for reporting this! I fixed it in a slightly different way. Please let me know if you still have issues with the code that's on master

PR #1417, opened 2026-08-28
Record the collection horizon in the commit slots

cberner:
Superseded by b28523a, which takes the simpler approach: no persistent horizon at all, and a process invalidates its cache when a transaction begins and observes a new committed transaction id.

Closing rather than reworking -- there is nothing left of this PR once the field goes.

The Codex finding that led here is on the thread below: the horizon claimed to bound page reuse, but the pages a commit frees from the system tree are freed outright once it is done, so they escape it and a peer could serve a stale copy. Keeping the field would have meant routing those through the freed table, and the invariant would have stayed open-ended -- correct only as long as every future path that frees a page respects it, which had already been broken twice.

If the coarser invalidation ever proves too expensive, this field is a format-compatible addition: it sits in slot padding inside the checksummed region, so it can be added back without a version bump.

### DIFF
--- a/src/transactions.rs
+++ b/src/transactions.rs
@@ -2040,12 +2040,16 @@ impl WriteTransaction {
         };
 
         let page_allocator = self.page_allocator();
+        // process_freed_pages ran through free_until_transaction, so reuse reaches the id below it
         self.mem.commit(
             user_root,
             system_root,
             self.transaction_id,
             self.two_phase_commit,
             self.shrink_policy,
+            Some(TransactionId::new(
+                free_until_transaction.raw_id().saturating_sub(1),
+            )),
         )?;
         // All of this transaction's allocations are durable; discard the per-txn tracker.
         let _ = page_allocator.take_allocated_since_commit();
--- a/src/tree_store/page_store/header.rs
+++ b/src/tree_store/page_store/header.rs
@@ -73,6 +73,7 @@ const USER_ROOT_OFFSET: usize = _UNUSED_OFFSET + size_of::<u8>() + PADDING;
 const SYSTEM_ROOT_OFFSET: usize = USER_ROOT_OFFSET + BtreeHeader::serialized_size();
 const _UNUSED2_OFFSET: usize = SYSTEM_ROOT_OFFSET + BtreeHeader::serialized_size();
 const TRANSACTION_ID_OFFSET: usize = _UNUSED2_OFFSET + BtreeHeader::serialized_size();
+const COLLECTION_HORIZON_OFFSET: usize = TRANSACTION_ID_OFFSET - size_of::<u64>();
 const TRANSACTION_LAST_FIELD: usize = TRANSACTION_ID_OFFSET + size_of::<u64>();
 
 const SLOT_CHECKSUM_OFFSET: usize = TRANSACTION_SIZE - size_of::<Checksum>();
@@ -405,17 +406,20 @@ impl DatabaseHeader {
         &self.transaction_slots[self.primary_slot ^ 1]
     }
 
-    // Overwrite the secondary slot with a newly committed transaction.
+    // Reuse an earlier commit made visible cannot be taken back, so the horizon only advances
     pub(super) fn write_secondary_slot(
         &mut self,
         transaction_id: TransactionId,
         user_root: Option<BtreeHeader>,
         system_root: Option<BtreeHeader>,
+        collection_horizon: Option<TransactionId>,
     ) {
+        let floor = self.transaction_slots[self.primary_slot].collection_horizon;
         let slot = &mut self.transaction_slots[self.primary_slot ^ 1];
         slot.transaction_id = transaction_id;
         slot.user_root = user_root;
         slot.system_root = system_root;
+        slot.collection_horizon = floor.max(collection_horizon.unwrap_or(floor));
         slot.corrupt_bytes = None;
     }
 
@@ -461,6 +465,9 @@ pub(super) struct TransactionHeader {
     pub(super) user_root: Option<BtreeHeader>,
     pub(super) system_root: Option<BtreeHeader>,
     pub(super) transaction_id: TransactionId,
+    // The newest transaction whose freed pages a commit's reclamation has processed for reuse.
+    // Bounds that alone: the pages a commit frees from the system tree once it is done escape it
+    pub(super) collection_horizon: TransactionId,
     // When present, the original on-disk bytes of a slot whose checksum did not verify. They are
     // written back verbatim, so a slot that failed verification keeps its invalid checksum instead
     // of being re-serialized as if it were valid. `None` once the slot holds a new commit.
@@ -474,6 +481,7 @@ impl TransactionHeader {
             user_root: None,
             system_root: None,
             transaction_id,
+            collection_horizon: TransactionId::new(0),
             corrupt_bytes: None,
         }
     }
@@ -519,12 +527,14 @@ impl TransactionHeader {
             None
         };
         let transaction_id = TransactionId::new(get_u64(&data[TRANSACTION_ID_OFFSET..]));
+        let collection_horizon = TransactionId::new(get_u64(&data[COLLECTION_HORIZON_OFFSET..]));
 
         let result = Self {
             version,
             user_root,
             system_root,
             transaction_id,
+            collection_horizon,
             corrupt_bytes: if corrupted {
                 Some(data[..TRANSACTION_SIZE].try_into().unwrap())
             } else {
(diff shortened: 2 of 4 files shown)


src/tree_store/page_store/header.rs
```diff
@@ -73,6 +73,10 @@ const USER_ROOT_OFFSET: usize = _UNUSED_OFFSET + size_of::<u8>() + PADDING;
 const SYSTEM_ROOT_OFFSET: usize = USER_ROOT_OFFSET + BtreeHeader::serialized_size();
 const _UNUSED2_OFFSET: usize = SYSTEM_ROOT_OFFSET + BtreeHeader::serialized_size();
 const TRANSACTION_ID_OFFSET: usize = _UNUSED2_OFFSET + BtreeHeader::serialized_size();
+// The last 8 bytes of what was padding before the collection horizon was recorded. Releases
+// that predate it zero the whole slot, so they read as a horizon of 0, and they checksum
+// these bytes like any others, so a slot carrying one still verifies for them
```
cberner: Remove

src/tree_store/page_store/header.rs
```diff
@@ -406,16 +410,22 @@ impl DatabaseHeader {
...
+    // through, or `None` for a commit that processed none; the stored horizon never moves
+    // backwards either way, since reuse an earlier commit made visible cannot be taken back
     pub(super) fn write_secondary_slot(
         &mut self,
         transaction_id: TransactionId,
         user_root: Option<BtreeHeader>,
         system_root: Option<BtreeHeader>,
+        collection_horizon: Option<TransactionId>,
     ) {
+        let floor = self.transaction_slots[self.primary_slot].collection_horizon;
```
cberner, line 417: Do we need to check read_secondary here?

src/tree_store/page_store/header.rs
```diff
@@ -406,16 +410,22 @@ impl DatabaseHeader {
     }
 
     // Overwrite the secondary slot with a newly committed transaction.
+    // `collection_horizon` is the newest transaction this commit processed freed pages
+    // through, or `None` for a commit that processed none; the stored horizon never moves
+    // backwards either way, since reuse an earlier commit made visible cannot be taken back
```
cberner: Needs work. Read all my past feedback on comments

src/tree_store/page_store/header.rs
```diff
@@ -461,6 +471,12 @@ pub(super) struct TransactionHeader {
     pub(super) user_root: Option<BtreeHeader>,
     pub(super) system_root: Option<BtreeHeader>,
     pub(super) transaction_id: TransactionId,
+    // The newest transaction whose freed pages have been processed for reuse. A page can only
+    // change after it is freed and reused, so this is the watermark a page cache is valid
+    // against: entries cached at or before it may be stale. It only ever advances. Nothing in a
+    // single process consults it -- the in-memory tracker knows more -- but a process that
+    // shares the file can learn it no other way, so every commit keeps it current
```
cberner: same

cberner (commented):
You need to do another pass on all the comments

src/tree_store/page_store/header.rs
cberner: Removed in 92e5697.

---

src/tree_store/page_store/header.rs
cberner: No -- and I think taking the primary is what makes it right rather than just adequate.

The horizon only ever advances at a durable commit, and that is also the only thing that swaps the slots. So the primary is always the last durable commit, which is the one holding the newest published horizon. A non-durable commit writes the secondary with `None`, which copies the primary's value forward rather than advancing it, so nothing is lost when the next durable commit overwrites that slot.

The case for *not* widening it to both slots: a commit that writes the secondary and then fails before the swap leaves a horizon there that was never published, and whose reclamation was rolled back with the rest of that transaction's allocator state. Reading the primary skips it; a `max` over both slots would adopt it and claim reuse that never happened.

---

src/tree_store/page_store/header.rs
cberner: Cut to the invariant in 92e5697, here and everywhere else the PR touches:

```rust
    // Overwrite the secondary slot with a newly committed transaction. The horizon never moves
    // backwards: reuse an earlier commit made visible cannot be taken back
```

What went: the restatement of the parameter, what `None` means (it is `unwrap_or(floor)` right there), and on the struct field the three lines of motivation for why this commit exists rather than what the field is. Same pass over the two test comments and the one in `transactions.rs`. The field is now:

```rust
    // The newest transaction whose freed pages a commit's reclamation has processed for reuse.
    // Bounds that reclamation alone: the pages a commit frees from the system tree once it is
    // done escape it
```

The second sentence is there for the Codex finding on the other thread, which is real -- it is the one thing a later consumer cannot work out from the code.

---

src/transactions.rs
```diff
@@ -2040,12 +2040,18 @@ impl WriteTransaction {
...
+        // horizon this commit publishes
         self.mem.commit(
             user_root,
             system_root,
             self.transaction_id,
             self.two_phase_commit,
             self.shrink_policy,
+            Some(TransactionId::new(
+                free_until_transaction.raw_id().saturating_sub(1),
+            )),
```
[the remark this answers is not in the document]
cberner: Confirmed. The immediate free is real reuse that the horizon does not account for, and it defeats the invariant the cache floor rests on.

That invariant, from #1413's coordinator: *a page cached from snapshot S was reachable at S, so only a transaction newer than S can have freed it -- a horizon above S may have reused it, one at or below cannot.* `revalidate_cache` is gated on `horizon > cache_floor` and nothing else, on both the read and the write-takeover path. So:

1. Process A holds a local read transaction at `r`. Peer B has a page P of the system tree cached at snapshot S, with `r < S`.
2. A commits `t1`, freeing P from the system tree. `process_freed_pages` is bounded by `free_until = r + 1`, so the published horizon is `r` -- but P is freed outright right after the commit, so it is allocatable now.
3. A commits `t2`, allocating P and writing new contents. The horizon is still `r`, because A's reader has not moved.
4. B checks `horizon (r) > cache_floor (S)` -- false. B keeps its stale copy of P.

The "only write transactions read them" property does not rescue it either: once P is reused it can come back as a *data* page, so B's next read transaction reaches it by page number through the same read cache and gets the stale bytes.

What I have done here is the half that belongs in this PR: the field no longer claims to be a cache watermark. It says what it actually bounds, and that these pages escape it (92e5697), so the protocol that consumes it cannot be written against a guarantee that is not there.

The fix itself belongs in #1413, where the coordinator exists to gate it, and I would rather not pick it unilaterally -- it is a change to core reclamation:

- **Defer the free under a coordinator.** Correct and precise, and what your second option describes. The hazard is that routing system-tree pages through the freed table means recording them, which frees more system-tree pages, which get recorded... I suspect the immediate free exists precisely to cut that loop, so this needs care rather than a one-line gate.
- **Quarantine instead of recording**: free them into a holding set the allocator only releases once a later commit's `free_until` passes `t1`. Keeps them out of the freed table, so no feedback loop, at the cost of new state.
- **Publish `max(free_until - 1, t1 - 1)` when system pages were freed.** Correct, one line, and safe in the conservative direction -- but the allocator-state table is rewritten every quick-repair commit, so system pages are freed on essentially every commit and the horizon would advance every time. Peers would drop their whole cache on every foreign commit, which is the cost the horizon exists to avoid.

I lean toward the quarantine. @cberner, which do you want, and does it ride here or in #1413?

---

cberner:
Second pass in 32b394c. The PR is down to six comments, 7 lines total:

```rust
// process_freed_pages ran through free_until_transaction, so reuse reaches the id below it
// Reuse an earlier commit made visible cannot be taken back, so the horizon only advances
// The newest transaction whose freed pages a commit's reclamation has processed for reuse.
// Bounds that alone: the pages a commit frees from the system tree once it is done escape it
// Read after the close, since on Windows the database's own locks are mandatory
// A torn write must invalidate the slot rather than parse as a different horizon
// What an older release writes: zeroed but for the fields it knows
```

What went in this pass: "Overwrite the secondary slot with a newly committed transaction", which restated the function name; a test doc comment that said the same thing as the inline comment on the assertion below it; and the two-line forms of the remaining four, each of which is now one line.

What stayed, and why each is not obvious from the code: the `-1` in the published value; why the `max` is there rather than a plain assignment; that the horizon does *not* bound the system-tree pages a commit frees outright, which is the Codex finding and the one thing a later consumer cannot derive; why the test reads the file after the close rather than during; and, in the test, what the two hand-built byte arrays represent.

---

PR #1422, opened 2026-08-29
Give the backends a locking interface of their own

cberner:
The locking a database does is a file backend's business today: each decides which locks to take, holds them for as long as the database is open, and releases them at close, which is the same decision written twice. The concurrency protocols of `docs/design.md` belong to redb rather than to a backend, and they need only one thing from one: the ability to lock a range of the storage.

So `InternalStorageBackend`, which requires `StorageBackend` and adds exactly that -- lock, lock shared, unlock -- and which the core takes in place of a `StorageBackend` from here on: `Database::new` and `ReadOnlyDatabase::new` through `TransactionalMemory::new` to `PagedCachedFile`.

It is not public. A backend a caller supplies has only the public trait to offer, so `create_with_backend` wraps it in `LocklessBackend`, which reports every lock unsupported. `create_with_backend` keeps taking `impl StorageBackend`, so there is no public API change.

## Taking the lock is the core's

`CheckedBackend` locks the whole storage when it is built and releases it at close, so a refused open is one place rather than two, and a backend is left to say only what it can do.

Whether it has locks to offer at all is the trait's to say. `locks_expected()` is true of everything but the caller's wrapper, whose absent locks are neither asked for nor reported -- `create_with_backend` does no locking, and that is expected, so it is not what the "File locks not supported on this platform" warning is about. That warning now has one site instead of one per backend.

A trait constant would say `locks_expected` without a receiver, but an associated constant makes a trait dyn-incompatible, and the core holds a `dyn InternalStorageBackend`, so it is a defaulted method.

## What the backends are left with

The platform reasoning alone:

| backend | whole storage | a narrower range |
|---|---|---|
| `optimized` | the protocol ranges, and -- where those are a namespace of their own -- the whole-file lock an older redb takes as well | the byte-range lock alone |
| `fallback` | the whole-file lock, which is all it has | unsupported |
| a caller's, wrapped | unsupported | unsupported |

The whole storage is the range a single-process open locks, and the whole-file lock covers exactly that, which is why both kinds are held for it and only for it. `FULL_RANGE` names it. Which locks are held became state rather than a constructor's result, since the interface takes them by shared reference.

**One fallback is left, and it is a question for review.** `lock_protocol_ranges` no longer falls back to the whole-file lock when range locks are unsupported. What remains is one arm in the `CONFLICTS_WITH_STD_FILE_LOCK == Some(true)` branch, the only place it can still be reached, which is Windows: its `RangeLock` impl is const-only, so `try_lock_range` returns `Unsupported` there, and that arm is what locks the database on Windows today. Removing it outright opens Windows unlocked and turns `basic_tests`' `DatabaseAlreadyOpen` assertions red. It should go the moment `RangeLock` is implemented for Windows.

## Scope

No behavior changes. The protocols move into the core in a later change, which is what the narrower ranges and the signal are for; #1419 rebases onto this and drives the trait rather than holding its locks itself.

## Testing

Two tests, one property each: a caller-supplied backend reports the locks unsupported and says that is expected, with the storage underneath it still working; and a range short of the whole storage is a byte-range lock alone, which is the interface the protocols will use. The existing lock tests cover the whole-storage range through a helper that does what the core does, since the backends no longer lock when they are built.

`cargo fmt`, `cargo clippy --all-targets` in four feature configurations (including `--features logging` alone, where the reporting is compiled differently), cross-clippy for `x86_64-pc-windows-msvc` and `aarch64-apple-darwin`, wasm32 in three configurations -- which is where `fallback.rs` actually compiles -- `wasm32-wasip1`, no_std in two, `cargo doc`, and `cargo test --all-features` all pass. `cargo deny` was not run; no dependency changed. The WASI test run is the one thing local validation here cannot do, and it is what caught an earlier revision.

### DIFF
--- a/src/tree_store/page_store/file_backend/fallback.rs
+++ b/src/tree_store/page_store/file_backend/fallback.rs
@@ -1,51 +1,83 @@
+use crate::db::{FULL_RANGE, InternalStorageBackend};
 use crate::{DatabaseError, Result, StorageBackend};
-#[cfg(feature = "logging")]
-use log::warn;
 use std::fs::{File, TryLockError};
 use std::io;
 use std::io::{Read, Seek, SeekFrom, Write};
+use std::ops::Range;
 use std::sync::Mutex;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 /// Stores a database as a file on-disk.
 #[derive(Debug)]
 pub struct FileBackend {
-    lock_supported: bool,
+    whole_file_locked: AtomicBool,
     file: Mutex<File>,
 }
 
 impl FileBackend {
     /// Creates a new backend which stores data to the given file.
     pub fn new(file: File) -> Result<Self, DatabaseError> {
-        Self::new_internal(file, false)
+        Ok(Self {
+            whole_file_locked: AtomicBool::new(false),
+            file: Mutex::new(file),
+        })
     }
 
-    pub(crate) fn new_internal(file: File, read_only: bool) -> Result<Self, DatabaseError> {
-        let result = if read_only {
+    fn lock_whole_file(&self, shared: bool) -> Result<bool, io::Error> {
+        let file = self.file.lock().unwrap();
+        let result = if shared {
             file.try_lock_shared()
         } else {
             file.try_lock()
         };
 
         match result {
-            Ok(()) => Ok(Self {
-                lock_supported: true,
-                file: Mutex::new(file),
-            }),
-            Err(TryLockError::WouldBlock) => Err(DatabaseError::DatabaseAlreadyOpen),
-            Err(TryLockError::Error(err)) if err.kind() == io::ErrorKind::Unsupported => {
-                #[cfg(feature = "logging")]
-                warn!(
-                    "File locks not supported on this platform. You must ensure that only a single process opens the database file, at a time"
-                );
-
-                Ok(Self {
-                    lock_supported: false,
-                    file: Mutex::new(file),
-                })
+            Ok(()) => {
+                self.whole_file_locked.store(true, Ordering::Release);
+                Ok(true)
             }
-            Err(TryLockError::Error(err)) => Err(err.into()),
+            Err(TryLockError::WouldBlock) => Ok(false),
+            Err(TryLockError::Error(err)) => Err(err),
         }
     }
+}
+
+/// This backend has no byte-range locks. Only the whole storage can be locked, which a whole-file
+/// lock does, so that is the one range it supports.
+impl InternalStorageBackend for FileBackend {
+    fn try_lock_range(&self, range: Range<u64>) -> Result<bool, io::Error> {
+        if range == FULL_RANGE {
+            self.lock_whole_file(false)
+        } else {
+            Err(unsupported())
+        }
+    }
+
+    fn try_lock_shared_range(&self, range: Range<u64>) -> Result<bool, io::Error> {
+        if range == FULL_RANGE {
+            self.lock_whole_file(true)
+        } else {
+            Err(unsupported())
+        }
+    }
+
+    fn unlock_range(&self, range: Range<u64>) -> Result<(), io::Error> {
+        if range != FULL_RANGE {
+            return Err(unsupported());
+        }
+        if self.whole_file_locked.swap(false, Ordering::AcqRel) {
+            self.file.lock().unwrap().unlock()?;
+        }
+
+        Ok(())
+    }
+}
+
+fn unsupported() -> io::Error {
+    io::Error::new(
+        io::ErrorKind::Unsupported,
+        "byte-range locks are not supported on this platform",
+    )
 }
 
 impl StorageBackend for FileBackend {
@@ -72,13 +104,5 @@ impl StorageBackend for FileBackend {
         let mut file = self.file.lock().unwrap();
         file.seek(SeekFrom::Start(offset))?;
         file.write_all(data)
-    }
-
-    fn close(&self) -> Result<(), io::Error> {
-        if self.lock_supported {
-            self.file.lock().unwrap().unlock()?;
-        }
-
-        Ok(())
     }
 }
--- a/src/tree_store/page_store/file_backend/optimized.rs
+++ b/src/tree_store/page_store/file_backend/optimized.rs
@@ -1,7 +1,9 @@
+use crate::db::{FULL_RANGE, InternalStorageBackend};
 use crate::{DatabaseError, Result, StorageBackend};
 use std::fs::{File, TryLockError};
 use std::io;
 use std::ops::Range;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 #[cfg(feature = "logging")]
 use log::warn;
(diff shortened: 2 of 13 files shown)


src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -46,79 +57,113 @@ impl FileBackend {
...
+        let acquired = if read_only {
+            backend.try_lock_shared_range(FULL_RANGE)
+        } else {
+            backend.try_lock_range(FULL_RANGE)
+        };
+        match acquired {
+            Ok(true) => Ok(backend),
+            Ok(false) => Err(DatabaseError::DatabaseAlreadyOpen),
+            Err(err) => Err(err.into()),
+        }
```
cberner: fallback has the same duplicated code. Let's move it out of this layer into the cached_file or another appropriate part of redb core

cberner (commented):
I haven't reviewed all of the optimized.rs changes, but they're bigger than I expected. Is any of that unnecessary given how it is going to be used?

src/tree_store/page_store/file_backend/optimized.rs
cberner: Moved in `10b84cc`. `CheckedBackend` takes the whole-storage lock when it is built and releases it at close, so the refused open is one place instead of two, and `new_internal` is gone from both file backends -- `FileBackend::new` now only opens.

That also answered your other question: a fair amount of the optimized.rs churn existed only to serve that duplicated constructor. What is left there is the platform reasoning alone, and most of its diff is transformation rather than new logic -- the same branches, moved from a constructor onto `&self` and returning `bool` instead of `DatabaseError`. Two things went with the move:

* `WholeFileLock` is gone. It existed to carry "the platform has no locks" back as a third outcome so the backend could warn and claim success. That is now just an `Unsupported` error travelling up to the one place that decides what to do about it, so `try_whole_file_lock` is a plain `io::Result<bool>`.
* Neither backend reports missing locks any more. `locks_expected()` gates the attempt: the wrapper says it has none, so an open does not ask it and does not warn, and the "File locks not supported on this platform" message has one site rather than two.

One behavior change worth flagging: a `FileBackend` handed to `create_with_backend` is no longer locked -- by the backend, which no longer locks in its constructor, or by the core, which sees a `dyn StorageBackend` and cannot know it has locks. `Database::create`/`open`/`create_file` are unaffected, since those construct the backend themselves. In-tree that path is only `FailingBackend`/`FuzzerBackend` wrapping a `FileBackend` in tests and the fuzzer, none of which rely on the lock. Say the word if you would rather keep it and I will find another way.

---

src/tree_store/page_store/file_backend/fallback.rs
```diff
@@ -1,53 +1,87 @@
...
+        Ok(Self {
+            whole_file_locked: AtomicBool::new(false),
+            file: Mutex::new(file),
+        })
     }
 
-    pub(crate) fn new_internal(file: File, read_only: bool) -> Result<Self, DatabaseError> {
-        let result = if read_only {
+    /// An `Unsupported` error is the platform having no file locks at all, which the caller
+    /// reports rather than the backend
```
cberner: Remove

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -34,91 +36,104 @@ fn report_failed_release(result: io::Result<()>) {
...
 
-    fn new_holding_ranges(file: File, read_only: bool) -> Result<Self, DatabaseError> {
-        match Self::try_lock_protocol_ranges(&file, read_only) {
-            Ok(true) => Ok(Self {
-                whole_file_locked: false,
-                ranges_locked: true,
-                file,
-            }),
+    /// The ranges alone, which is what a namespace shared with the whole-file lock calls for,
+    /// falling back to the whole-file lock where the filesystem has no range locks
```
cberner: The optimized platforms all have range locks, so remove this fallback

cberner (commented):
double check your changes in optimized. I didn't look at all of it, but it seems like it could be polished more

src/tree_store/page_store/file_backend/optimized.rs
cberner: Fixed in `bfcabe8`, rather than left for you to decide -- Codex flagged the same thing as a P1 with the corruption scenario spelled out, and two independent readings of it is enough.

`FileBackend::new` takes the whole-storage lock again. It is not a second copy of the decision: the mapping lives in one `lock_whole_storage()` next to the trait, which both `CheckedBackend` and the two public constructors call. The opens redb makes for itself start from a new unlocked `new_internal()` instead, so nothing is locked twice and the read-only open still takes its lock shared.

`close()` is back on both file backends for the same reason -- whatever `new` took has to come back when redb only sees a `StorageBackend`. On the core's own path that unlock runs twice, which is a no-op: the flags are swapped, so the second finds nothing held.

`a_caller_supplied_file_backend_still_excludes_a_second_open` covers it through the `Builder`, and I checked it fails when the constructor's lock is stubbed out.

---

src/tree_store/page_store/file_backend/optimized.rs
cberner: Removed from `lock_protocol_ranges` in `842bc16` -- but not entirely, because one platform that compiles `optimized.rs` does not have range locks: Windows. Its `RangeLock` impl is const-only (`CONFLICTS_WITH_STD_FILE_LOCK = Some(true)` and nothing else), so `try_lock_range` returns `Unsupported` there, and that fallback is what locks the database on Windows today. Dropping it outright would open Windows unlocked and turn `basic_tests`' `DatabaseAlreadyOpen` assertions red on windows-latest.

So what is gone is the *general* fallback -- "the filesystem has no range locks" as a condition anywhere the ranges are taken. What is left is one arm in the `CONFLICTS_WITH_STD_FILE_LOCK == Some(true)` branch, which is the only place it can now be reached, named for what it is:

```rust
// Prefer range locks
if File::CONFLICTS_WITH_STD_FILE_LOCK == Some(true) {
    return match self.lock_protocol_ranges(shared) {
        // Windows, whose range locks are not implemented yet and whose whole-file lock is
        // itself one over every byte
        Err(err) if err.kind() == io::ErrorKind::Unsupported => self.lock_whole_file(shared),
        result => result,
    };
}
```

Happy to delete it the moment `RangeLock` is implemented for Windows, or now if dropping Windows locking until then is what you want -- say which.

The rest of the polish went in with it:

* `try_lock_protocol_ranges` and `release_protocol_ranges` folded into `lock_protocol_ranges`, so taking the ranges, releasing the ones already taken, and recording that they are held is one function rather than three.
* `whole_file_locked` is now set where the lock is taken rather than at each call site, as `ranges_locked` already was, so neither flag is stored from more than one place.
* The tail of `lock_whole_storage` no longer repeats the success case: it keeps the result and releases the whole-file lock only on the paths that need it.
* The doc comment you flagged on the fallback backend is gone, and its twin in `optimized.rs` with it.

---

src/tree_store/page_store/file_backend/fallback.rs
cberner: Removed in `842bc16`, along with the same comment on `optimized.rs`'s whole-file lock.

---

src/tree_store/page_store/file_backend/fallback.rs
```diff
@@ -1,53 +1,97 @@
...
+    whole_file_locked: AtomicBool,
     file: Mutex<File>,
 }
 
 impl FileBackend {
     /// Creates a new backend which stores data to the given file.
+    ///
+    /// The whole storage is locked, so that no other process opens the same file: a backend
+    /// handed to [`Builder::create_with_backend`](crate::Builder::create_with_backend) reaches
+    /// redb as a [`StorageBackend`], whose locks it cannot see.
```
cberner: No, remove this. create_with_backend doesn't do locking and that is expected.

src/tree_store/page_store/file_backend/fallback.rs
```diff
@@ -1,53 +1,97 @@
...
     pub fn new(file: File) -> Result<Self, DatabaseError> {
-        Self::new_internal(file, false)
+        let backend = Self::new_internal(file);
+        lock_whole_storage(&backend, false)?;
```
cberner: remove

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -34,131 +36,157 @@ fn report_failed_release(result: io::Result<()>) {
...
+    ranges_locked: AtomicBool,
     file: File,
 }
 
 impl FileBackend {
     /// Creates a new backend which stores data to the given file.
+    ///
+    /// The whole storage is locked, so that no other process opens the same file: a backend
+    /// handed to [`Builder::create_with_backend`](crate::Builder::create_with_backend) reaches
+    /// redb as a [`StorageBackend`], whose locks it cannot see.
```
cberner: same

src/tree_store/page_store/file_backend/fallback.rs
cberner: Understood, and reverted in `0ab6975`. Both constructors are plain again, `new_internal` is gone from both, and the `close()` overrides with it -- whoever takes the lock releases it, and that is only the core now.

`lock_whole_storage`/`unlock_whole_storage` went back to being `CheckedBackend` methods too, since it is their only caller once the constructors do not lock. `db.rs` is left with the trait and `FULL_RANGE`.

`a_caller_supplied_file_backend_still_excludes_a_second_open` is deleted -- it asserted exactly the behavior you are saying is not expected. That also removes the WASI failure it caused on the previous head, where there are no file locks and so no `DatabaseAlreadyOpen` to assert.

---

cberner:
lgtm

PR #1425, opened 2026-08-30
Move file locking to core instead of the backend

cberner:
Split out of #1419, which this is a prerequisite for. No behaviour change and no API change.

## What moves

`CheckedBackend` took the lock an open holds for the life of the database, and decided the policy with it: which range, shared or exclusive, and what an unsupported or a refused lock means. That is the locking protocol, and with the multi-process modes coming it stops being expressible there at all -- those take several ranges, chosen by a mode the backend has no way to know.

So `CheckedBackend` now records what it was asked to take and releases it at close, and `TransactionalMemory::lock_whole_storage` decides. `lock_storage`/`release_storage` become `try_lock_range(range, shared)`/`release_locks()`, over a recorded vector of ranges rather than a hardcoded `FULL_RANGE`.

Nothing about the result changes: the same single range, taken shared for a read-only open, `DatabaseAlreadyOpen` where it conflicts, a `logging` warning and an open that carries on where the platform has no locks, and a caller-supplied backend still passing without one.

Two consequences worth naming:

Constructing the backend no longer locks, so `CheckedBackend::new` and `PagedCachedFile::new` cannot fail and stop returning `Result`.

`release_locks()` drops the two `is_unsupported`/`locks_expected` guards `release_storage()` needed. They are unreachable now rather than removed: nothing is recorded unless the lock was actually acquired, so there is nothing to release in either case.

## The probe byte

`NAMESPACE_PROBE_BYTE` moves to `db.rs` alongside `FULL_RANGE`, expressed against a `LOCK_BASE` constant rather than a repeated `(1 << 62) - 2`. It describes an offset in the protocol but lived in one backend, where the core could not name it -- and the core is where the rest of the protocol's offsets are headed.

Both are gated on `any(windows, unix, target_os = "wasi")`, the same cfg as the file backend that uses them and as `range_lock` beside them: on wasm32 and thumbv7em there is no `optimized.rs`, so without the gate they are dead code and CI's `--deny warnings` rejects them.

I left `docs/design.md` alone. Adding the probe byte to the byte table was your call to make and you already made it on #1419: it is an implementation detail that goes away if the range locks become unconditional.

## Testing

`lock_failure_test` drives `lock_whole_storage` over a backend that answers as told, one test per outcome: unsupported opens anyway, a refused lock is `DatabaseAlreadyOpen`, any other error fails the open, and `the_lock_is_released_at_close` asserts that closing releases exactly the range that was taken -- and that nothing is released when the lock was never acquired, which is what makes dropping those two guards safe.

`cargo fmt --all -- --check`, `cargo clippy --all --all-targets --all-features` as CI runs it, clippy over `experimental-api-5`/`logging`/no features, cross-clippy for `x86_64-pc-windows-msvc` and `aarch64-apple-darwin`, the three `wasm32-unknown-unknown` variants CI checks, no_std on `thumbv7em-none-eabihf`, and `cargo test` all pass, all with `RUSTFLAGS=--deny warnings`. Also run for `x86_64-pc-windows-gnu` under wine, where the only failure is `a_whole_file_lock_holder_reads_as_already_open`, which fails identically on unmodified master there: wine answers std's second `UnlockFile` with `ERROR_LOCK_VIOLATION` where Windows answers `ERROR_NOT_LOCKED`.

No changelog entry -- nothing here is visible to users.

### DIFF
--- a/src/tree_store/page_store/cached_file.rs
+++ b/src/tree_store/page_store/cached_file.rs
@@ -1,21 +1,18 @@
-use crate::db::{FULL_RANGE, InternalStorageBackend};
+use crate::db::InternalStorageBackend;
 use crate::io;
 use crate::sync::{Mutex, MutexGuard, RwLock};
 use crate::tree_store::page_store::base::PageHint;
 use crate::tree_store::page_store::lru_cache::LRUCache;
-use crate::{CacheStats, DatabaseError, Result, StorageError};
+use crate::{CacheStats, Result, StorageError};
 use alloc::boxed::Box;
 use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
-use core::ops::{Index, IndexMut};
+use core::ops::{Index, IndexMut, Range};
 use core::slice::SliceIndex;
 #[cfg(feature = "cache_metrics")]
 use core::sync::atomic::AtomicU64;
 use core::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-#[cfg(feature = "logging")]
-use log::warn;
-
 // Allocates an `Arc<[u8]>` in one step. `Arc::<[u8]>::from(vec![0; len])` would
 // allocate the Vec and then allocate a new Arc and memcpy into it.
 fn zero_filled_arc(len: usize) -> Arc<[u8]> {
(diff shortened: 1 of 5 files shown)


src/tree_store/page_store/cached_file.rs
```diff
@@ -135,58 +134,50 @@ struct CheckedBackend {
...
-            return Ok(());
-        }
+    fn locks_expected(&self) -> bool {
+        self.file.locks_expected()
+    }
 
-        let acquired = if read_only {
-            self.file.try_lock_shared_range(FULL_RANGE)
+    /// Takes a lock held for as long as the database is open. `Ok(false)` means a conflicting
+    /// lock is held elsewhere.
```
cberner: remove

cberner (commented):
the comments need major work

src/tree_store/page_store/cached_file.rs
```diff
@@ -135,58 +134,48 @@ struct CheckedBackend {
...
-            result => result,
+    fn release_locks(&self) -> Result<(), io::Error> {
+        // Every lock is released even where one fails: a lock left behind outlives this backend
+        // wherever the file description is shared
+        let mut result = Ok(());
+        for range in self.locked.lock().unwrap().drain(..) {
+            result = result.and(self.file.unlock_range(range));
         }
+
+        result
```
cberner: This doesn't seem right. To implement the MP protocol the TransactionalMemory needs to be taking and releasing granular range locks. So let's refactor this to expose that up through the layers and make TransactionalMemory do it. And to make the drop maximally safe, I think the optimized.rs implementation should remember its locked ranges and clear them on close(), and the fallback should remember if it has a lock and clear its

cberner: Done in `b37af4d`. `CheckedBackend`'s `locked` vec is gone; the granular range interface is exposed up to `TransactionalMemory`, and each backend records what it holds and releases it in `close()` -- `optimized.rs` its ranges plus the whole-file lock, `fallback.rs` its whole-file lock.

On your follow-up question, since it changes what the recording is for: the auto-release premise is right on all three platforms, but it does not make the explicit release redundant, and I measured rather than assuming.

**Does closing the description release?** Yes, on the last descriptor. Linux, `F_OFD_SETLK`:

```
unlock superset(0..1000) of lock(100..110) -> rc=0, still held=false
after dropping one of two FDs on the same OFD, held=true
after dropping the last FD to the OFD,        held=false
```

Windows, `LockFileEx`, same probe:

```
unlock superset(0..1000) of lock(100..110) -> Err(33), still held=true
unlock subset(100..105)   of lock(100..110) -> Err(33)
unlock exact(100..110)                      -> Ok(()), still held=false
after closing one of two duplicated handles, held=true
after closing the last handle,              held=false
```

(Windows run under wine, so treat it as corroborating the documented behaviour rather than proving it. `ERROR_LOCK_VIOLATION` is 33.)

**So why release explicitly?** Not because of `try_clone()` -- agreed that is the caller's problem. Because redb itself holds the file past `close()`. A read transaction owns an `Arc<TransactionalMemory>`, so dropping the `Database` calls `close()` while the reader keeps the file alive. Stubbing the release out of `close()` and reopening under exactly that:

```
reopen with a live reader: Err("Database already open. Cannot acquire lock.")
```

With the release in place it reopens fine. `close_releases_while_the_backend_is_still_alive` pins it, and I checked it fails when the release is stubbed rather than passing vacuously.

**So why remember the ranges, rather than one span?** That is Windows alone. On Linux and the Apple platforms `fcntl` splits and merges, so unlocking `0..u64::MAX` would release everything -- the first line of the Linux output. `UnlockFile` refuses both a superset and a subset, so the exact ranges have to be known. A fixed set would have done while the only ranges were `PROTOCOL_RANGES`; recording them is what lets `TransactionalMemory` take arbitrary granular ranges, which is the point.

---

src/tree_store/page_store/file_backend/optimized.rs
```diff
@@ -37,7 +34,9 @@ fn report_failed_release(result: io::Result<()>) {
 #[derive(Debug)]
 pub struct FileBackend {
     whole_file_locked: AtomicBool,
-    ranges_locked: AtomicBool,
+    // Every range this handle holds, so that close() can release exactly what was taken:
+    // LockFileEx neither splits nor merges, so an unlock must cover the range it was locked over
+    locked_ranges: Mutex<Vec<Range<u64>>>,
```
cberner: Vec is the wrong data structure, once the protocol is implemented we'll be adding and removing things so use a HashSet

PR #1426, opened 2026-08-30
Lock the header while a shared-mode process reads or writes it

cberner:
The next slice of the multi-process work, after #1419's concurrency modes. Those negotiate the open and then coordinate nothing, so two processes sharing a database can read a header the other is part way through writing. This is the first real coordination: every read of the header bytes runs inside a shared hold and every write inside an exclusive one.

Only the shared modes take it. A single-process handle holds the whole storage already, and there is no other process to exclude.

## Where the lock lives

The holds are taken in `TransactionalMemory`, at the eight places it reads or writes the header. `write_header()` is the only one of those that writes during normal operation, so the reload path's repair write goes through it too rather than writing the header itself.

The lock covers the header bytes themselves rather than a byte standing for them. That is what makes Windows' mandatory locks hold back another process's read of those bytes, rather than merely advising against it.

Covering the real bytes means the holder has to be the open file description performing the I/O it serializes -- on Windows, a lock taken on some other description would block this process's own header reads. So the lock is taken through the storage backend, which is that description by construction. `InternalStorageBackend` gains blocking `lock_range()` and `lock_shared_range()`, alongside the `try_` forms an open already uses; a custom backend can implement them, and the ones that have no locks report `Unsupported` as they do for the rest.

The alternative would be handing the hold its own `File` and trusting it to be the same description. Going through the backend removes that obligation.

## Getting the write inside the hold

`PagedCachedFile::write()` only fills the write buffer; the bytes reach the file later, from `flush_write_buffer()` or from an eviction under cache pressure. A hold around it would cover a memcpy and nothing else.

So `write_header()` writes directly instead, through a new `write_direct()` that mirrors the existing `read_direct()`: it drops the cached copy and hands the bytes to the backend. The fsync stays outside, in the flush that follows.

Writing the header publishes it -- a peer can read it as soon as the hold ends -- so the pages it names have to be on the file by then. That ordering comes from two-phase commit, which the shared modes require: the flush between `commit()`'s two `write_header()` calls puts the pages down before the primary slot swap is published. `write_header()` itself does not flush.

## The in-process mutex

Each hold also takes a mutex for its duration, which is not redundant with the byte lock. Both holds run on one file description, where a second acquisition converts the existing hold rather than conflicting with it, and one unlock then drops whatever remains -- so two overlapping holds in this process would dismantle each other. The mutex is what keeps them from overlapping.

The `state` mutex cannot serve instead. The commit path clones the header and drops it before writing, on purpose, so as not to hold it across the file I/O; `clear_cache_and_reload()` and `durable_primary_slot_corrupt()` do not hold it either; and the open path has no `TransactionalMemory` yet.

## Why a single-process handle takes nothing

The same mechanism, and it is the sharp edge here. A single-process open holds the whole storage. Locking the header bytes on that description and releasing them again would convert part of that hold and then drop it -- punching a hole in the whole-storage lock, and letting a second process open the database. `a_hold_does_not_puncture_the_whole_storage_lock` covers exactly that: it runs both kinds of hold on a single-process handle and then asserts a second open is still refused.

## What is not here

The header I/O is atomic; a multi-step read-modify-write over the header is not. An open that reads the header, decides it needs recovery and writes it back does so under two holds rather than one. What makes that safe against a live cohort is holding the writer byte across construction, which arrives with the rest of the concurrency-mode wiring.

Also still to come: the pin protocol, which is what a reclaiming writer needs in order to know what the other processes are still reading, and the transaction-level reload and cache invalidation. A database opened in a shared mode still is not safe to actually share.

## Testing

Six tests in `header_lock_test`, on real files, driving two `TransactionalMemory` handles over one path -- two open file descriptions contend exactly as two processes do:

* `an_exclusive_hold_excludes_a_shared_one`: the reader proceeds only once the writer's hold ends. Not vacuous -- stubbing the lock out so every mode takes nothing makes it fail with "the reader entered its hold while the writer held one".
* `shared_holds_admit_each_other`, since only a write needs the header to itself.
* `the_hold_covers_the_write_reaching_the_file`: a held shared hold blocks `write_header()` until it is released.
* `the_header_reaches_the_file_before_write_header_returns`: the other half of that, and the one that pins the direct write down. It reads the file around redb entirely, with no flush in between. The test above passes either way, since a buffered write blocks on the hold too; this one fails if the write is left to the next flush. I re-checked that by mutating `write_header()` back to the buffered shape, which fails this test alone.
* `a_read_only_handle_takes_shared_holds`, through the read-only backend an open wraps its storage in.
* `a_hold_does_not_puncture_the_whole_storage_lock`, the single-process case above.

`cargo fmt`, the justfile's `cargo clippy --all-targets --all-features` and `cargo clippy --all --all-targets --all-features` as CI runs them, clippy with no features, cross-clippy for `x86_64-pc-windows-msvc` and `aarch64-apple-darwin`, the three `wasm32` variants, both `thumbv7em-none-eabihf` no_std configurations CI checks, and `cargo test` in all three configurations that run tests -- no features, `experimental-multiprocess`, and `--all-features` -- all pass.

Also run for `x86_64-pc-windows-gnu` under wine, where all six header-lock tests pass. Two tests in the wider suite fail there and fail identically on this branch's merge base, so neither is this change's: `a_whole_file_lock_holder_reads_as_already_open`, because wine answers std's second `UnlockFile` with `ERROR_LOCK_VIOLATION` where Windows answers `ERROR_NOT_LOCKED`, and `does_not_exist`, because a deleted-but-still-open file keeps its name on Windows.

---

### DIFF
--- a/src/tree_store/page_store/cached_file.rs
+++ b/src/tree_store/page_store/cached_file.rs
@@ -159,6 +159,21 @@ impl CheckedBackend {
         self.file.try_lock_shared_range(range)
     }
 
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_range(&self, range: Range<u64>) -> Result<(), io::Error> {
+        self.file.lock_range(range)
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_shared_range(&self, range: Range<u64>) -> Result<(), io::Error> {
+        self.file.lock_shared_range(range)
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    fn unlock_range(&self, range: Range<u64>) -> Result<(), io::Error> {
+        self.file.unlock_range(range)
+    }
+
     fn query_lock_range(&self, range: Range<u64>) -> Result<bool, io::Error> {
         self.file.query_lock_range(range)
     }
@@ -351,6 +366,21 @@ impl PagedCachedFile {
 
     pub(crate) fn try_lock_shared_range(&self, range: Range<u64>) -> Result<bool, io::Error> {
         self.file.try_lock_shared_range(range)
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn lock_range(&self, range: Range<u64>) -> Result<(), io::Error> {
+        self.file.lock_range(range)
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn lock_shared_range(&self, range: Range<u64>) -> Result<(), io::Error> {
+        self.file.lock_shared_range(range)
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn unlock_range(&self, range: Range<u64>) -> Result<(), io::Error> {
+        self.file.unlock_range(range)
     }
 
     /// Whether an exclusive lock over the range would conflict with one held elsewhere.
@@ -549,6 +579,14 @@ impl PagedCachedFile {
         if self.write_buffer_bytes.load(Ordering::Acquire) > 0 {
             self.committed_pages_buffered.store(true, Ordering::Release);
         }
+    }
+
+    // Write directly to the file, bypassing the write buffer, so the bytes are on the file when
+    // this returns rather than whenever the buffer is next flushed
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(super) fn write_direct(&self, offset: u64, data: &[u8]) -> Result<()> {
+        self.invalidate_cache(offset, data.len());
+        self.file.write(offset, data)
     }
 
     // Read directly from the file, ignoring any cached data
--- a/src/tree_store/page_store/file_backend/range_lock.rs
+++ b/src/tree_store/page_store/file_backend/range_lock.rs
@@ -21,6 +21,17 @@ pub(crate) trait RangeLock {
         Err(unsupported())
     }
 
+    /// Waits for the range. No deadlock detection, like the whole-file locks.
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_range(&self, _range: Range<u64>) -> io::Result<()> {
+        Err(unsupported())
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_shared_range(&self, _range: Range<u64>) -> io::Result<()> {
+        Err(unsupported())
+    }
+
     fn unlock_range(&self, _range: Range<u64>) -> io::Result<()> {
         Err(unsupported())
     }
@@ -62,6 +73,8 @@ mod windows_imp {
     const LOCKFILE_FAIL_IMMEDIATELY: Dword = 0x0000_0001;
     const LOCKFILE_EXCLUSIVE_LOCK: Dword = 0x0000_0002;
     const ERROR_LOCK_VIOLATION: i32 = 33;
+    #[cfg(feature = "experimental-multiprocess")]
+    const ERROR_IO_PENDING: i32 = 997;
 
     #[repr(C)]
     struct Overlapped {
@@ -89,6 +102,22 @@ mod windows_imp {
             bytes_low: Dword,
             bytes_high: Dword,
         ) -> Bool;
+        #[cfg(feature = "experimental-multiprocess")]
+        fn CreateEventW(
+            attributes: *mut core::ffi::c_void,
+            manual_reset: Bool,
+            initial_state: Bool,
+            name: *const u16,
+        ) -> Handle;
+        #[cfg(feature = "experimental-multiprocess")]
+        fn GetOverlappedResult(
+            file: Handle,
+            overlapped: *mut Overlapped,
+            transferred: *mut Dword,
+            wait: Bool,
+        ) -> Bool;
+        #[cfg(feature = "experimental-multiprocess")]
+        fn CloseHandle(handle: Handle) -> Bool;
     }
 
     // The Win32 calls take 64-bit offsets and lengths as dword halves
@@ -140,6 +169,64 @@ mod windows_imp {
         }
     }
 
+    // Waits for the range, the way std's own File::lock waits for the whole file. A handle
+    // opened for asynchronous I/O answers ERROR_IO_PENDING and completes the request later,
+    // which is what the event is for: waiting on it settles the request before this returns,
+    // so the overlapped structure is safe on the stack.
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_blocking(file: &File, exclusive: bool, range: Range<u64>) -> io::Result<()> {
+        debug_assert!(!range.is_empty());
+        let flags = if exclusive {
+            LOCKFILE_EXCLUSIVE_LOCK
+        } else {
+            0
+        };
+        let len = range.end - range.start;
+        let handle = file.as_raw_handle();
+        unsafe {
+            let event = CreateEventW(core::ptr::null_mut(), 0, 0, core::ptr::null());
+            if event.is_null() {
+                return Err(io::Error::last_os_error());
+            }
+            let mut overlapped = Overlapped {
+                internal: 0,
+                internal_high: 0,
+                offset: low_dword(range.start),
+                offset_high: high_dword(range.start),
+                event,
+            };
+            let acquired = LockFileEx(
+                handle,
+                flags,
+                0,
+                low_dword(len),
+                high_dword(len),
+                &raw mut overlapped,
+            );
+
+            let result = if acquired != 0 {
+                Ok(())
+            } else {
+                let err = io::Error::last_os_error();
+                if err.raw_os_error() == Some(ERROR_IO_PENDING) {
+                    let mut transferred: Dword = 0;
+                    let completed =
+                        GetOverlappedResult(handle, &raw mut overlapped, &raw mut transferred, 1);
+                    if completed != 0 {
+                        Ok(())
+                    } else {
+                        Err(io::Error::last_os_error())
+                    }
+                } else {
+                    Err(err)
+                }
+            };
+            CloseHandle(event);
+
+            result
+        }
+    }
+
     impl RangeLock for File {
         // std's whole-file lock is itself a LockFileEx over every byte, so it is one of these
         const CONFLICTS_WITH_STD_FILE_LOCK: Option<bool> = Some(true);
@@ -150,6 +237,16 @@ mod windows_imp {
 
         fn try_lock_shared_range(&self, range: Range<u64>) -> io::Result<bool> {
             try_lock(self, false, range)
+        }
+
+        #[cfg(feature = "experimental-multiprocess")]
+        fn lock_range(&self, range: Range<u64>) -> io::Result<()> {
+            lock_blocking(self, true, range)
+        }
+
+        #[cfg(feature = "experimental-multiprocess")]
+        fn lock_shared_range(&self, range: Range<u64>) -> io::Result<()> {
+            lock_blocking(self, false, range)
         }
 
         // UnlockFile rather than its Ex form, as std's unlock uses: it takes the range as
@@ -249,6 +346,30 @@ fn set_lock(file: &File, exclusive: bool, range: Range<u64>) -> io::Result<bool>
     }
 }
 
+// EINTR is the caller's signal handler having run, not a failure to take the lock.
+#[cfg(all(
+    feature = "experimental-multiprocess",
+    any(target_os = "linux", target_vendor = "apple")
+))]
+fn set_lock_blocking(file: &File, exclusive: bool, range: Range<u64>) -> io::Result<()> {
+    let kind = lock_type(if exclusive {
+        libc::F_WRLCK
+    } else {
+        libc::F_RDLCK
+    });
+    loop {
+        let mut lock = flock_struct(kind, range.clone());
+        let rc = unsafe { libc::fcntl(file.as_raw_fd(), libc::F_OFD_SETLKW, &raw mut lock) };
+        if rc == 0 {
+            return Ok(());
+        }
+        let err = io::Error::last_os_error();
+        if err.kind() != io::ErrorKind::Interrupted {
+            return Err(err);
+        }
+    }
+}
+
 #[cfg(any(target_os = "linux", target_vendor = "apple"))]
 impl RangeLock for File {
     // The Apple platforms answer flock() with the same locks fcntl() takes; Linux keeps them
@@ -265,6 +386,16 @@ impl RangeLock for File {
 
     fn try_lock_shared_range(&self, range: Range<u64>) -> io::Result<bool> {
         set_lock(self, false, range)
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_range(&self, range: Range<u64>) -> io::Result<()> {
+        set_lock_blocking(self, true, range)
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_shared_range(&self, range: Range<u64>) -> io::Result<()> {
+        set_lock_blocking(self, false, range)
     }
 
     fn unlock_range(&self, range: Range<u64>) -> io::Result<()> {
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -31,6 +31,8 @@ use core::cmp::{max, min};
 use core::convert::TryInto;
 use core::marker::PhantomData;
 use core::mem;
+#[cfg(feature = "experimental-multiprocess")]
+use core::ops::Range;
 use core::sync::atomic::{AtomicBool, Ordering};
 #[cfg(feature = "logging")]
 use log::warn;
@@ -506,6 +508,27 @@ impl UnpersistedState {
     }
 }
 
+/// The header bytes themselves rather than a byte standing for them, so that Windows' mandatory
+/// locks hold back another process's read of them.
+#[cfg(feature = "experimental-multiprocess")]
+pub(crate) const HEADER_LOCK: Range<u64> = 0..DB_HEADER_SIZE as u64;
+
+/// A hold on the header lock, released when it drops.
+#[cfg(feature = "experimental-multiprocess")]
+struct HeaderGuard<'a> {
+    storage: Option<&'a PagedCachedFile>,
+    _in_process: crate::sync::MutexGuard<'a, ()>,
+}
+
+#[cfg(feature = "experimental-multiprocess")]
+impl Drop for HeaderGuard<'_> {
+    fn drop(&mut self) {
+        if let Some(storage) = self.storage {
+            let _ = storage.unlock_range(HEADER_LOCK);
+        }
+    }
+}
+
 pub(crate) struct TransactionalMemory {
     unpersisted: Mutex<UnpersistedState>,
     storage: PagedCachedFile,
@@ -522,6 +545,14 @@ pub(crate) struct TransactionalMemory {
     // While set, no allocator state is persisted and shutdowns are not recorded as clean,
     // forcing the next open to rebuild the state from the committed trees
     needs_repair: AtomicBool,
+    #[cfg(feature = "experimental-multiprocess")]
+    concurrency_mode: ConcurrencyMode,
+    // Held for the duration of every header file lock. Ensures in-process threads do not overlap
+    // acquiring the lock. Because it's an OFD lock on a single FD, only one may take it at a time.
+    // TODO: This could probably be optimized, so that multiple threads can read at once, by making
+    // this a counter or RwLock
+    #[cfg(feature = "experimental-multiprocess")]
+    in_process_header_lock: Mutex<()>,
     page_size: u32,
     // We store these separately from the layout because they're static, and accessed on the get_page()
     // code path where there is no locking
@@ -624,6 +655,34 @@ impl TransactionalMemory {
         }
     }
 
+    /// Takes the storage and the mutex rather than `&self`, so that the open path can hold the
+    /// header before there is a `TransactionalMemory` to hold it from.
+    #[cfg(feature = "experimental-multiprocess")]
+    fn lock_header<'a>(
+        storage: &'a PagedCachedFile,
+        in_process: &'a Mutex<()>,
+        concurrency_mode: ConcurrencyMode,
+        exclusive: bool,
+    ) -> Result<HeaderGuard<'a>> {
+        let guard = in_process.lock().unwrap();
+        if !concurrency_mode.is_multi_process_writable() {
+            return Ok(HeaderGuard {
+                storage: None,
+                _in_process: guard,
+            });
+        }
+        if exclusive {
+            storage.lock_range(HEADER_LOCK)?;
+        } else {
+            storage.lock_shared_range(HEADER_LOCK)?;
+        }
+
+        Ok(HeaderGuard {
+            storage: Some(storage),
+            _in_process: guard,
+        })
+    }
+
     pub(crate) fn new(
         file: Box<dyn InternalStorageBackend>,
         // Allow initializing a new database in an empty file
@@ -646,11 +705,16 @@ impl TransactionalMemory {
         let storage = PagedCachedFile::new(file, page_size as u64, cache_size);
         // Dropping the storage releases whatever this took, so an open that fails below does too
         Self::lock_for_open(&storage, read_only, concurrency_mode)?;
+        #[cfg(feature = "experimental-multiprocess")]
+        let in_process_header_lock = Mutex::new(());
 
         let initial_storage_len = storage.raw_file_len()?;
 
         let magic_number: [u8; MAGICNUMBER.len()] =
             if initial_storage_len >= MAGICNUMBER.len() as u64 {
+                #[cfg(feature = "experimental-multiprocess")]
+                let _guard =
+                    Self::lock_header(&storage, &in_process_header_lock, concurrency_mode, false)?;
                 storage
                     .read_direct(0, MAGICNUMBER.len())?
                     .try_into()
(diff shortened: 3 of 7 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -986,6 +1115,10 @@ impl TransactionalMemory {
     }
 
     fn write_header(&self, header: &DatabaseHeader) -> Result {
+        // The data pages are written out before this, so the section covers the header bytes
+        // alone and never spans an fsync
+        #[cfg(feature = "experimental-multiprocess")]
+        let _section = self.header_section(true)?;
```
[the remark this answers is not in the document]
cberner: Correct, and it made the whole change a no-op. Fixed in `a328827`.

`PagedCachedFile::write()` fills the write buffer; the bytes reach the file from `flush_write_buffer()`, and also from `flush_lowest_priority()`, which writes a buffered page out under cache pressure -- a second path the finding does not name but which has the same problem. Bracketing `write_header()` covered a memcpy.

I did not take the "write the header directly while the lock is held" option: the commit protocol depends on the header being ordered with the rest of the buffer flush, and 2-phase commit is exactly where that ordering matters. Instead the storage takes the section itself, at each of the three places it touches the header bytes -- the flush, the eviction, and the direct read -- so the section covers the physical write and the ordering is untouched. `page_manager` now brackets nothing.

`the_section_covers_the_flush_that_writes_the_header` holds the header from a second handle and asserts the first one's flush waits. I checked it fails, with "the header reached the file while a reader held the lock", when the flush stops taking the section.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -646,6 +744,10 @@ impl TransactionalMemory {
         let storage = PagedCachedFile::new(file, page_size as u64, cache_size);
         // Dropping the storage releases whatever this took, so an open that fails below does too
         Self::lock_for_open(&storage, read_only, concurrency_mode)?;
+        #[cfg(feature = "experimental-multiprocess")]
+        let shared_mode = !matches!(concurrency_mode, ConcurrencyMode::SingleProcess);
+        #[cfg(feature = "experimental-multiprocess")]
+        let header_sections = Mutex::new(());
```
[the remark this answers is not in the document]
cberner: Right, and fixed in `a328827` as a consequence of the P1 fix rather than separately.

The section is now taken inside `PagedCachedFile::read_direct()`, keyed on the offset, so every read of the header bytes is covered whether or not its caller thought of itself as reading the header. The magic-number probe is a `read_direct(0, ..)` like any other, so it is covered without naming it -- which is the point of moving the bracketing into the storage: a caller can no longer forget.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -506,6 +508,66 @@ impl UnpersistedState {
...
+
+#[cfg(feature = "experimental-multiprocess")]
+impl<'a> HeaderSection<'a> {
+    fn take(
+        storage: &'a PagedCachedFile,
+        serializer: &'a Mutex<()>,
+        shared_mode: bool,
+        exclusive: bool,
+    ) -> Result<Self, DatabaseError> {
+        let guard = serializer.lock().unwrap();
```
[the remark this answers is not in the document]
cberner: The mechanism is right, and the restructure in `a328827` does not close it: the serializer moved from `TransactionalMemory` to `PagedCachedFile`, which is still one per handle, so two handles on descriptors cloned from one description still have separate mutexes and can still overlap.

Leaving it, on the repo owner's standing call: a caller who hands two databases descriptors sharing one open file description has already broken the open negotiation, since the mode bytes cannot exclude a description from itself either -- both opens succeed where they should refuse each other. That is the same `try_clone()` hazard, and the ruling on it was that it is the caller's problem if they do that.

Sharing the serializer across handles would not be enough on its own anyway: it would order the sections, but two handles on one description still cannot exclude each other from the header, because the lock owner is the description. Making that case work needs the open to refuse a description another handle already holds, which is a different change from this one.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -738,6 +854,9 @@ impl TransactionalMemory {
         }
         let (header, _) = unrepaired.finalize(file_len)?;
         if needs_recovery {
+            #[cfg(feature = "experimental-multiprocess")]
+            let _section =
+                Self::open_header_section(&storage, &header_sections, shared_mode, true)?;
```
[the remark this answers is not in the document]
cberner: Real, and still real after `a328827` -- the sections moved into the storage but the open still reads the header in one and writes it in another, so a commit landing in the gap is overwritten.

Not fixing it here, because the fix is a different mechanism rather than a wider section. What makes an open's recovery safe against a live cohort is holding the writer byte across construction, so no other process can commit while this one decides whether to repair. That byte arrives with the concurrency-mode wiring, not with this change.

Widening the section instead does not work cleanly: `read_direct` and the flush each take their own section now, so an outer exclusive hold spanning the read-modify-write would nest, and the in-process serializer is not reentrant -- it would deadlock against itself. Making it reentrant to paper over that seems worse than doing it with the byte that is designed for it.

Worth being explicit that this is not a regression: the shared modes coordinate nothing else yet either -- no allocator, cache or transaction coordination -- so a concurrent commit during another process's open is already unsafe for several reasons. This change makes header I/O atomic; it does not make a multi-step header read-modify-write atomic, and I have not claimed it does.

---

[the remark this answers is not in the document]
cberner: Correcting myself: I was too concessive here. The finished protocol closes this, and by two independent mechanisms rather than the one I named.

The whole open runs under a single exclusive header hold -- `open_writable` takes `exclusive_header_hold()` and keeps it across `Database::new_multiprocess`, so the header read, the recovery decision and the write it leads to are one section, not a read section followed by a write section. There is no gap for a commit to land in.

The writer byte closes it a second time, and in both modes: a multi-writer open takes it across construction, and a single-writer process holds the shared writer byte exclusively for its whole lifetime, so no other writer exists to commit.

So this is not a defect in the design, and it is not something to carry forward as work. What is true is narrower: in this change alone, with none of that wiring present, the two sections are separate. That is a property of an intermediate state in which the shared modes coordinate nothing else either -- no allocator, cache or transaction coordination -- rather than a hazard the finished protocol has.

---

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -140,6 +152,42 @@ mod windows_imp {
...
+        };
+        let ok = unsafe {
+            LockFileEx(
+                file.as_raw_handle(),
+                flags,
+                0,
+                low_dword(len),
+                high_dword(len),
+                &raw mut overlapped,
+            )
```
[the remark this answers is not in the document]
cberner: Confirmed and fixed in `cd5d639`. The structure is heap-allocated now, and on `ERROR_IO_PENDING` the box is leaked and the request refused, so a completion the kernel delivers later lands somewhere that stays valid.

Refusing rather than waiting is deliberate: nothing in the crate opens an asynchronous handle and std cannot, so a caller reaching this has handed `create_file` a handle constructed through `FromRawHandle`, and the rest of the storage's I/O is not written for one either. Waiting for the completion would make the lock work on a handle whose reads and writes still would not.

Not extending this to `try_lock`: `LOCKFILE_FAIL_IMMEDIATELY` is answered rather than queued, and that path predates this change.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -643,9 +643,15 @@ impl TransactionalMemory {
...
 
-        let storage = PagedCachedFile::new(file, page_size as u64, cache_size);
+        #[allow(unused_mut)]
+        let mut storage = PagedCachedFile::new(file, page_size as u64, cache_size);
         // Dropping the storage releases whatever this took, so an open that fails below does too
         Self::lock_for_open(&storage, read_only, concurrency_mode)?;
+        // Only now, so the mode bytes are held before anything locks or reads the header
+        #[cfg(feature = "experimental-multiprocess")]
+        if !matches!(concurrency_mode, ConcurrencyMode::SingleProcess) {
+            storage.enable_header_lock();
```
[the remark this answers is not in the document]
cberner: Same answer as the recovery finding above: the finished protocol closes this, so it is not work to carry forward.

`open_writable` holds one exclusive header hold across the whole of `Database::new_multiprocess`, and initialization happens inside that. A second creator blocks on the hold rather than racing the length check, and by the time it proceeds the file is no longer empty, so it takes the ordinary path instead of the initialization branch. A multi-writer open also holds the writer byte across construction, which excludes the commit the finding has landing in the gap.

What is true is that this change alone does not serialize the decision, because none of that wiring is present in it. That is the same intermediate state the shared modes are in for everything else -- they coordinate no allocator, cache or transactions yet either.

---

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -140,6 +154,54 @@ mod windows_imp {
...
+                &raw mut *overlapped,
+            )
+        };
+        if ok != 0 {
+            return Ok(());
+        }
+
+        let err = io::Error::last_os_error();
+        if err.raw_os_error() == Some(ERROR_IO_PENDING) {
+            Box::leak(overlapped);
```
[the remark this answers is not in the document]
cberner: Right -- leaking the structure stops the corruption but leaves the request live. Fixed in `a3c6ab0`: `CancelIoEx` withdraws it, and the range is unlocked as well in case the grant won the race, since unlocking a range that was never locked fails harmlessly. The structure is still leaked, because the completion may be delivered regardless of the cancel.

Two things worth saying about the shape of this, for the reviewer rather than the bot.

An overlapped handle is not merely unsupported by the lock layer: `seek_read` and `seek_write`, which the backend uses for every page, drive the same overlapped mechanism and would fail the same way. A caller who reaches this has a database whose first read fails. So the point of the fix is not to make that configuration work -- it is to make sure a broken configuration cannot leave a lock held against other processes.

There is a way to eliminate the class rather than handle it: implement the blocking wait as a retry loop over the existing `try_lock`, which passes `LOCKFILE_FAIL_IMMEDIATELY` and so never queues. No pending request, no cancel, no leak, no new Win32 surface. The cost is polling instead of a kernel wait, which is a real downgrade under contention and a call about this database's commit path rather than mine, so I have not made it. Happy to switch if you would rather have the simpler code.

---

src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -140,6 +156,70 @@ mod windows_imp {
...
+            // The completion may still be delivered into the structure, so it outlives us
+            Box::leak(overlapped);
+            return Err(io::Error::other(
+                "file handles opened for asynchronous I/O are not supported",
+            ));
+        }
+
+        Err(err)
+    }
+
```
cberner: Did you check the implementation in std::File?

cberner (commented):
There are way too many comments in the cached_file.rs file. Apply my usual feedback about comments to the whole PR. Also, I was surprised to find the header locking there instead of TransactionalMemory. But maybe PagedCachedFile is the right place for it...hmm. Explain why you put it there and not in TransactionalMemory

src/tree_store/page_store/file_backend/range_lock.rs
cberner: I had not, and it answers this better than what I wrote. Adopted in `8c135cc`.

`std`'s `acquire_lock` creates an event, puts it in the overlapped structure, and on `ERROR_IO_PENDING` calls `GetOverlappedResult` with `bWait` -- so it waits for the pending request rather than treating it as a failure. That is strictly better than the cancel-and-refuse I had: the request is settled before the function returns, so the structure is safe on the stack with nothing leaked and nothing left queued, and a handle opened for asynchronous I/O gets its lock instead of an error.

It also disposes of the alternative I offered you a few comments up. Polling over `try_lock` was only worth considering because the pending case looked awkward to handle; std shows it is not, and the kernel wait stays.

`try_lock` needs none of this either way, for the same reason std's does not: `LOCKFILE_FAIL_IMMEDIATELY` is answered rather than queued.

---

cberner:
Comments cut from 54 lines to 32 in `8c135cc`, `cached_file.rs` from 25 to 11. The duplicated explanation of why a single-process handle takes no section went; so did the field comments restating field names, and the test docs restating test names.

## Why `PagedCachedFile` and not `TransactionalMemory`

I started in `TransactionalMemory`, which is where you expected it, and it was wrong. Worth saying how, because the failure is the argument.

`TransactionalMemory` knows the *logical* header writes: `write_header()`, the recovery rewrite at open, the reload's repair write. So that is where I bracketed. It looked right and did nothing, because `PagedCachedFile::write()` only fills the write buffer. The bytes reach the file later, from `flush_write_buffer()` -- and from `flush_lowest_priority()`, which writes a buffered page out under cache pressure at a moment no caller chose. Every `write_header()` caller flushes afterwards, so the header always hit the disk outside the section. Codex caught it; the first commit of this PR is that mistake.

So the constraint is: the section has to cover the physical I/O, and `TransactionalMemory` does not know when that happens. Only the storage does, because it is the storage that defers the write and later chooses when to issue it.

Three consequences fell out of moving it down, all of which argue for the placement rather than merely permitting it.

The eviction path is covered without anyone thinking about it. A buffered header page evicted under memory pressure is a header write that no caller made and `TransactionalMemory` cannot see.

The reads are covered by offset rather than by intent. `read_direct(0, ..)` is a header read whether or not its caller framed it as one -- the magic-number probe at open is one, and I had missed it; Codex's second finding. Keying on the offset in one place means a caller can no longer forget.

And it removed six bracketing sites in `page_manager.rs` in favour of three in the storage, which is the whole surface: the flush, the eviction, the direct read. There is no fourth way for those bytes to move.

The lock also has to be taken through the same open file description as the I/O it serializes -- on Windows the lock is mandatory, so a lock held on some other description would block this process's own header reads. Taking it through the storage's own backend makes that true by construction. #1413 got this by handing the coordinator an `Arc<File>` and relying on it being the storage's; going through the backend means there is nothing to get wrong, and a custom backend can implement it.

What is lost: `TransactionalMemory` no longer expresses "this is a header write" anywhere, and the storage decides it by offset instead. If you would rather see it stated at the call sites, the way to keep both is a `write_header`/`read_header` pair on `PagedCachedFile` that takes the section and asserts the offset -- the callers name the intent, the storage still owns the timing. Happy to do that if you prefer it; I did not, because the flush and eviction paths still have to key on the offset regardless, so it would add a second way of saying the same thing.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -237,6 +252,31 @@ impl CheckedBackend {
...
+pub(super) struct HeaderSection<'a> {
+    file: Option<&'a CheckedBackend>,
+    _in_process: Option<MutexGuard<'a, ()>>,
+}
+
+#[cfg(feature = "experimental-multiprocess")]
+impl Drop for HeaderSection<'_> {
+    fn drop(&mut self) {
+        if let Some(file) = self.file {
+            let _ = file.unlock_range(HEADER_LOCK);
```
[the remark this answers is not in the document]
cberner: The second half is right and fixed in `8574004`. `unlock_range()` forgot the range before attempting the syscall, so a failed release left the lock held with nothing recorded for `close()` to retry. It now forgets the range only once it is really released.

That ordering is not new in this change -- it is how the range bookkeeping has worked since the whole-storage lock moved into the core. What is new is that it matters: the mode bytes are released once, at close, whereas a header section releases on every read and write of the header, and a reader that failed to release a shared one would block every writer for as long as it stayed open. So it is in scope here even though the line predates this PR.

Not retrying on `EINTR`. The unlock is `F_OFD_SETLK` with `F_UNLCK`, which does not block, so there is nothing for a signal to interrupt -- unlike `F_OFD_SETLKW`, where the blocking acquisition does retry, for exactly that reason. Adding a loop for a wait that cannot happen would suggest one can.

Leaving the `Drop` discarding the error, since `Drop` has nowhere to report it. Retaining the range is what makes that safe rather than silent: the next release attempt is `close()`, which does propagate. Logging it under the `logging` feature, the way `report_failed_release` does for the open paths, would be a reasonable addition -- say the word if you want it.

---

cberner:
Reworked your way in `e226324`. You were right, and the evidence was already in this thread: a few comments up I told the review bot that widening a section was impossible because "`read_direct` and the flush each take their own section, so an outer hold would nest and the serializer is not reentrant". I wrote that as a reason not to fix something. It is really the design telling me it was wrong.

Checking it against the staging branch settles it. The finished protocol has `exclusive_header_hold()` spanning the whole of `Database::new_multiprocess`, `shared_header_hold()` spanning the open-time reads, `set_header_write_lock()` installing per-write bracketing as a hook on `TransactionalMemory`, and a test named `the_id_is_read_and_pinned_under_one_header_hold` -- a read and a pin under one hold. Every one of those is a caller above the storage holding the lock across more than one I/O, which is exactly the "special acquire header lock API" you predicted, and exactly what the per-I/O sections foreclose.

So the lock is back in `TransactionalMemory`. `write_direct()` is the counterpart of `read_direct()`, and it is what makes that work: the write reaches the file where the caller makes it, so a section around the call covers it. The layering is the one you described -- `TransactionalMemory` takes the section, calls `write_direct()`, drops the section; `write_direct()` takes no lock, and `cached_file.rs` brackets nothing.

Two things fell out that are worth flagging.

The pages the header names are flushed before it is written. Writing the header directly otherwise puts it on the file ahead of buffered data pages, and a crash in that window leaves a primary slot naming pages that were never written. `flush_write_buffer` handles that; the fsync still happens outside the section, in the flush the caller already makes.

And a test of mine was weaker than I claimed. `the_section_covers_the_write_reaching_the_file` passes whether the section covers the real write or only a buffered copy, because both block on acquiring the section -- I found this by mutating `write_header` back to the buffered shape and watching the test stay green. `the_header_reaches_the_file_before_write_header_returns` is the missing half: it reads the file around redb with no flush in between, and fails when the write is deferred. I checked that it does.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -847,7 +937,11 @@ impl TransactionalMemory {
         self.storage.invalidate_cache_all();
         self.storage.sync_file()?;
 
-        let header_bytes = self.storage.read_direct(0, DB_HEADER_SIZE)?;
+        let header_bytes = {
+            #[cfg(feature = "experimental-multiprocess")]
+            let _section = self.header_section(false)?;
+            self.storage.read_direct(0, DB_HEADER_SIZE)?
```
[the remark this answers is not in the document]
cberner: Correct, and a regression I introduced when the sections moved back out of the storage: the reload path had its own header write that never went through `write_header()`. Fixed in `947d910` by routing it through `write_header()`, the one place that brackets, so the repair cannot drift back out of the section again.

Not holding one section across the read, the finalize and the write. That is a multi-step read-modify-write, which this change does not make atomic anywhere -- the same answer as the open's recovery path, where the writer byte is what closes it in the finished protocol. What is in scope here is that every header write is bracketed, and now every one is.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -646,11 +718,18 @@ impl TransactionalMemory {
...
+        #[cfg(feature = "experimental-multiprocess")]
+        let header_sections = Mutex::new(());
 
         let initial_storage_len = storage.raw_file_len()?;
 
         let magic_number: [u8; MAGICNUMBER.len()] =
             if initial_storage_len >= MAGICNUMBER.len() as u64 {
+                #[cfg(feature = "experimental-multiprocess")]
+                let _section =
+                    Self::open_header_section(&storage, &header_sections, shared_mode, false)?;
```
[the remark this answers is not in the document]
cberner: Third time for this one, and the answer is the same: the finished protocol closes it. `open_writable` holds a single exclusive header hold across all of `Database::new_multiprocess`, so a second creator blocks before it reaches the length check, and by the time it proceeds the file is no longer empty and it takes the ordinary path.

What did change in `947d910` is narrower than the finding: both initialization writes are now under an exclusive section, so this change no longer leaves a header write unbracketed anywhere. The decision that precedes them is still not serialized here, because the hold that serializes it belongs to the concurrency-mode wiring rather than to this change.

---

src/tree_store/page_store/cached_file.rs
```diff
@@ -446,7 +476,7 @@ impl PagedCachedFile {
         }
     }
 
-    fn flush_write_buffer(&self) -> Result {
+    pub(super) fn flush_write_buffer(&self) -> Result {
```
cberner: Why is this pub(super)??

src/tree_store/page_store/page_manager.rs
```diff
@@ -506,6 +508,31 @@ impl UnpersistedState {
...
+/// One section of the header lock, released when it drops.
+///
+/// The mutex is not redundant with the byte lock. Both sections run on one open file
+/// description, where a second acquisition converts the existing hold and one unlock drops
+/// whatever remains, so two overlapping sections would dismantle each other.
+#[cfg(feature = "experimental-multiprocess")]
+struct HeaderSection<'a> {
+    storage: Option<&'a PagedCachedFile>,
+    _in_process: crate::sync::MutexGuard<'a, ()>,
+}
```
cberner: HeaderSection is a confusing name. Call it HeaderGuard or HeaderLockGuard

src/tree_store/page_store/page_manager.rs
```diff
@@ -522,6 +549,10 @@ pub(crate) struct TransactionalMemory {
     // While set, no allocator state is persisted and shutdowns are not recorded as clean,
     // forcing the next open to rebuild the state from the committed trees
     needs_repair: AtomicBool,
+    #[cfg(feature = "experimental-multiprocess")]
+    shared_mode: bool,
```
cberner: Let's store the ConcurrencyMode instead. We can add pub(crate) methods to it to indicate functionality like this flag

src/tree_store/page_store/page_manager.rs
```diff
@@ -522,6 +549,10 @@ pub(crate) struct TransactionalMemory {
...
+    #[cfg(feature = "experimental-multiprocess")]
+    header_sections: Mutex<()>,
```
cberner: How is this different than the Mutex on `state`?

src/tree_store/page_store/page_manager.rs
```diff
@@ -624,6 +655,47 @@ impl TransactionalMemory {
...
+            &self.storage,
+            &self.header_sections,
+            self.shared_mode,
+            exclusive,
+        )
+        .map_err(|err| match err {
+            DatabaseError::Storage(storage) => storage,
+            other => StorageError::Corrupted(other.to_string()),
+        })
+    }
```
cberner: Let's just have one function. Also the name is bad. Name it lock_header()

src/tree_store/page_store/page_manager.rs
cberner: The `state` mutex is not held at three of the places that take the header lock, so it cannot serialize them.

The commit path is the clearest: it clones the header and drops the state lock on purpose -- "Copy the header so that we can release the state lock, while we flush the file" -- before both `write_header()` calls. Reusing `state` would mean holding it across the header write and the fsync, which is exactly what that line avoids. `clear_cache_and_reload()` and `durable_primary_slot_corrupt()` also take the header without it, and the open path has no `TransactionalMemory` at all yet, so there is no `state` to lock.

The other three writers (`begin_writable()`, `clear_recovery_required()`, `flush_shutdown_header()`) do happen to hold `state`. That is coincidence rather than a rule, and it is not enough: what has to be serialized is the byte-range lock, because both holds run on one open file description, where a second acquisition converts the existing hold and one unlock drops whatever remains. Two overlapping holds would dismantle each other regardless of which other mutex the callers were holding.

I put that reasoning on the field in `bf9be3d`, since the previous comment explained the mechanism but not why `state` does not cover it.

---

src/tree_store/page_store/cached_file.rs
cberner: It shouldn't be. Private again in `bf9be3d`.

It was `pub(super)` because `write_header()` flushed the buffer before writing the header directly. I had that guarding two things, and neither holds up:

Ordering -- getting the pages the header names onto the file first. But redb has never relied on that: in the buffered scheme the header and the pages go out in one `flush_write_buffer()` in arbitrary order, and the OS can write them back in any order after that. The durability boundary is the fsync, and every `write_header()` is still followed by one, with two-phase commit and the slot checksums covering the torn cases as before.

A stale buffered header overwriting the direct write, since `write_direct()` invalidates the read cache but not the write buffer. That is real but unreachable: the only buffered header writes are the two in `new()`, and each flushes inside its own hold, so the buffer never carries header bytes out of the open path.

---

src/tree_store/page_store/page_manager.rs
cberner: `HeaderGuard` in `bf9be3d`. I took the vocabulary with it: the tests and comments now say "hold" where they said "section", so `a_hold_does_not_puncture_the_whole_storage_lock`, `shared_holds_admit_each_other` and so on -- nothing is called a section any more.

---

src/tree_store/page_store/page_manager.rs
cberner: Done in `bf9be3d`. The field is `concurrency_mode: ConcurrencyMode`, and `ConcurrencyMode::is_shared()` is the `pub(crate)` method that answers what the flag used to.

---

src/tree_store/page_store/page_manager.rs
cberner: One `lock_header()` in `bf9be3d`.

It stays an associated function taking the storage and the mutex rather than `&self`, because the open path takes the header before there is a `TransactionalMemory` to take it from -- that split is what the second function existed for. The three `&self` callers now pass `&self.storage, &self.in_process_header_lock, self.concurrency_mode`.

Collapsing them also removed the `map_err` the old pair needed: `lock_header()` returns `StorageError`, so the open path's `DatabaseError` conversion is just `?`.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -986,6 +1078,19 @@ impl TransactionalMemory {
...
+            // Written directly, so that the hold covers the write reaching the file rather than a
+            // buffered copy of it. The fsync stays outside, in the flush that follows.
+            let bytes = header.to_bytes(true);
+            let _guard = Self::lock_header(
+                &self.storage,
+                &self.in_process_header_lock,
+                self.concurrency_mode,
+                true,
+            )?;
+            return self.storage.write_direct(0, &bytes);
```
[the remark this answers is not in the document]
cberner: This is not a real issue. In the full protocol, two phase commit is required

src/tree_store/page_store/cached_file.rs
cberner: Correcting the reasoning I gave above, since the conclusion is right but the argument was not.

I justified dropping the flush on crash-safety grounds -- the fsync is the durability boundary, so header-vs-page order within a flush never mattered. That part is true, but it is not the reason this is safe, and it does not address a peer reading the header live, which is the thing this whole change exists for. The load-bearing reason is the one you gave: the shared modes require two-phase commit, so the flush between the two `write_header()` calls puts the pages on the file before the primary swap is ever published. Without that requirement a 1-phase commit really would publish a primary slot naming buffered pages.

So the ordering does matter here; it is just already provided, by two-phase commit rather than by a flush inside `write_header()`. I have put that in the PR description in place of my crash-safety version.

---

PR #1428, opened 2026-08-31
Lock the active transaction a read transaction is reading

cberner:
The next slice after #1427's writer byte. A shared-mode handle now holds the "active transaction byte" -- `TXN_BASE + t` -- shared for every transaction it is keeping live, so a writer in another process can see the snapshots it must not reclaim.

Rebased onto master now that #1430 has merged, so this is one commit carrying only the lock.

## Where the lock is taken

#1430 gave `live_read_transactions` a single pair of accessors. The byte hangs off exactly those two transitions: taken as a count rises from zero, released as it returns there, both inside `State::reference_transaction()` / `dereference_transaction()` and so under the `state` lock that holds the count. There is no second counter.

Doing it under that lock is what makes it correct rather than merely tidy. Deciding inside the lock and locking outside it leaves a window where the count has reached zero but the byte is still held: another thread registers that id, sees 1 rather than 0, declines to lock, and starts reading -- then the first thread's release lands and the reader is unprotected. That is why the tracker takes the memory to lock through as an argument rather than returning a "reached zero" flag. It stores no reference of its own; every caller already had one.

`TransactionalMemory::lock_mp_transaction()` / `unlock_mp_transaction()` are therefore plain and uncounted. They take the byte under a shared header hold, which is what orders them against a writer's reclamation scan: the scan holds the header exclusively, so it either sees the lock or ran entirely before it.

## What that covers

Every reference in `live_read_transactions`, which is more than read transactions:

* read transactions and read-only handles;
* savepoints, which hold a read transaction through their own `TransactionGuard`;
* a persistent savepoint restored from its record at open;
* the durable ancestor a non-durable commit builds on.

Wiring only the read-transaction pair and leaving the rest sharing the counter would be worse than a separate counter, not better: a reference taken first by one of the others leaves the count at 1, so the read transaction that follows sees no zero-to-one transition and never locks. The other two need protecting from a peer's reclamation in their own right, so this is the right side to err on.

## The guard

`TransactionGuard::Read` carries one field where it used to carry two:

```rust
reference: Option<Arc<TransactionalMemory>>,
```

`Some` means this guard owns a reference and carries what the tracker releases it through; `None` means it owns none -- a savepoint deserialized from its record, or one whose reference the database has taken over. `owns_reference()` is `reference.is_some()`, `release_to_database()` is `reference.take()`, and drop releases through the tracker.

A persistent savepoint's snapshot staying locked after its handle drops now falls out of that rather than being arranged: the handle is dropped inside `persistent_savepoint()` itself while its record still names the snapshot, and taking the reference is exactly what stops the drop from releasing it.

The byte does not nest through one file description -- a second acquisition is absorbed and one release drops whatever remains -- which is why the count decides and the lock is plain.

## Which modes take it

Only the shared ones. A single-process open holds the whole file, which covers these bytes; locking one and releasing it would punch a hole in that lock, the same hazard as the header lock and the writer byte.

## What is not here

Nothing consumes these locks. The scan that reads them -- and the reclamation bound it feeds -- follows separately, because it only makes sense with its consumer. A database opened in a shared mode still is not safe to actually share: the reader takes its id from this handle's cached header rather than rereading it under the hold, which the transaction-level reload closes.

Separately, a `ReadTransaction` outliving its `Database` loses this byte, because `Database::drop` defers only for a live write transaction and `FileBackend::close()` releases the whole range. That is a master bug rather than this PR's -- `SHARED_WRITER_BYTE` goes the same way -- and the open thread on this PR has the reproduction and a proposed fix to `Database::drop`.

## What this replaces

This branch previously carried a `Coordinator` in a new `src/multiprocess.rs`, holding its own file handle and its own copies of the header lock, `HEADER_LOCK`, `LOCK_BASE` and `byte_range`. It predated the header lock landing in `TransactionalMemory`, and could not work against master: `FileBackend` owns a plain `File`, so nothing could hand the coordinator the description the storage does its I/O through, and two header-lock implementations with two serializers over one description would dismantle each other's holds.

Rewritten onto the merged layering, which also removed:

* a second `Query` enum threaded through `RangeLock`, `InternalStorageBackend` and both backends, so the Unix scan could descend by conflict offset. Bisection is correct on both platforms and is what Windows already did, so one path replaces two and `query_lock_range` stays a `bool`. `range_lock.rs` and `optimized.rs` are untouched by this PR now.
* a `redb_multiprocess` cfg computed in `build.rs`. Nothing needs it once the code lives beside the rest of the protocol.

## Testing

Eight tests in `db::active_transaction_test`, probing the byte from a second file description, which conflicts with the database's exactly as another process's would:

* `a_read_transaction_locks_the_id_it_reads_until_it_ends`: no byte held before, exactly one while the transaction is open, none after.
* `concurrent_readers_of_one_snapshot_share_the_lock`: two readers, one byte; still held when the first ends, released when the second does.
* `a_read_only_handle_locks_what_it_reads`.
* `an_ephemeral_savepoint_locks_the_snapshot_it_references`.
* `a_persistent_savepoint_keeps_its_lock_after_its_handle_drops`: still held after `persistent_savepoint()` returns.
* `reopening_locks_a_persistent_savepoints_snapshot`: nothing held once the database closes, one byte held again once it reopens.
* `a_non_durable_commit_locks_its_durable_ancestor`, and the durable commit that flushes it releases the ancestor.
* `a_single_process_read_does_not_puncture_the_whole_file_lock`.

Mutation-tested rather than assumed, seven mutations, each caught by the test that names it: never locking; never releasing; locking and releasing on every reference instead of at zero (caught by the shared-lock test); `release_to_database()` keeping the reference (caught by the persistent-savepoint test); removing both single-process gates; and registering a persistent savepoint or a non-durable commit without locking -- the narrow version above -- caught by the two tests for those paths.

One correction worth recording: my first draft of `a_non_durable_commit_locks_its_durable_ancestor` asserted that nothing was held after the following durable commit. It is not: that commit's epilogue takes a reference on its own id, cleared by the commit after it. The test asserts what is actually true -- the ancestor is released -- rather than what I assumed.

`cargo fmt`, the justfile's `cargo clippy --all-targets --all-features` and `cargo clippy --all --all-targets --all-features` as CI runs them, clippy with no features, cross-clippy for `x86_64-pc-windows-msvc` and `aarch64-apple-darwin`, the four `wasm32` variants, both `thumbv7em-none-eabihf` no_std configurations CI checks, `cargo build --release` with the feature, and `cargo test` in all three configurations that run tests -- all pass.

Also run for `x86_64-pc-windows-gnu` under wine, where all eight tests pass. The suite's one failure there, `a_whole_file_lock_holder_reads_as_already_open`, fails identically on master.

---

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -99,6 +99,10 @@ pub(crate) const SHARED_WRITER_BYTE: u64 = LOCK_BASE + 1;
 pub(crate) const SHARED_READER_BYTE: u64 = LOCK_BASE + 2;
 #[cfg(feature = "experimental-multiprocess")]
 pub(crate) const IMMUTABLE_READER_BYTE: u64 = LOCK_BASE + 3;
+/// Base of the "active transaction range": a handle reading transaction `t` holds `TXN_BASE + t`
+/// shared for as long as it is reading it
+#[cfg(feature = "experimental-multiprocess")]
+pub(crate) const TXN_BASE: u64 = LOCK_BASE + 1024;
 
 pub(crate) fn byte_range(offset: u64) -> Range<u64> {
     offset..offset + 1
@@ -349,9 +353,10 @@ pub(crate) enum TransactionGuard {
     Read {
         tracker: Arc<TransactionTracker>,
         transaction_id: TransactionId,
-        // Cleared when a savepoint becomes persistent: the database record owns the
-        // transaction's reference from then on, and releases it when the savepoint is deleted
-        owns_reference: bool,
+        // `Some` while this guard owns the transaction's reference, carrying what the tracker
+        // needs to release it. `None` when a savepoint becomes persistent: the database
+        // owns the reference from then on and releases it when the savepoint is deleted
+        reference: Option<Arc<TransactionalMemory>>,
     },
     Write {
         tracker: Arc<TransactionTracker>,
@@ -372,11 +377,12 @@ impl TransactionGuard {
     pub(crate) fn new_read(
         transaction_id: TransactionId,
         tracker: Arc<TransactionTracker>,
+        mem: &Arc<TransactionalMemory>,
     ) -> Self {
         Self::Read {
             tracker,
             transaction_id,
-            owns_reference: true,
+            reference: Some(mem.clone()),
         }
     }
 
@@ -388,7 +394,7 @@ impl TransactionGuard {
         Self::Read {
             tracker,
             transaction_id,
-            owns_reference: false,
+            reference: None,
         }
     }
 
@@ -396,14 +402,14 @@ impl TransactionGuard {
     // reference count.
     // Dropping this guard then releases nothing.
     pub(crate) fn release_to_database(&mut self) {
-        let Self::Read { owns_reference, .. } = self else {
+        let Self::Read { reference, .. } = self else {
             unreachable!("only a read transaction's reference may be leaked")
         };
-        *owns_reference = false;
+        drop(reference.take());
     }
 
     pub(crate) fn owns_reference(&self) -> bool {
-        matches!(self, Self::Read { owns_reference, .. } if *owns_reference)
+        matches!(self, Self::Read { reference, .. } if reference.is_some())
     }
 
     pub(crate) fn tracker(&self) -> &Arc<TransactionTracker> {
@@ -415,10 +421,11 @@ impl TransactionGuard {
 
     pub(crate) fn allocate_read(
         tracker: Arc<TransactionTracker>,
-        mem: &TransactionalMemory,
+        mem: &Arc<TransactionalMemory>,
     ) -> Result<Self> {
         let id = tracker.register_read_transaction(mem)?;
-        Ok(Self::new_read(id, tracker))
+
+        Ok(Self::new_read(id, tracker, mem))
     }
 
     pub(crate) fn new_write(
@@ -469,10 +476,10 @@ impl Drop for TransactionGuard {
             Self::Read {
                 tracker,
                 transaction_id,
-                owns_reference,
+                reference,
             } => {
-                if *owns_reference {
-                    tracker.deallocate_read_transaction(*transaction_id);
+                if let Some(mem) = reference {
+                    tracker.deallocate_read_transaction(mem, *transaction_id);
                 }
             }
             Self::Write {
@@ -566,13 +573,9 @@ impl Sealed for ReadOnlyDatabase {}
 #[cfg(not(redb_no_std))]
 impl ReadableDatabase for ReadOnlyDatabase {
     fn begin_read(&self) -> Result<ReadTransaction, TransactionError> {
-        let id = self
-            .transaction_tracker
-            .register_read_transaction(&self.mem)?;
+        let guard = TransactionGuard::allocate_read(self.transaction_tracker.clone(), &self.mem)?;
         #[cfg(feature = "logging")]
-        debug!("Beginning read transaction id={id:?}");
-
-        let guard = TransactionGuard::new_read(id, self.transaction_tracker.clone());
+        debug!("Beginning read transaction id={:?}", guard.id());
 
         ReadTransaction::new(self.mem.clone(), guard)
     }
@@ -1362,7 +1365,7 @@ impl Database {
                 },
             };
             db.transaction_tracker
-                .register_persistent_savepoint(&savepoint);
+                .register_persistent_savepoint(&db.mem, &savepoint)?;
         }
         txn.abort()?;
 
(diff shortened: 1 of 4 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -567,6 +567,11 @@ pub(crate) struct TransactionalMemory {
     // this a counter or RwLock
     #[cfg(feature = "experimental-multiprocess")]
     in_process_header_lock: Mutex<()>,
+    // How many of this handle's transactions are reading each pinned id. The byte does not nest
+    // through one file description -- a second acquisition is absorbed and one release drops it
+    // -- so the lock is taken as a count rises from zero and released as it returns there.
+    #[cfg(feature = "experimental-multiprocess")]
+    pins: Mutex<BTreeMap<u64, u64>>,
```
cberner: Let's make the key a TransactionID

src/db.rs
```diff
@@ -432,7 +453,15 @@ impl Drop for TransactionGuard {
...
-            } => tracker.deallocate_read_transaction(*transaction_id),
+                #[cfg(feature = "experimental-multiprocess")]
+                mem,
+            } => {
+                #[cfg(feature = "experimental-multiprocess")]
+                if let Some(mem) = mem {
+                    let _ = mem.unpin_transaction(*transaction_id);
+                }
+                tracker.deallocate_read_transaction(*transaction_id);
+            }
```
cberner: It seems a bit unfortunate that these are separate. Is there an easy way to get the TransactionalMemory into the txn tracker so it can do both?

cberner: Agreed it is unfortunate, and looking into it the answer is better than "tidier" -- but it changes a lifetime, so I want your call before doing it.

Passing the memory per call, mirroring `register_read_transaction(&self, mem)`, does not work: `Savepoint::drop` calls `deallocate_savepoint()` holding only the tracker, so there is nowhere for it to come from. Owning it in the tracker is the way.

That also closes a gap I had missed. Savepoints go through the same pair -- `allocate_savepoint()` calls `register_read_transaction()` and `deallocate_savepoint()` calls `deallocate_read_transaction()` -- so a savepoint holds a read-transaction reference and its snapshot is exactly what a peer must not reclaim. My version pins from `TransactionGuard`, which savepoints never build, so they pin nothing. Pairing it in the tracker fixes that for free, and the `pinned` flag on the guard disappears too, since the tracker can roll back its own registration when the pin fails.

Two costs:

* `TransactionTracker::new()` gains a `TransactionalMemory`, and two unit tests construct a tracker where none exists -- `savepoint::test::corrupted_record_errors` and `transaction_tracker::test::non_durable_commit_without_freed_pages_is_not_unprocessed`. Both would need to build one over an in-memory backend, or the field becomes an `Option`, which I would rather avoid.
* A live `Savepoint` holds the tracker, so it would then transitively keep the `TransactionalMemory` alive. Today it does not. That changes when the storage is closed, which is why I have not just done it.

Happy to make the change if the second point is fine by you, or to leave savepoint pinning to the slice that consumes the pins.

---

src/tree_store/page_store/page_manager.rs
cberner: Done in `eca2601`. `TransactionId` already derives `Ord`, so the map keys directly and the `raw_id()` calls in `pin_transaction()` and `unpin_transaction()` are gone.

---

src/db.rs
```diff
@@ -530,7 +559,12 @@ impl ReadableDatabase for ReadOnlyDatabase {
         #[cfg(feature = "logging")]
         debug!("Beginning read transaction id={id:?}");
 
-        let guard = TransactionGuard::new_read(id, self.transaction_tracker.clone());
+        let guard = TransactionGuard::new_read(
+            id,
+            self.transaction_tracker.clone(),
+            #[cfg(feature = "experimental-multiprocess")]
+            None,
```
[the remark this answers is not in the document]
cberner: Real, and fixed in `e5ac5dd`. `ReadOnlyDatabase::begin_read()` now goes through `TransactionGuard::allocate_read()` like the writable one, so it pins and the duplicated construction is gone.

My comment on that `None` claimed it was "a read guard the database itself holds, outside any transaction". That was simply wrong -- I misread the site. It is a user-facing read transaction, and a read-only handle is the participant that most needs to pin: in a single-writer cohort the readers are read-only opens.

`a_read_only_handle_pins_what_it_reads` covers it, and I checked it discriminates: restoring the hand-built guard fails that test alone.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -712,6 +717,82 @@ impl TransactionalMemory {
...
+            false,
+        )?;
+        let id = read_id()?;
+        let byte = Self::transaction_byte(id)?;
+        if let Some(count) = pins.get_mut(&id.raw_id()) {
+            *count += 1;
+            return Ok(id);
+        }
+        // Refused only by an exclusive holder, which the shared header hold excludes
+        if !self.storage.try_lock_shared_range(byte_range(byte))? {
```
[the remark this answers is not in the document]
cberner: Real, and fixed in `e5ac5dd`. Same shape as the writer-byte slot leak on #1427: the in-process registration must be owned before the cross-process lock is attempted.

`pin_transaction()` no longer takes a closure -- it takes an id. `allocate_read()` registers, builds the `TransactionGuard` that owns the registration, then pins into it, so a failure drops a guard that deallocates on the way out. A `pinned` flag keeps the guard from unpinning what it never pinned.

Note this drops the property that the id is read inside the same header hold as the pin. That property was not doing anything: the id comes from this handle's in-memory state, not the header, which is the substance of the finding below. When the reload lands it will read the id from the header under the hold and this will move back inside it.

---

src/db.rs
```diff
@@ -369,19 +377,32 @@ impl TransactionGuard {
...
     }
 
     pub(crate) fn allocate_read(
         tracker: Arc<TransactionTracker>,
-        mem: &TransactionalMemory,
+        mem: &Arc<TransactionalMemory>,
     ) -> Result<Self> {
+        // Registered under the pin, so the id pinned is the one this transaction will read
+        #[cfg(feature = "experimental-multiprocess")]
+        let id = mem.pin_transaction(|| tracker.register_read_transaction(mem))?;
```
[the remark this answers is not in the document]
cberner: Right, and I checked it against `docs/design.md` rather than reasoning from the code alone. "Active transaction set" says the reader "acquires a shared header lock, reads the latest committed transaction id, and then while still holding the header lock, acquires a shared lock on the active transaction byte", and "All other processes must not cache the header". The id here comes from `state.latest_slot()`, which is exactly the cached header the doc forbids.

Not fixed, because the fix is the reload rather than a wider hold. Reading the id from the file under the hold on its own would make things worse, not better: the `ReadTransaction` takes its roots from the same in-memory state, so it would pin one snapshot and read another. What is needed is to reread the header under the hold, adopt its transaction slots, and invalidate the cache -- the "Cache invalidation" section -- and that is the transaction-level reload this PR defers.

What I have corrected is the claim. The doc comment on `pin_transaction()` said the hold established that the pinned id was still published; it does not, and now says so, pointing at the design doc section. So this is not silently wrong, it is explicitly not done yet.

Worth being plain about the consequence: a shared-mode reader can pin an id whose pages a peer has already reused, so the pins are not yet load-bearing. That is the same "not safe to actually share" the PR describes, and the reload is what closes it.

---

src/db.rs
cberner: I tried it and hit an obstacle I had not seen when I asked, so I have backed it out and left the branch at `eca2601`. The blocker is bigger than the lifetime point you approved, so it needs your call again.

`Database::check_integrity()` uses `Arc::get_mut(&mut self.mem)` twice: once as its "is any transaction alive?" test, and once for real, to get the `&mut TransactionalMemory` that `clear_cache_and_reload()` requires. Putting an `Arc<TransactionalMemory>` in the tracker makes both fail -- five tests go red with `TransactionInProgress`, among them `repair_commit_reserves_transaction_id` and `check_integrity_clears_leak_latch`.

I then tried a `Weak`, expecting the reference count to be the whole problem. It is not: `Arc::get_mut` requires the weak count to be zero as well, so it still returned `None` with `strong=1` -- which I only established by instrumenting the branch rather than reasoning about it. So the tracker cannot hold a reference of either kind.

The ways out that I can see, none of which I want to take unilaterally:

* Split the liveness test from the mutable access: `Arc::strong_count(&self.mem) == 1` for the first, which ignores weak refs, and something other than `Arc::get_mut` for `clear_cache_and_reload()`. That means making that method work through `&self`, which is a change to storage internals rather than to this PR.
* Give `Savepoint` the memory directly and unpin in its own `Drop`. Same wall: a live persistent savepoint would then hold a strong reference and break `check_integrity()` the same way. Ephemeral ones are already refused there, persistent ones are not.
* Leave it. The branch keeps pinning in `TransactionGuard`, savepoints do not pin, and the slice that adds the reclamation scan closes it -- that slice has to touch this area anyway, and by then it will be clear whether the scan wants the tracker to own the pins.

I lean towards the third for this PR and the first when the scan lands, but it is your call.

---

cberner: Done, via the route you picked: the prep PR gave `Savepoint` a `TransactionGuard` (merged as `b28cda8`), and this branch is now rebased on it with the pinning consolidated into that guard.

So the pair you flagged is no longer split. `TransactionGuard::allocate_read()` registers, builds the guard, and pins into it -- one path for every holder of a snapshot, savepoints included, and the tracker never needs the `TransactionalMemory` at all. `allocate_savepoint()` returns the guard rather than a bare `TransactionId`, and `Savepoint::drop` releases the transaction by dropping it.

That also sidesteps the `Arc::get_mut` wall entirely: nothing new holds a reference to the memory except the guards, which are exactly the things `check_integrity()` already means to detect.

One deliberate leak, which I have written up in the PR body rather than buried here. `release_to_database()` now hands the pin over along with the reference, because a persistent savepoint's handle is dropped inside `persistent_savepoint()` while its record still names the snapshot. Nothing releases that pin yet, and a persistent savepoint rebuilt from its record after a reopen is not pinned at all. Releasing on deletion and re-pinning at open both want the scan, so they come with it -- and until then the error is over-pinning, which is the safe direction.

`an_ephemeral_savepoint_pins_the_snapshot_it_references` and `a_persistent_savepoint_keeps_its_pin_after_its_handle_drops` cover the two halves. I checked they discriminate: putting `allocate_savepoint()` back on `register_read_transaction()` + `new_read()` fails exactly those two, and deleting the `pinned.take()` fails exactly the persistent one.

---

src/db.rs
```diff
@@ -352,6 +356,12 @@ pub(crate) enum TransactionGuard {
         // Cleared when a savepoint becomes persistent: the database record owns the
         // transaction's reference from then on, and releases it when the savepoint is deleted
         owns_reference: bool,
+        // Set once the pin is taken, naming what releases it. None where this guard does not
```
cberner: I don't like the "pin" terminology. Call it multi-process transaction lock or something like that

src/tree_store/page_store/page_manager.rs
```diff
@@ -712,6 +717,82 @@ impl TransactionalMemory {
...
+    /// Taken under a shared header hold, which is what orders it against a writer's reclamation
+    /// scan: the scan holds the header exclusively, so it either sees this pin or ran entirely
+    /// before it.
+    ///
+    /// The caller must guarantee `id` cannot be collected while this runs. Reading it from the
+    /// header under this same hold is what will guarantee that -- see `docs/design.md`, "Active
+    /// transaction set" -- and is not yet done: the id a caller passes is this handle's cached
+    /// view, which a peer's commits can leave behind.
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn pin_transaction(&self, id: TransactionId) -> Result {
```
cberner: See other comment. I don't like the "pin" terminology. The design doc doesn't use that. Let's call it "increment_active_multiprocess_read_transaction()". I know it's long, but unless you have a better idea I don't

cberner (commented):
Do the usual trimming of comments

src/db.rs
cberner: Done in `4ebff82`. The field is `multi_process_transaction_lock`, which reads alongside `multi_process_writer` on the `Write` variant.

The rest of the file follows the design doc rather than my term: `TXN_BASE` is documented as the base of the "active transaction range", the test module is `active_transaction_test`, and the tests are named for locking rather than pinning. No "pin" is left in anything this PR touches.

Comments trimmed in the same commit -- the `pinned` field comment, `release_to_database()`, `increment_active_multiprocess_read_transaction()`'s docs, `allocate_savepoint()`, and the test module's doc. What I kept is the reasoning that is not in the code: the header hold's ordering against the reclamation scan, why release takes no hold, why single-process mode locks nothing, and the "not done yet" on rereading the id.

---

src/tree_store/page_store/page_manager.rs
cberner: Done in `4ebff82`, with your name and its symmetric partner: `increment_active_multiprocess_read_transaction()` and `decrement_active_multiprocess_read_transaction()`. No better idea -- long, but it says what it does and matches the doc.

Two more renames fell out of it, both taking terms the design doc already uses: the map is `active_multiprocess_read_transactions`, and `transaction_byte()` is `active_transaction_byte()`, which is what "Active transaction set" calls it. The `Corrupted` message says "active transaction byte" now too.

---

cberner: Renamed to `increment_active_mp_transaction()` / `decrement_active_mp_transaction()` in `4d3a4a5`. The map follows as `active_mp_transactions`.

Dropping `read` matters beyond the length: `docs/design.md` says write transactions take the same shared lock on the transaction they are reading, so the name would have gone stale the moment the write path takes its own hold.

The signature now fits on one line, where the old one wrapped across four.

`multi_process_transaction_lock` on the guard stays spelled out -- it sits directly under `multi_process_writer` on the `Write` variant, and there is no length problem there.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -567,6 +567,10 @@ pub(crate) struct TransactionalMemory {
     // this a counter or RwLock
     #[cfg(feature = "experimental-multiprocess")]
     in_process_header_lock: Mutex<()>,
+    // How many of this handle's transactions are reading each id. The byte does not nest through
+    // one file description, so it is locked as a count rises from zero and unlocked as it returns
+    #[cfg(feature = "experimental-multiprocess")]
+    active_mp_transactions: Mutex<BTreeMap<TransactionId, u64>>,
```
cberner: Hang on. Isn't this a duplicate of `live_read_transactions` in TransactionTracker? I don't like that. Maybe we can have `deallocate_read_transaction()` return a bool indicating whether it reached a count of zero.

Similarly, register_read() could handle locking the active mp txn slot. And then we could rename the methods on TransactionalMemory as lock_mp_transaction() and unlock_mp_transaction()

src/db.rs
```diff
@@ -352,6 +356,10 @@ pub(crate) enum TransactionGuard {
         // Cleared when a savepoint becomes persistent: the database record owns the
         // transaction's reference from then on, and releases it when the savepoint is deleted
         owns_reference: bool,
+        // Set once the "active transaction byte" is locked, naming what releases it. None where
+        // this guard releases nothing: a single-process open, or a lock given to the database
+        #[cfg(feature = "experimental-multiprocess")]
+        multi_process_transaction_lock: Option<Arc<TransactionalMemory>>,
```
cberner: I think it's very confusing that this is named "_lock" but then the type is a optional TransactionalMemory. Either we need to wrap that in a guard, or maybe it can just be renamed "mem: Arc<TransactionalMemmory>" after doing the other refactoring of counters that I suggested with the tracker

src/tree_store/page_store/page_manager.rs
cberner: You are right, it was the same counter twice. Done in `0087862`: the map is gone, `lock_mp_transaction()` / `unlock_mp_transaction()` are plain and uncounted, and `live_read_transactions` is the only count.

I went further than the bool, though, because the bool has a race. Two private helpers on `State` do it:

```rust
fn reference_transaction(&mut self, mem: &TransactionalMemory, id: TransactionId) -> Result;
fn dereference_transaction(&mut self, mem: &TransactionalMemory, id: TransactionId);
```

The byte is taken as a count rises from zero and released as it returns there, **under the same `state` lock that holds the count**. Returning a bool and locking in the caller means the count reaches zero, `state` is released, and the byte is still held: another thread can register that id, see a count of 1 rather than 0, decline to lock, and start reading -- then the first thread's unlock lands and the reader is unprotected. So the tracker needs the memory to lock through, and `deallocate_read_transaction(mem, id)` takes it. Every caller already had it: `TransactionGuard::drop` through the `Arc` it holds, and `deallocate_savepoint()` from `WriteTransaction`.

The bigger consequence of unifying: `live_read_transactions` has **four** mutation sites, not two, so all four now lock. Besides `register_read_transaction()` / `deallocate_read_transaction()`:

* `register_non_durable_commit()` / `clear_pending_non_durable_commits()` -- the durable ancestor a non-durable commit builds on.
* `register_persistent_savepoint()` -- a persistent savepoint restored at open.

Wiring only the read-transaction pair would have been worse than the separate counter, not better: a reference taken first by one of the others leaves the count at 1, so the reader that follows sees no zero-to-one transition and never locks. Both of those genuinely need protecting from a peer's reclamation anyway, so this is the right side to err on.

It also closes the gap the PR body used to describe: **a persistent savepoint reopened from its record is now locked**, where before it was not. `reopening_locks_a_persistent_savepoints_snapshot` covers it and `a_non_durable_commit_locks_its_durable_ancestor` covers the other, both mutation-checked against exactly the narrow version above.

Two smaller things fell out:

* Codex's rollback finding is handled inside `reference_transaction()` now: it owns the count, so a failed lock undoes its own increment. `allocate_read()` is back to two lines.
* `register_non_durable_commit()` and `register_persistent_savepoint()` became fallible. `TransactionTracker`'s unit test now builds an `InMemoryBackend` memory, opened `SingleProcess` where the lock is a no-op, following the pattern in `btree.rs`.

---

src/db.rs
cberner: Agreed, and the counter refactor made it go away rather than just get renamed. In `0087862` the variant is:

```rust
Read {
    // Some while this guard owns the transaction's reference, carrying what the tracker
    // needs to release it. Taken when a savepoint becomes persistent: the database record
    // owns the reference from then on, and releases it when the savepoint is deleted
    reference: Option<Arc<TransactionalMemory>>,
}
```

That is one field where there were two: `owns_reference: bool` is gone, because the `Option` now answers exactly the question it asked. `Some` means this guard owns a reference and carries what the tracker releases it through; `None` means it owns none. `owns_reference()` is `reference.is_some()`, `release_to_database()` is `reference.take()`, and drop is:

```rust
if let Some(mem) = reference {
    tracker.deallocate_read_transaction(mem, *transaction_id);
}
```

The persistent-savepoint leak I had to arrange by hand now falls out: taking the reference is what stops the drop from releasing it, so there is no second piece of state to keep in step.

It is not `Arc<TransactionalMemory>` unconditionally, because `new_read_unowned()` -- the guard for a savepoint deserialized from its record -- has no memory to hold and owns nothing to release. The field is also no longer behind `#[cfg(experimental-multiprocess)]`, since the tracker takes the memory in both builds; the whole suite passes without the feature, so `check_integrity()`'s `Arc::get_mut` is not disturbed by it.

---

src/db.rs
```diff
@@ -372,11 +377,12 @@ impl TransactionGuard {
     pub(crate) fn new_read(
         transaction_id: TransactionId,
         tracker: Arc<TransactionTracker>,
+        mem: &Arc<TransactionalMemory>,
     ) -> Self {
         Self::Read {
             tracker,
             transaction_id,
-            owns_reference: true,
+            reference: Some(mem.clone()),
```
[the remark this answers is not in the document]
cberner: Real, and I reproduced it rather than reasoning from the code. But it is not this PR's, and the fix does not belong here.

Probing from a second file description, with a `ReadTransaction` held across the drop:

```
before:                      txn=[4]  shared_writer=true
after db drop, reader alive: txn=[]   shared_writer=false
```

So the active transaction byte does vanish under a live reader, exactly as described. `FileBackend::close()` is `unlock_range(FULL_RANGE)`, and `Database::drop` defers only for a live *write* transaction (`defer_close_if_write_transaction_live`), so everything goes at once.

Which is the point: `SHARED_WRITER_BYTE` goes with it, and that is on master today. A `ReadTransaction` outliving its `Database` already leaves this process holding no protocol state at all while it is still reading -- announcing "no writer here" to a peer that is about to open. This PR adds one more byte to what is released early; it does not create the hazard, and it is no more load-bearing than the mode byte is, since nothing consumes either yet.

The fix is to `Database::drop`, not to anything in this diff, and it changes when the file closes for every mode. Roughly:

* defer on `any_user_read_reference_exists()` as well -- that predicate already exists and already excludes the internal refs that only hold durable ancestors of pending non-durable commits, which must not keep the database open;
* have `deallocate_read_transaction()` return the deferred `Arc<TransactionalMemory>` when the last user reference goes, so `TransactionGuard::drop`'s `Read` arm runs `close_database()` the way the `Write` arm already does.

That is a change to redb's close semantics that fixes a master bug, so it wants its own PR and @cberner's call on whether a `Database` dropped under live readers should stay open. Widening this one to carry it would bury it.

---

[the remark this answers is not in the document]
cberner: @cberner ruled: the close should still happen, and the read transaction should raise the existing `DatabaseClosed` error. Fixed in #1431, on master rather than here.

Investigating it turned up why the error was not already firing, which is narrower than I expected. `deferred_close_invalidates_read_transactions` asserts this exact behavior today -- it only passes because it sets `cache_size(0)`. The check lives at the backend, so a read that reaches the file does fail; `PagedCachedFile::read()` serves a warm page and returns before ever getting there.

So the fix is one `check_io_errors()?` ahead of the caches, and the reader now errors instead of reading a snapshot whose locks are gone. That covers the mode bytes as well as the active transaction byte, which is the reason it belongs on master.

---

cberner (commented):
lgtm

PR #1430, opened 2026-09-01
Refactor read transaction reference counting

cberner:
Split out of #1428 at your request: the refactoring parts, with no behavior change and nothing multi-process about them. #1428 rebases onto this and keeps only the lock.

## The tracker's refcount

`live_read_transactions` was mutated at five sites, each hand-rolling the same arithmetic -- three incrementing with `entry().and_modify().or_insert(1)`, two decrementing and removing at zero. `State::reference_transaction()` and `State::dereference_transaction()` are now the only places that touch the map, and the five callers are one line each.

That is worth doing on its own, and it is also what the multi-process work needs: the "active transaction byte" hangs off exactly these two transitions, so there needs to be one place to hook rather than five.

Worth noting what the five sites are, since I had assumed two:

* `register_read_transaction()` / `deallocate_read_transaction()` -- read transactions, read-only handles, and savepoints.
* `register_non_durable_commit()` / `clear_pending_non_durable_commits()` -- the durable ancestor a non-durable commit builds on.
* `register_persistent_savepoint()` -- a persistent savepoint restored from its record at open.

## The savepoint constructor

`Savepoint::new_ephemeral()` took a tracker and a transaction id and built its own `TransactionGuard` internally. It takes the guard instead, and `allocate_savepoint()` builds it through `TransactionGuard::allocate_read()` -- the same path every other holder of a read transaction uses, rather than a second way of getting one.

Fewer arguments, so the `clippy::too_many_arguments` exemption goes too.

## Testing

No new tests: nothing here is a new behavior, and the existing suite covers the paths -- `deallocate_savepoint()`, the non-durable commit refcount, and persistent savepoint create/delete/restore all run through the two helpers now.

`cargo fmt`, `cargo clippy --all --all-targets --all-features` and with no features as CI runs them, cross-clippy for `x86_64-pc-windows-msvc` and `aarch64-apple-darwin`, the `wasm32` variants, both `thumbv7em-none-eabihf` no_std configurations, `cargo build --release`, and `cargo test` in all three configurations that run tests -- all pass.

---

### DIFF
--- a/src/transaction_tracker.rs
+++ b/src/transaction_tracker.rs
@@ -99,6 +99,25 @@ struct State {
     deferred_close: Option<Arc<TransactionalMemory>>,
 }
 
+impl State {
+    // The reference count on `id` is what keeps the pages its snapshot reaches from being freed,
+    // so every holder takes one here and gives it back here
+    fn reference_transaction(&mut self, id: TransactionId) {
+        self.live_read_transactions
+            .entry(id)
+            .and_modify(|x| *x += 1)
+            .or_insert(1);
+    }
+
+    fn dereference_transaction(&mut self, id: TransactionId) {
+        let count = self.live_read_transactions.get_mut(&id).unwrap();
+        *count -= 1;
+        if *count == 0 {
+            self.live_read_transactions.remove(&id);
+        }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


cberner (commented):
lgtm

PR #1431, opened 2026-09-01
Fail cached reads once the database is closed

cberner:
Your ruling from the #1428 thread: dropping a `Database` under a live read should still close it, and the read transaction should then raise the `DatabaseClosed` error that already exists.

It does not today, and the reason is narrow.

## What was wrong

`deferred_close_invalidates_read_transactions` already asserts exactly that behavior. It passes only because it sets `cache_size(0)`.

The check lives at the backend, in `CheckedBackend::read()` -> `check_failure()`, so a read that reaches the file after close does fail. But `PagedCachedFile::read()` serves a warm page out of the write buffer or read cache and returns before it ever gets there. With a normal cache, a read transaction that outlives its handle keeps reading a snapshot out of memory after `FileBackend::close()` has released every byte-range lock the process held.

Measured on the branch before the fix, holding a `ReadTransaction` across `drop(db)`:

```
before drop: get("hello") = Ok(Some("world"))
after drop:  get("hello") = Ok(Some("world"))   // should be DatabaseClosed
```

## The fix

One `check_io_errors()?` at the top of `PagedCachedFile::read()`, ahead of the caches.

That is an acquire load on the hottest read path, so it is worth being explicit about the cost: it sits immediately before a mutex or striped-`RwLock` acquisition the same function already makes, and it is the same latch every read that reaches the file already pays. I did not benchmark it -- say the word if you want numbers before this lands.

The alternative I considered and rejected was clearing the caches in `close()`, which would keep the read path untouched: a read would then miss and get the error from the backend. It is racy -- a read already in flight can repopulate a cache entry after the clear, and a later read hits it -- so the latch check is the one that actually holds.

## Testing

`a_warm_cache_does_not_outlive_the_database` in `basic_tests.rs`, next to the existing test, doing the same thing with a real cache: read once so every page the later read needs is warm, drop the database, then assert `DatabaseClosed` from both `open_table()` and `list_tables()`.

Mutation-checked: removing the added line fails that test and only that test.

Found by Codex on #1428, which flagged the multi-process consequence -- the active transaction byte disappears while the reader is still live. This is the general case underneath it, and it is a master bug rather than that PR's: `SHARED_WRITER_BYTE` goes the same way today.

`cargo fmt`, `cargo clippy --all --all-targets --all-features` and with no features as CI runs them, cross-clippy for `x86_64-pc-windows-msvc` and `aarch64-apple-darwin`, the `wasm32` variants, both `thumbv7em-none-eabihf` no_std configurations, `cargo build --release`, and `cargo test` in all three configurations that run tests -- all pass.

No public API changes. The behavior change is that a read after close now errors where it previously returned stale data, which is the documented contract rather than a new one, so no changelog entry -- tell me if you would rather have one.

---

### DIFF
--- a/src/tree_store/page_store/cached_file.rs
+++ b/src/tree_store/page_store/cached_file.rs
@@ -608,6 +608,10 @@ impl PagedCachedFile {
     // Read with caching. Caller must not read overlapping ranges without first calling invalidate_cache().
     // Doing so will not cause UB, but is a logic error.
     pub(super) fn read(&self, offset: u64, len: usize, hint: PageHint) -> Result<Arc<[u8]>> {
+        // Before the caches, which would otherwise serve a page the file can no longer be read
+        // for: a read transaction outliving its Database must see DatabaseClosed, not a snapshot
+        // of whatever happened to be cached
+        self.check_io_errors()?;
         debug_assert_eq!(0, offset % self.page_size);
         #[cfg(feature = "cache_metrics")]
         self.reads_total.fetch_add(1, Ordering::AcqRel);
--- a/tests/basic_tests.rs
+++ b/tests/basic_tests.rs
@@ -194,6 +194,38 @@ fn deferred_close_invalidates_read_transactions() {
     ));
 }
 
+/// The guarantee above has to hold with a cache too: a warm page must not be served after the
+/// database is closed, or a read transaction outliving its handle keeps reading a snapshot no
+/// lock protects.
+#[test]
+fn a_warm_cache_does_not_outlive_the_database() {
+    let tmpfile = create_tempfile();
+    let db = Database::create(tmpfile.path()).unwrap();
+    let setup_txn = db.begin_write().unwrap();
+    {
+        let mut table = setup_txn.open_table(STR_TABLE).unwrap();
+        table.insert("hello", "world").unwrap();
+    }
+    setup_txn.commit().unwrap();
+
+    let read_txn = db.begin_read().unwrap();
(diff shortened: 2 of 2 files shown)


cberner:
Closing: @cberner has said serving cached data after the close is fine, and only new reads must fail, without paying for it on the read path. **Master already does exactly that**, so there is nothing to change.

* A new read after the close already fails. `PagedCachedFile::read()` misses, falls through to the checked backend, and `check_failure()` returns `DatabaseClosed`. That is what `deferred_close_invalidates_read_transactions` covers -- its `cache_size(0)` is not a workaround, it is the case being tested.
* A read served from a warm page or an already-resolved B-tree root returns data. Under the requirement as stated, that is allowed.

So the premise this PR was built on was wrong. I read the existing test's `cache_size(0)` as hiding a gap, when it was pinning down the guarantee that actually applies.

Codex's P1 on this PR was correct and I confirmed it: `Btree::get()`, `first()`, `last()` short-circuit on `cached_root`, and `len()` answers from the cached header, so the check I added never ran for a table opened before the close. Fixing it needed checks in four `Btree` methods plus `ReadOnlyMultimapTable::len()`, all mutation-verified. That is exactly the surface that would need to be audited and kept in step to make the guarantee exact -- and it is the cost being declined, correctly.

The multi-process consequence I raised on #1428 stands on its own terms and is unaffected: the byte disappears at close, and any read that still needs the file fails, so nothing reads a snapshot whose locks are gone without an error.

---

PR #1438, opened 2026-09-03
Let a read-only handle open a database another process is writing

cberner:
The second of three pieces cut from #1432. #1437 (forced 2-phase commits), #1439 (the `Durability::None` refusal) and #1440 (the refactor that moves the header read into `read_header()`) are merged, so this is based on master and its own change to that function is the shared reader's branch; stacked above it are #1432 (the per-read reload), #1433 and #1413.

## What it fixes

A read-only open refuses a file whose recovery flag is set, and a writable handle sets that flag for as long as it is open, so a database another process was writing could not be opened read-only at all.

## What it does

Opening such a file is a different job from opening a file this process owns, so `read_shared_header()` does it separately rather than threading the difference through the owned path:

```rust
let (header_bytes, consistent) = {
    let _guard = Self::lock_header(storage, in_process_header_lock, concurrency_mode, false)?;
    (
        storage.read_direct(0, DB_HEADER_SIZE)?,
        storage.query_lock_range(byte_range(CONSISTENT_BYTE))?,
    )
};
let unrepaired = UnrepairedDatabaseHeader::from_bytes(&header_bytes, page_size.try_into().unwrap())?;
if unrepaired.unclean() && !consistent {
    return Err(DatabaseError::RepairAborted);
}
Ok(unrepaired.finalize_transaction_slots()?)
```

The only refusal is an unclean file with no writer holding the consistent byte from #1436: a writer publishes the recovery flag under its exclusive hold and holds the byte only while that flag means a live writer rather than an unrepaired file. Such a handle repairs nothing, loads no allocator state, and consults neither the layout nor the file's length. It never allocates, and the read path addresses pages through the immutable geometry alone (the static region size and region header size, not the region counts), so torn counts never reach it, and a page the file no longer holds fails at the read rather than at the open. This replaces the layout and length checks earlier revisions carried.

One platform caveat. On Windows a probe answers by taking the range it queries, so two readers opening at the same instant can each read the other's probe as a live writer and admit an unclean file that no writer holds. Such a reader still reads correctly: a primary that verifies is a commit whose pages were flushed before the header that published it, 2-phase or not, and a torn one is refused. The probes cannot be serialized under an exclusive header hold: a read-only descriptor cannot take one on POSIX, and `a_read_only_handle_takes_shared_holds` pins that.

It still reads the header once, at open. Following the writer's later commits is #1432.

1. **The primary slot is taken as recorded.** `finalize_transaction_slots()` never chooses the newer slot the way a writer's `finalize()` does. Choosing between the slots is a repairing writer's job: a newer secondary a shared reader can see is a commit whose pages have not reached the file yet, or one a repair has rolled back and republished with the recovery flag already clear. This came out of a Codex finding on the reload path, and the open path ran the same selection. The alternative is a writer-side change to what a repair republishes.
2. **The refusal error.** An unclean file with no writer, and a writer mid-repair, return `RepairAborted`. A distinct variant meaning "retry" is the alternative.
3. **Merging while cross-process pins are not yet consumed.** Before this PR no reader could be open beside a shared writer, so the unconsumed active transaction bytes from #1428 were unreachable. After it, a writer can reclaim pages a reader in another process is reading, until the scan slice from #1413 lands. Either merge behind the experimental flag with that stated, or hold this until the scan is out.

## Testing

`a_shared_read_refuses_an_unclean_file_with_no_writer`; `a_repairing_writer_does_not_admit_a_shared_reader`, where a shared reader probing from inside `do_repair()` is refused and admitted once the repair is done, with the panic it prevents in #1436's sweep; `a_read_only_participant_takes_the_primary_as_recorded`, where a newer slot whose pages the file never received is spliced into the secondary under the god byte a repair leaves and the id the reader locks is the primary's; and `a_shared_reader_ignores_the_layout`, which tears the region counts to zero under a live writer and reads through them. The timing smoke test and the four truncation and torn-layout tests are gone with the checks they covered. The full local check set passes.

The `experimental-multiprocess` changelog bullet gains the read-only-open sentence.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -533,6 +533,11 @@ pub trait ReadableDatabase: SealedInApi5 {
 ///
 /// Multiple processes may open a [`ReadOnlyDatabase`], but it may not be opened concurrently
 /// with a [`Database`].
+#[cfg_attr(
+    feature = "experimental-multiprocess",
+    doc = "",
+    doc = "The multi-process concurrency modes are the exception: a reader shares the file with the writer. See [`Builder::set_concurrency_mode`](crate::Builder::set_concurrency_mode)."
+)]
 ///
 /// # Examples
 ///
@@ -620,17 +625,24 @@ impl ReadOnlyDatabase {
             concurrency_mode,
         )?;
         let mem = Arc::new(mem);
-        // If the last transaction used 2-phase commit and updated the allocator state table, then
-        // we can just load the allocator state from there. Otherwise, we need a full repair
-        if let Some(tree) = Database::get_allocator_state_table(&mem)? {
-            mem.load_allocator_state(&tree)?;
-        } else {
-            #[cfg(feature = "logging")]
-            warn!(
-                "Database {:?} not shutdown cleanly. Repair required",
-                &file_path
-            );
-            return Err(DatabaseError::RepairAborted);
+        // A reader beside a multi-process writer never allocates, so it loads no allocator state
+        #[cfg(feature = "experimental-multiprocess")]
+        let multiprocess_writer = concurrency_mode.is_multi_process_writable();
+        #[cfg(not(feature = "experimental-multiprocess"))]
+        let multiprocess_writer = false;
+        if !multiprocess_writer {
+            // If the last transaction used 2-phase commit and updated the allocator state table, then
+            // we can just load the allocator state from there. Otherwise, we need a full repair
+            if let Some(tree) = Database::get_allocator_state_table(&mem)? {
+                mem.load_allocator_state(&tree)?;
+            } else {
+                #[cfg(feature = "logging")]
+                warn!(
+                    "Database {:?} not shutdown cleanly. Repair required",
+                    &file_path
+                );
+                return Err(DatabaseError::RepairAborted);
+            }
         }
 
         let next_transaction_id = mem.get_last_committed_transaction_id()?.next();
@@ -1728,6 +1740,11 @@ impl Builder {
     /// If the file has been opened for writing (i.e. as a [`Database`]) [`DatabaseError::DatabaseAlreadyOpen`]
     /// will be returned on platforms which support file locks (macOS, Windows, Linux). On other platforms,
     /// the caller MUST avoid calling this method when the database is open for writing.
+    #[cfg_attr(
+        feature = "experimental-multiprocess",
+        doc = "",
+        doc = "Only a single-process writer is refused: in the multi-process concurrency modes a reader opened in the same mode as the writer shares the file with it. See [`Self::set_concurrency_mode`]."
+    )]
     #[cfg(not(redb_no_std))]
     pub fn open_read_only(
         &self,
@@ -2477,8 +2494,8 @@ mod active_transaction_test {
     use std::path::Path;
 
     const TABLE: TableDefinition<u64, u64> = TableDefinition::new("x");
-    // The ids a freshly created database can have committed
-    const SEARCHED: std::ops::Range<u64> = 0..8;
+    // The ids a freshly created database can have committed, with room for a few more commits
+    const SEARCHED: std::ops::Range<u64> = 0..16;
 
     fn create(path: &Path, mode: ConcurrencyMode) -> Database {
         let mut builder = Database::builder();
@@ -2673,6 +2690,66 @@ mod active_transaction_test {
         ));
     }
 
+    /// A read-only participant takes the primary as recorded: choosing between the slots is a
+    /// repairing writer's job, and a newer secondary it can see is a commit whose pages are not in
+    /// the file yet, or one a repair has just rolled back
+    #[test]
+    fn a_read_only_participant_takes_the_primary_as_recorded() {
+        use std::io::{Read, Write};
+
+        let tmpfile = crate::create_tempfile();
+        drop(create(tmpfile.path(), ConcurrencyMode::SingleProcess));
+
+        let mut builder = Database::builder();
+        builder.set_concurrency_mode(ConcurrencyMode::MultiWriterProcess);
+
+        // A newer commit whose pages this file never received: made in a copy, then its slot
+        // spliced into this file's secondary slot
+        let copy = crate::create_tempfile();
+        std::fs::copy(tmpfile.path(), copy.path()).unwrap();
+        let db = Database::open(copy.path()).unwrap();
+        let write = db.begin_write().unwrap();
+        {
+            let mut table = write.open_table(TABLE).unwrap();
+            table.insert(0, 1).unwrap();
+        }
+        write.commit().unwrap();
+        drop(db);
+
+        let header = |path: &Path| -> Vec<u8> {
+            let mut header = vec![0u8; 320];
+            File::open(path).unwrap().read_exact(&mut header).unwrap();
+            header
+        };
+        let slot = |index: usize| 64 + index * 128..64 + (index + 1) * 128;
+        let id = |header: &[u8], index: usize| {
+            u64::from_le_bytes(header[slot(index)][104..112].try_into().unwrap())
+        };
+        let newer = header(copy.path());
+        let mut spliced = header(tmpfile.path());
+        let primary = usize::from(spliced[9] & 1);
+        spliced[slot(1 - primary)].copy_from_slice(&newer[slot(usize::from(newer[9] & 1))]);
+        // The god byte a repair leaves on a 1-phase history: recovery clear, 2-phase clear
+        spliced[9] &= !6;
+        OpenOptions::new()
+            .write(true)
+            .open(tmpfile.path())
+            .unwrap()
+            .write_all(&spliced)
+            .unwrap();
+        assert!(id(&spliced, 1 - primary) > id(&spliced, primary));
+        let probe = probe(tmpfile.path());
+
+        let opened = builder.open_read_only(tmpfile.path()).unwrap();
+        let read = opened.begin_read().unwrap();
+        assert_eq!(
+            held_ids(&probe),
+            vec![id(&spliced, primary)],
+            "the reader adopted a slot whose pages the file never received"
+        );
+        drop(read);
+    }
+
     /// Sharing the file forces 2-phase, whatever the transaction asked for
     #[test]
     fn a_shared_commit_is_two_phase() {
@@ -2691,6 +2768,37 @@ mod active_transaction_test {
             db.mem.used_two_phase_commit(),
             "a shared commit published without the flush between the header writes"
         );
+    }
+
+    /// With no writer holding the file, an unclean one is refused: nothing is coming to repair it
+    #[test]
+    fn a_shared_read_refuses_an_unclean_file_with_no_writer() {
+        use std::io::{Read, Seek, SeekFrom, Write};
+
+        let tmpfile = crate::create_tempfile();
+        drop(create(tmpfile.path(), ConcurrencyMode::MultiWriterProcess));
+
+        // The god byte's recovery-required bit, which a clean shutdown clears
+        {
+            let mut file = OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(tmpfile.path())
+                .unwrap();
+            let mut byte = [0u8; 1];
+            file.seek(SeekFrom::Start(9)).unwrap();
+            file.read_exact(&mut byte).unwrap();
+            byte[0] |= 2;
+            file.seek(SeekFrom::Start(9)).unwrap();
+            file.write_all(&byte).unwrap();
+        }
+
+        let mut builder = Database::builder();
+        builder.set_concurrency_mode(ConcurrencyMode::MultiWriterProcess);
+        assert!(matches!(
+            builder.open_read_only(tmpfile.path()),
+            Err(crate::DatabaseError::RepairAborted)
+        ));
     }
 
     /// A single-process open holds the whole file, which covers these bytes. Locking one and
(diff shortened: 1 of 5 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -699,6 +699,82 @@ impl TransactionalMemory {
...
+                storage.query_lock_range(byte_range(CONSISTENT_BYTE))?,
+            )
+        };
+        let unrepaired =
+            UnrepairedDatabaseHeader::from_bytes(&header_bytes, page_size.try_into().unwrap())?;
+        // A short file is nobody's to repair; an unclean one is a live writer's, if there is one
+        if unrepaired.truncated(file_len) || (unrepaired.unclean() && !consistent) {
+            return Err(DatabaseError::RepairAborted);
+        }
+        Ok(unrepaired.finalize_transaction_slots()?)
```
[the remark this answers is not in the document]
cberner: By design for this piece, and the PR body says so under "What it does": such a handle reads the header once, at open. Following the writer's later commits is the next PR up the stack, #1432, where `latest_committed_transaction_id()` rereads the commit slots under the caller's header hold on every read transaction, and #1433 drops the cached pages when the id moves. #1432 was split into this PR, #1437 and itself today so that each carries few decisions; this one is the open path only.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -699,6 +699,82 @@ impl TransactionalMemory {
...
+    fn read_header(
+        storage: &PagedCachedFile,
+        #[cfg(feature = "experimental-multiprocess")] in_process_header_lock: &Mutex<()>,
+        #[cfg(feature = "experimental-multiprocess")] concurrency_mode: ConcurrencyMode,
+        page_size: usize,
+        read_only: bool,
+    ) -> Result<DatabaseHeader, DatabaseError> {
+        #[cfg(feature = "experimental-multiprocess")]
+        if read_only && concurrency_mode.is_multi_process_writable() {
+            return Self::read_shared_header(
```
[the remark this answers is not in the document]
cberner: Correct on the facts, and it is decision 4 in the PR body, put to the maintainer rather than fixed here: before this PR no reader could be open beside a shared writer, so the unconsumed active transaction bytes from #1428 were unreachable, and after it a writer can reclaim pages a reader in another process is reading until the scan lands. The scan is `TransactionTracker::cross_process_free_floor()` in #1413, which bounds `durable_commit()`'s reclamation by the oldest transaction any process pins, with `a_readers_pin_survives_heavy_reclamation` as its test; it is the next slice out. Whether this merges behind the experimental flag ahead of that, with the gap stated, or waits for the scan, is cberner's call.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -699,6 +699,82 @@ impl TransactionalMemory {
...
+                storage.read_direct(0, DB_HEADER_SIZE)?,
+                storage.raw_file_len()?,
+                storage.query_lock_range(byte_range(CONSISTENT_BYTE))?,
+            )
+        };
+        let unrepaired =
+            UnrepairedDatabaseHeader::from_bytes(&header_bytes, page_size.try_into().unwrap())?;
+        // A short file is nobody's to repair; an unclean one is a live writer's, if there is one
+        if unrepaired.truncated(file_len) || (unrepaired.unclean() && !consistent) {
+            return Err(DatabaseError::RepairAborted);
```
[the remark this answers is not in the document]
cberner: Real, and fixed in `9f6e6e0` by checking in the order you suggest: `(unclean() && !consistent) || truncated(file_len)`, so the layout is read only once the unclean-with-no-writer refusal has not fired. `from_bytes()` validates the region counts only on a clean header, and a writer's repair is what reconciles torn ones, so for a shared reader the layout is trustworthy exactly when the file is clean or a writer holds the consistent byte -- which is the condition that now guards it.

`a_shared_read_refuses_an_unclean_file_with_no_writer` now also tears the region counts to zero. On the old order it panicked with `attempt to subtract with overflow` at `layout.rs:223`; it returns `RepairAborted` on the new.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -699,6 +699,83 @@ impl TransactionalMemory {
...
+                storage.read_direct(0, DB_HEADER_SIZE)?,
+                storage.raw_file_len()?,
+                storage.query_lock_range(byte_range(CONSISTENT_BYTE))?,
+            )
+        };
+        let unrepaired =
+            UnrepairedDatabaseHeader::from_bytes(&header_bytes, page_size.try_into().unwrap())?;
+        // An unclean file is a live writer's, if there is one, and its layout is unvalidated until
+        // that writer has reconciled it; a short file is nobody's to repair
+        if (unrepaired.unclean() && !consistent) || unrepaired.truncated(file_len) {
```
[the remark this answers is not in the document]
cberner: Real, and fixed on the current head `487b0c5`: `layout_readable()` runs the two count checks that `from_bytes()` applies only to a clean header, and the reader refuses a header that fails them whether or not a writer holds the consistent byte. `a_live_writer_does_not_excuse_a_torn_layout` tears the counts under a live writer holding the byte; it panicked at `layout.rs:223` on the old check and returns `RepairAborted` on the new.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -699,6 +699,86 @@ impl TransactionalMemory {
...
+                storage.query_lock_range(byte_range(CONSISTENT_BYTE))?,
+            )
+        };
+        let unrepaired =
+            UnrepairedDatabaseHeader::from_bytes(&header_bytes, page_size.try_into().unwrap())?;
+        // An unclean file is a live writer's, if there is one; a short file is nobody's to repair.
+        // The counts an unclean header leaves unvalidated are checked before the layout is read
+        if (unrepaired.unclean() && !consistent)
+            || !unrepaired.layout_readable()
+            || unrepaired.truncated(file_len)
```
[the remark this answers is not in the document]
cberner: Admitted, yes; harmed, no. `do_repair()` rolls the primary back only when `primary_verifies()` fails, and a clean header means the last commit completed before the close, so a clean file's primary verifies unless the file was damaged from outside. What a clean-file repair does is rebuild the allocator state and commit it on top of the primary as recorded, with the data root unchanged, so a reader admitted meanwhile reads a tree the repair keeps.

Run locally against this head, not committed: a clean file with its 2-phase bit cleared, so the writer's open goes through `do_repair()`, and a shared reader opened from the repair callback at progress 0.0, 0.6 and 0.9 was admitted each time and read the committed value, as did one opened after the writer's open returned. Requiring the consistent byte whenever `SHARED_WRITER_BYTE` is held would refuse readers for the whole of every such open, for no gain.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -699,6 +699,81 @@ impl TransactionalMemory {
...
+                Self::lock_header(storage, in_process_header_lock, concurrency_mode, false)?;
+            (
+                storage.read_direct(0, DB_HEADER_SIZE)?,
+                storage.query_lock_range(byte_range(CONSISTENT_BYTE))?,
+            )
+        };
+        let unrepaired =
+            UnrepairedDatabaseHeader::from_bytes(&header_bytes, page_size.try_into().unwrap())?;
+        // An unclean file is a live writer's to repair, if there is one
+        if unrepaired.unclean() && !consistent {
```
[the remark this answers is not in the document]
cberner: Not fixed, by decision: cberner asked for the length and layout checks to go, and this head is that. A shared reader never allocates, so it never consults the layout, and the read path addresses pages through the immutable geometry alone; a page the file no longer holds fails at the read with an I/O error rather than at the open. Refusing at open only moved that error earlier for a file damaged from outside, which a writer's own open repairs and a reader never does. Truncation that removes only unused tail pages is harmless to a reader by construction, since no committed tree references those pages. `raw_file_len()` is not read here any more because nothing the reader does depends on it.

Also, `d9b30c9` is not an object in this repository (`git cat-file -t d9b30c9` fails), so whatever tree that refers to is not this PR's.

---

src/db.rs
```diff
@@ -620,17 +625,24 @@ impl ReadOnlyDatabase {
...
-                "Database {:?} not shutdown cleanly. Repair required",
-                &file_path
-            );
-            return Err(DatabaseError::RepairAborted);
+        // A shared reader never allocates, so it has no allocator state to load
+        #[cfg(feature = "experimental-multiprocess")]
+        let shared_read = concurrency_mode.is_multi_process_writable();
+        #[cfg(not(feature = "experimental-multiprocess"))]
+        let shared_read = false;
+        if !shared_read {
```
cberner: I think "shared_read" is confusing. Because it could be a group of read-only processes using Immutable mode. How about "multiprocess_writer"? That better matches the flag anyway

src/tree_store/page_store/header.rs
```diff
@@ -334,6 +340,19 @@ impl UnrepairedDatabaseHeader {
...
+    // rolled back.
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(super) fn finalize_transaction_slots(self) -> Result<DatabaseHeader> {
+        if self.primary_corrupted {
+            return Err(StorageError::Corrupted(
+                "Primary commit slot is corrupted".to_string(),
+            ));
+        }
+        Ok(self.inner)
+    }
```
cberner, line 355: Hmm, this is only called in one place. We should inline it, unless it is going to be used more broadly in future PRs

src/tree_store/page_store/header.rs
```diff
@@ -213,6 +213,12 @@ impl UnrepairedDatabaseHeader {
         self.inner.recovery_required || self.inner.layout().len() != file_len
     }
 
+    // The last writer did not shut down cleanly
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(super) fn unclean(&self) -> bool {
+        self.inner.recovery_required
+    }
+
```
cberner, line 221: recovery_required is pub(super), so I think we can remove this and let the caller just use it?

src/tree_store/page_store/page_manager.rs
```diff
@@ -895,33 +970,17 @@ impl TransactionalMemory {
...
+            &in_process_header_lock,
             #[cfg(feature = "experimental-multiprocess")]
-            let _guard =
-                Self::lock_header(&storage, &in_process_header_lock, concurrency_mode, true)?;
-            storage
-                .write(0, DB_HEADER_SIZE, true)?
-                .mem_mut()
-                .copy_from_slice(&header.to_bytes(true));
-            storage.flush()?;
-        }
```
cberner: Let's do a prep PR that just does the refactoring, without any behavior change, of moving this code into read_header()

src/tree_store/page_store/header.rs
cberner: It is called again in the next PR up: `reload_transaction_slots()` in #1432 adopts the slots a peer has written through it on every read transaction. Kept for that; say the word if you would rather it be inlined here and pulled out again there.

---

src/tree_store/page_store/header.rs
cberner: The field is `DatabaseHeader`'s, and at that point the reader holds an `UnrepairedDatabaseHeader`, whose `inner` is private. The check also has to come before `finalize_transaction_slots()`: a corrupted primary on an unclean file with no writer is a torn 1-phase commit for the next writer's repair to roll back, so it wants `RepairAborted` rather than the `Corrupted` that method returns. Exposing `inner`, or checking after the finalize and reporting that case as `Corrupted`, are the alternatives; either is fine by me if you prefer it to the accessor.

---

cberner: oh right. nm, you can ignore my comment

src/tree_store/page_store/header.rs
cberner: got it, ok we can keep it then

src/tree_store/page_store/page_manager.rs
cberner: Done: #1440 moves that block into `read_header()` with no behavior change, and this PR is rebased onto it (efe2d83), so what it adds to that function is the shared reader's branch, plus `read_shared_header()` itself.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -737,6 +746,34 @@ impl TransactionalMemory {
...
+        in_process_header_lock: &Mutex<()>,
+        concurrency_mode: ConcurrencyMode,
+        page_size: usize,
+    ) -> Result<DatabaseHeader, DatabaseError> {
+        let (header_bytes, consistent) = {
+            let _guard =
+                Self::lock_header(storage, in_process_header_lock, concurrency_mode, false)?;
+            (
+                storage.read_direct(0, DB_HEADER_SIZE)?,
+                storage.query_lock_range(byte_range(CONSISTENT_BYTE))?,
```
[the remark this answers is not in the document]
cberner: The race is real and is the platform caveat in the PR body, where the one fix, an exclusive header hold around the probe, is ruled out: a read-only descriptor cannot take one on POSIX, and `a_read_only_handle_takes_shared_holds` pins that. Probing `SHARED_WRITER_BYTE` as well does not close it: a read-only handle holds `SHARED_READER_BYTE`, not that byte, so the second probe is the same Windows emulation, and two readers running the same sequence are fooled on both bytes by the same interleaving.

It also changes nothing about what the admitted reader reads. The primary is taken as recorded: a slot that verifies is a commit whose pages were flushed before the header that published it, 1-phase or 2-phase, and a torn one is refused by `finalize_transaction_slots()` as `Corrupted`. So a reader admitted by a peer's probe reads a committed snapshot; what it skips is a repair it would never run anyway. The body's caveat now says so for 1-phase files too.

---

PR #1440, opened 2026-09-03
Move the header read out of TransactionalMemory::new()

cberner:
A prep PR for #1438, asked for in review there: the block in `TransactionalMemory::new()` that reads the header, reconciles its layout against the file's length and writes back a repair becomes `read_header()`, and the assertion that followed it moves with it. No behavior change: the moved code differs only in taking `storage` and the header lock by reference.

#1438 rebases onto this, leaving it the shared reader's branch from `read_header()` and `read_shared_header()` itself.

## Testing

Refactor only, so the existing suite covers it. The full local check set passes.

---

### DIFF
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -699,6 +699,44 @@ impl TransactionalMemory {
         })
     }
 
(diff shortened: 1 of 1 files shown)


cberner (commented):
lgtm. I'll merge it once CI is green

PR #1441, opened 2026-09-03
Bound reclamation by the transactions other processes are reading

cberner:
The scan slice from #1413, cut as its own change. It closes the gap the merged reader slices (#1438, #1432, #1433) flag in their bodies: a read transaction in another process pins its transaction id by holding a byte in the active transaction range (#1428), but no writer consulted those bytes, so its snapshot could be reclaimed and reused under it.

Compaction is refused while the file is shared in #1442, cut on master rather than folded in here: without it, `compact()` beside a reading peer would spin in the drain it begins with once this floor holds, which Codex found on this PR. Merge #1442 first, or rebase this onto it.

## What it does

A durable commit's reclamation floor is now the oldest transaction still being read in any process, rather than in this one:

```rust
let free_until_transaction = self
    .mem
    .oldest_active_transaction(self.transaction_tracker.oldest_local_read_transaction())?
    .map_or(self.transaction_id, |x| x.next());
```

`TransactionalMemory::oldest_active_transaction()` takes this process's oldest read and returns the older of it and a peer's oldest pin. It looks for that pin under the exclusive header hold, which every registration waits behind, so no pin lands below what it reports, and reads the ceiling, the last committed transaction, under the same hold. Only a byte below this process's own oldest read can lower the answer, so only that range is probed; that is also why this process's own locks, which a probe on Windows sees, never come into it. The bytes cannot be listed, only probed over a range, so `active_transactions::lowest_held()` bisects. Each scan starts from what the one before it found: a pin only ever lands at or above the ceiling of the scan before it, since every handle reads the id it pins from the file, so the remembered id is probed alone first. A lingering reader costs one probe, an idle file two, and a reader that moved up about `log2` of the distance.

The post-commit free pass is skipped where the file is shared. It runs after publication, where a peer's pin on the transaction just superseded can land below any floor the scan produced; those pages are reclaimed by the next durable commit instead.

1. **Bisection on every platform, rather than descending along the conflict offsets `F_OFD_GETLK` reports.** #1413 did the latter on POSIX and bisected only on Windows, which cannot report an offset. With the remembered start the common cases are one or two probes on either, each a syscall that costs nothing next to the commit's fsync. One algorithm, no `Query` type on the lock trait, no platform branch, and the storage's existing boolean probe is enough.
2. **The scan is bounded by this process's oldest read, rather than stepping over its own pins.** #1413 handed the scan the set of this process's read ids to skip, since a Windows probe sees its own locks. Nothing at or above the local oldest read can lower the floor, so the scan stops below it, and the set, the step-over and the sampling race between the two are gone.
3. **The post-commit pass waits, rather than rescanning.** A second exclusive hold and scan after publication would let it run; skipping it costs a shared writer one transaction of delay before it reuses the pages the previous one freed.
4. **Both shared modes scan, and both remember where the last scan stopped.** A `SingleWriterProcess` writer has no writing peers, but its read-only peers pin all the same. The remembered start relies on a pin never landing below the previous scan's ceiling, which holds because every handle reads the id it pins from the file; a writable multi-writer handle that has not refreshed is unusable whatever the floor until the writer-side reload in #1413, so the mode makes no difference here.

## Testing

`a_readers_pin_survives_heavy_reclamation`: twenty rounds overwriting all 128 keys under a live reader on a second open file description, asserting the pinned snapshot still reads its own values, and that a fresh read after the pin is dropped sees the last round. Mutation-checked: without the foreign floor it fails at the pinned read, and so does it with the post-commit pass left on. The scan itself has `the_scan_finds_the_lowest_held_byte_from_its_start`, which also checks that nothing below the start is looked at, and `the_scan_finds_the_oldest_of_many`, over raw file descriptions holding bytes at random ids. `the_oldest_active_transaction_is_the_lower_of_ours_and_a_peers` drives `oldest_active_transaction()` through a peer's pin on a second description: the pin bounds the floor while it is older than this process's read, and this process's read does once it is released. The full local check set passes.

---

### DIFF
--- a/src/transaction_tracker.rs
+++ b/src/transaction_tracker.rs
@@ -469,7 +469,7 @@ impl TransactionTracker {
             .map(|(id, txn_id)| (*id, *txn_id))
     }
 
-    pub(crate) fn oldest_live_read_transaction(&self) -> Option<TransactionId> {
+    pub(crate) fn oldest_local_read_transaction(&self) -> Option<TransactionId> {
         self.state
             .lock()
             .unwrap()
--- a/src/transactions.rs
+++ b/src/transactions.rs
@@ -2003,8 +2003,8 @@ impl WriteTransaction {
         }
 
         let free_until_transaction = self
-            .transaction_tracker
-            .oldest_live_read_transaction()
+            .mem
+            .oldest_active_transaction(self.transaction_tracker.oldest_local_read_transaction())?
             .map_or(self.transaction_id, |x| x.next());
         self.process_freed_pages(free_until_transaction)?;
         // Flush allocated pages (including previously unpersisted allocations that are now
@@ -2087,7 +2087,14 @@ impl WriteTransaction {
 
         self.apply_savepoint_state_on_commit();
 
-        if self.post_commit_free == PostCommitFree::Enabled {
+        // The post-commit pass runs after publication, where a peer's pin on the transaction just
+        // superseded lands below any floor the scan above produced, so a multi-process writer's
+        // pages wait for the next durable commit
+        #[cfg(feature = "experimental-multiprocess")]
+        let multiprocess_writer = self.mem.concurrency_mode().is_multi_process_writable();
+        #[cfg(not(feature = "experimental-multiprocess"))]
+        let multiprocess_writer = false;
+        if self.post_commit_free == PostCommitFree::Enabled && !multiprocess_writer {
             self.process_data_freed_pages_after_commit(
                 user_root,
                 &page_allocator,
@@ -2107,10 +2114,10 @@ impl WriteTransaction {
         let epilogue_transaction = self.transaction_id.next();
         let mut free_until = self
             .transaction_tracker
-            .oldest_live_read_transaction()
+            .oldest_local_read_transaction()
             .map_or(epilogue_transaction, |x| x.next());
         // Clamp the free horizon to the savepoint horizon captured during the purge. A savepoint
-        // is also a live read, so absent concurrency `oldest_live_read_transaction()` never
+        // is also a live read, so absent concurrency `oldest_local_read_transaction()` never
         // exceeds it and this is a no-op. But an ephemeral `Savepoint::drop` racing this commit
         // (legal since `WriteTransaction: Sync`) can land between the purge and here, advancing
         // the oldest live read past the savepoint the purge kept entries for. Freeing those pages
@@ -2323,7 +2330,7 @@ impl WriteTransaction {
         let page_allocator = self.page_allocator();
         let mut free_page = |page| {
             // These pages cannot be unpersisted: free_until is bounded by
-            // oldest_live_read_transaction, which pins back to the durable_ancestor of every
+            // oldest_local_read_transaction, which pins back to the durable_ancestor of every
             // pending non-durable commit (see register_non_durable_commit). As a result, no entry
             // whose pages are still unpersisted is eligible for processing here.
             debug_assert!(!self.mem.unpersisted(page));
--- a/src/tree_store/page_store/active_transactions.rs
+++ b/src/tree_store/page_store/active_transactions.rs
@@ -1,0 +1,133 @@
+//! The active transaction range of the multi-process locking protocol (`docs/design.md`): a
+//! handle reading transaction `t` holds the byte `TXN_BASE + t` shared for as long as it reads
+//! it, and a writer about to reclaim pages asks which of those bytes other processes hold.
+
+use crate::db::TXN_BASE;
+use core::ops::Range;
+
+/// The lowest id in `low..=high` whose byte `query` reports held, if any. The bytes cannot be
+/// listed, only probed over a range, so the id is found by bisection, after a probe of `low`
+/// alone: a scan starts from what the one before it found, which is the likeliest answer.
+pub(super) fn lowest_held<E>(
+    query: impl Fn(Range<u64>) -> Result<bool, E>,
+    mut low: u64,
+    mut high: u64,
+) -> Result<Option<u64>, E> {
+    if low > high {
+        return Ok(None);
+    }
+    if query(bytes(low, low))? {
+        return Ok(Some(low));
+    }
+    if low == high || !query(bytes(low + 1, high))? {
+        return Ok(None);
+    }
+    low += 1;
+    while low < high {
+        let mid = low + (high - low) / 2;
+        if query(bytes(low, mid))? {
+            high = mid;
+        } else {
+            low = mid + 1;
+        }
+    }
+    Ok(Some(low))
+}
+
+/// The bytes of the ids `low..=high`
+fn bytes(low: u64, high: u64) -> Range<u64> {
+    let end = TXN_BASE + high + 1;
+    TXN_BASE + low..end
+}
+
+#[cfg(all(test, any(target_os = "linux", target_vendor = "apple", windows)))]
+mod test {
+    use super::lowest_held;
+    use crate::db::{TXN_BASE, byte_range};
+    use crate::tree_store::file_backend::range_lock::RangeLock;
+    use alloc::vec::Vec;
+    use std::fs::{File, OpenOptions};
+    use std::path::Path;
+
+    fn reopen(path: &Path) -> File {
+        OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open(path)
+            .unwrap()
+    }
+
+    // Stands in for a read transaction's lock: the scan only cares that the byte is held
+    fn pin(file: &File, id: u64) {
+        assert!(
+            file.try_lock_shared_range(byte_range(TXN_BASE + id))
+                .unwrap()
+        );
+    }
+
+    fn unpin(file: &File, id: u64) {
+        file.unlock_range(byte_range(TXN_BASE + id)).unwrap();
+    }
+
+    #[test]
+    fn the_scan_finds_the_lowest_held_byte_from_its_start() {
+        let tmpfile = crate::create_tempfile();
+        let scanning = reopen(tmpfile.path());
+        let other = reopen(tmpfile.path());
+        let scan = |low| lowest_held(|range| scanning.query_lock(range), low, 1 << 20).unwrap();
+
+        assert_eq!(scan(0), None);
+
+        // The oldest of another handle's pins is what bounds reclamation, wherever they sit
+        pin(&other, 77777);
+        pin(&other, 1234);
+        assert_eq!(scan(0), Some(1234));
+        assert_eq!(scan(1234), Some(1234));
+        // Nothing below the start is looked at
+        assert_eq!(scan(1235), Some(77777));
+        assert_eq!(scan(77778), None);
+
+        // ... and the scan follows them up as the older ones are released
+        unpin(&other, 1234);
+        assert_eq!(scan(0), Some(77777));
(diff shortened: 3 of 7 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -812,6 +821,46 @@ impl TransactionalMemory {
...
+        ceiling: TransactionId,
+        own: &BTreeSet<u64>,
+    ) -> Result<Option<TransactionId>> {
+        // Every id scanned is at most the ceiling, so this bounds the whole scan's offsets
+        Self::active_transaction_byte(ceiling)?;
+        let _guard = self.lock_header_exclusive()?;
+        let mut scan_from = self.scan_from.lock().unwrap();
+        let oldest = active_transactions::oldest_foreign_pin(
+            |range| self.storage.query_lock_range(range),
+            *scan_from,
```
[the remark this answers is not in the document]
cberner: Restricted the remembered start to `SingleWriterProcess` in 5b086ad. The remembered start relies on every other handle reading the id it pins from the file, which a read-only handle does and a writable multi-writer handle does not yet, so `MultiWriterProcess` scans from zero until the writer-side reload (still in #1413) makes those handles current. `only_a_single_writer_scans_from_where_it_stopped` pins the difference: a byte held below the last scan's ceiling is found in multi-writer mode and skipped in single-writer mode.

---

CHANGELOG.md
```diff
@@ -35,7 +35,8 @@
   takes a `ConcurrencyMode` configuring how processes may share the database.
   When `SingleWriterProcess` or `MultiWriterProcess` is configured, commits are always 2-phase,
   `Durability::None` is refused, and the database may be opened read-only while another process has
-  it open for writing; each new read transaction then sees that process's durable commits.
+  it open for writing; each new read transaction then sees that process's durable commits, and
+  the writer keeps the pages a read transaction in another process is using.
```
cberner: Revert. This is obvious

cberner: Reverted in 6ba3f79: the bullet is back to master's text, and #1413's own bullet drops the same clause.

---

src/transactions.rs
```diff
@@ -2002,10 +2002,11 @@ impl WriteTransaction {
...
 
-        let free_until_transaction = self
-            .transaction_tracker
-            .oldest_live_read_transaction()
-            .map_or(self.transaction_id, |x| x.next());
+        let free_until_transaction = self.transaction_tracker.free_until_transaction(
+            #[cfg(feature = "experimental-multiprocess")]
+            &self.mem,
+            self.transaction_id,
+        )?;
```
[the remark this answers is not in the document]
cberner: Confirmed: with this floor, `compact()` beside a reading peer spins in the drain it begins with (the test hangs until killed without a refusal). Compaction is refused in both shared modes in #1442, cut on master as its own change per cberner, since the refusal stands on master too, where compaction proceeds over a peer's read. This PR should merge after #1442 or be rebased onto it; its body says so.

---

src/transaction_tracker.rs
```diff
@@ -479,6 +479,36 @@ impl TransactionTracker {
...
+            .keys()
+            .map(|id| id.raw_id())
+            .collect();
+        drop(state);
+
+        let local = oldest_local.map_or(transaction_id, TransactionId::next);
+        #[cfg(feature = "experimental-multiprocess")]
+        if mem.concurrency_mode().is_multi_process_writable()
+            && let Some(oldest) =
+                mem.oldest_foreign_pin(mem.get_last_committed_transaction_id()?, &own)?
```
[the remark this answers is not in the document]
cberner: That is the writer-side reload, a slice of its own still in #1413 (decision 5 in the body, now with a sentence on this). On master a writable `MultiWriterProcess` handle that fell behind another writer reads a stale snapshot whatever the floor, since it refreshes neither for reads nor for writes yet; a second writer is not usable until that slice lands. This PR closes the gap for the handles that do refresh, the read-only ones (#1432), and makes no writable-handle case worse, so the reload stays where it is rather than widening this change.

---

src/transaction_tracker.rs
```diff
@@ -479,6 +479,36 @@ impl TransactionTracker {
             .copied()
     }
 
+    /// The transaction up to which a durable commit may reclaim, exclusive: the one after the
+    /// oldest still being read, here or in another process, or `transaction_id` when there is
+    /// none. This process's reads are sampled once, for the local floor and for the ids the scan
+    /// steps over, so a read that begins or ends in between cannot hide a peer's pin on its id.
+    pub(crate) fn free_until_transaction(
+        &self,
```
cberner: This is the tracker so free until sounds like the wrong name. It should be oldest active or something like that. The old code said oldest live read

src/transactions.rs
```diff
@@ -2087,7 +2088,14 @@ impl WriteTransaction {
 
         self.apply_savepoint_state_on_commit();
 
-        if self.post_commit_free == PostCommitFree::Enabled {
+        // The post-commit pass runs after publication, where a peer's pin on the transaction just
+        // superseded lands below any floor the scan above produced, so a shared file's pages wait
+        // for the next durable commit
+        #[cfg(feature = "experimental-multiprocess")]
+        let shared = self.mem.concurrency_mode().is_multi_process_writable();
```
cberner: Shared js the wrong name. See my previous feedback

src/transaction_tracker.rs
cberner: Renamed in 73c12f3: the tracker's method is `oldest_active_transaction()`, returning the oldest transaction still being read here or in another process as an `Option<TransactionId>`, and `durable_commit()` derives `free_until_transaction` from it with the same `map_or(self.transaction_id, |x| x.next())` master had.

---

src/transactions.rs
cberner: Renamed to `multiprocess_writer` in 73c12f3, as in `ReadOnlyDatabase::new()`; the comment above it says "a multi-process writer's pages" now too.

---

src/tree_store/page_store/active_transactions.rs
```diff
@@ -0,0 +1,168 @@
...
+    while low < high {
+        let mid = low + (high - low) / 2;
+        if query(bytes(low, mid))? {
+            high = mid;
+        } else {
+            low = mid + 1;
+        }
+    }
+    Ok(low)
+}
```
cberner: Does std provide a bisect or something for us to use?

src/transaction_tracker.rs
```diff
@@ -479,6 +479,33 @@ impl TransactionTracker {
...
+            .live_read_transactions
+            .keys()
+            .map(|id| id.raw_id())
+            .collect();
+        drop(state);
+
+        #[cfg(feature = "experimental-multiprocess")]
+        if mem.concurrency_mode().is_multi_process_writable()
+            && let Some(foreign) =
+                mem.oldest_foreign_pin(mem.get_last_committed_transaction_id()?, &own)?
```
cberner: Hmm, doesn't the design doc say the writer had to hold the header lock exclusive for this entire scan? The get last committed is done outside that lock. Is that really safe?

src/tree_store/page_store/active_transactions.rs
cberner: No. `partition_point()` and `binary_search_by()` live on slices, so they need something to index, and the ids run to `2^63`; and the probe is a syscall that can fail, which their infallible predicates cannot propagate. There is no bisection over a range and a fallible predicate in core or std.

---

src/transaction_tracker.rs
cberner: The scan itself, every probe, was under the exclusive hold; only the ceiling was read before it. That was safe because the ceiling is this process's own last committed id, which only its own commit advances, and that commit is the one running: in multi-writer mode it holds the writer byte for the transaction, and in single-writer mode there is no other writer. A peer can pin nothing above it, since the id it reads from the file is this writer's last published one. Moved the read under the hold anyway in b768470, so `oldest_foreign_pin()` decides nothing about the scan outside it; the tracker no longer passes a ceiling.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -812,6 +821,48 @@ impl TransactionalMemory {
...
+    ) -> Result<Option<TransactionId>> {
+        // Every id scanned is at most the ceiling, so this bounds the whole scan's offsets
+        Self::active_transaction_byte(ceiling)?;
+        let _guard = self.lock_header_exclusive()?;
+        let mut scan_from = self.scan_from.lock().unwrap();
+        // Only while every peer reads the id it pins from the file: see `scan_from`
+        let remembers = matches!(self.concurrency_mode, ConcurrencyMode::SingleWriterProcess);
+        let oldest = active_transactions::oldest_foreign_pin(
+            |range| self.storage.query_lock_range(range),
+            if remembers { *scan_from } else { 0 },
```
cberner: Why is this gated on single process writer? It seems fine to do it on multi writer also. Just start at the oldest each time , since it can't go backward

src/transaction_tracker.rs
```diff
@@ -479,6 +479,33 @@ impl TransactionTracker {
...
+        {
+            return Ok(Some(local.map_or(foreign, |local| local.min(foreign))));
+        }
+        Ok(local)
```
cberner: This seems off. Is it simpler to just get our oldest local id , instead of building the own set, pass that in and make the active function return the oldest global, instead of differentiating foreign?

src/tree_store/page_store/page_manager.rs
cberner: Both modes remember now, in ea4fe77: `remembers` and the mode check are gone, and the next scan starts at the last result in either mode. The argument that a pin cannot go backward holds once every handle reads the id it pins from the file; a multi-writer handle that has not refreshed is unusable whatever the floor until the reload lands, so the gate bought nothing.

---

src/transaction_tracker.rs
cberner: Done in ea4fe77: `durable_commit()` passes the tracker's `oldest_live_read_transaction()` to `TransactionalMemory::oldest_active_transaction(local)`, which returns the oldest in any process. It probes only below the local oldest, since nothing at or above it can lower the floor, so the own set, the step-over and the Windows own-lock question are gone, and so is the tracker wrapper. The bisection is `active_transactions::lowest_held(query, low, high)`.

---

src/transactions.rs
```diff
@@ -2003,8 +2003,8 @@ impl WriteTransaction {
         }
 
         let free_until_transaction = self
-            .transaction_tracker
-            .oldest_live_read_transaction()
```
cberner, line 2007: I think we should rename this function to add "local" since it's not the oldest global. Or maybe just remove it? Who else calls it and do need updating?

cberner (commented):
One last question. The rest looks good I think

src/transactions.rs
cberner: Renamed to `oldest_local_read_transaction()` in e58026c. Its only other caller is `process_data_freed_pages_after_commit()`, the post-commit pass, which is skipped for a multi-process writer, so the local oldest is the right bound there and nothing needs the global one; the two comments naming it are updated too. Its sibling `oldest_live_read_nondurable_transaction()` is untouched: non-durable commits are refused in the shared modes, so it only ever runs single-process. Say the word if you want it renamed to match.

---

PR #1442, opened 2026-09-03
Refuse compaction while the file is shared

cberner:
One of the refusals from #1413, cut as its own change on master: `Database::compact()` is refused in `SingleWriterProcess` and `MultiWriterProcess`.

## Why

`compact()` checks for live read transactions and savepoints in this process's tracker alone, so a read transaction in another process is invisible to it: its snapshot's pages would be relocated and freed under it. And once the writer bounds reclamation by the pins those transactions hold (#1441), the drain `compact()` begins with, which commits until no free is pending, would wait on a peer's pin for as long as it holds; Codex found that on #1441. Either way compaction cannot proceed while the file is shared, so the shared modes refuse it before any of its checks.

#1441 should merge after this one, or be rebased onto it: without the refusal, `compact()` beside a reading peer hangs there.

1. **Refuse, rather than compact around a peer.** A drain that stops when a commit reclaims nothing would let `compact_pages()` run with frees pending, and `compact()` would return with the file not compacted; refusing keeps its contract.
2. **The error is `CompactionError::Storage(StorageError::Io(_))` of kind `Unsupported`**, as #1413 reports it and as a shared mode on a caller-supplied backend is reported today. The alternative is a `CompactionError` variant of its own; the enum is `#[non_exhaustive]`, so one can still be added. The method's doc comment names the error and the reason.

## Testing

`compaction_is_refused_while_the_file_is_shared`: in each shared mode, a read-only handle on a second file description holds a read transaction while the writer overwrites its data, then `compact()` returns the unsupported error. Without the refusal the test fails at `unwrap_err()` on master, where compaction proceeds over the peer; on #1441 it hangs in the drain. The changelog bullet names the refusal. The full local check set passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,8 +34,9 @@
 * Add the `experimental-multiprocess` feature flag, under which `Builder::set_concurrency_mode()`
   takes a `ConcurrencyMode` configuring how processes may share the database.
   When `SingleWriterProcess` or `MultiWriterProcess` is configured, commits are always 2-phase,
-  `Durability::None` is refused, and the database may be opened read-only while another process has
-  it open for writing; each new read transaction then sees that process's durable commits.
+  `Durability::None` and `Database::compact()` are refused, and the database may be opened read-only
+  while another process has it open for writing; each new read transaction then sees that process's
+  durable commits.
 
 ### redb-derive (unreleased)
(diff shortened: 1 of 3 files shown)


cberner:
Closing in favour of #1444: rather than refusing compaction in the shared modes, `compact()` holds the exclusive header lock for its whole duration and refuses a peer's read transaction as a transaction in progress, the same way it refuses a local one.

---

PR #1445, opened 2026-09-04
Give the writer lock an owner every transaction shares

cberner:
Prep for #1444, which needs a compaction to hold the writer lock across the several write transactions it runs. Nothing here is user-visible.

## What it does

Which lock admits a process to write depends on the concurrency mode: the whole file where the database is not shared, `WRITER_BYTE` where one process writes, and `WRITER_BYTE` taken per transaction where several do. `TransactionGuard::new_write()` re-derived that on every begin, through `lock_multi_process_writer()` returning `None` in the first two modes, so a transaction in them held nothing and ran under whatever the open had left locked.

`TransactionalMemory` now holds the lock the open took, and `share_writer_lock()` is the one way to get a share of it: it clones that lock, or takes the writer byte where a transaction is what takes it. A write transaction holds a share for its whole life. Two things follow from the refcount that were previously true only by construction — a transaction that outlives its `Database` keeps the lock, and no range is ever locked twice on one file description — and one thing that was not true before: `close()` writes the shutdown header while the lock is still held, because the memory that owns it is the thing being closed.

`WriterLock` holds the storage rather than the `TransactionalMemory` around it, because `check_integrity()` takes the latter by `Arc::get_mut()`, which refuses while any other `Arc` or `Weak` to it exists. A read-only handle gets no writer lock, which it never had a use for.

1. **The memory owns the open's lock.** It was on the `Database` first, which meant threading it into `close_database()`, `ensure_allocator_state_table_and_trim()` and the deferred close, and getting the shutdown ordering right by hand. On the memory, none of that is threaded and the ordering is structural. The alternative to owning the real range at all was a guard that guards nothing in the two modes that settle the writer at open; holding the range makes the lifetime the truth rather than a convention.
2. **The write slot is taken before the share.** This description's locks on the writer byte are one lock, so a thread that took the byte while another held the slot would have it released from under it when that one ended. `Database::write_guard()` is where both are taken, in that order, and `begin_write_with_allocation_policy()` receives the guard already built.
3. **The release is handed to `end_write_transaction()`.** It drops the share before clearing the slot, under the lock the slot uses, so a waiting thread cannot take the byte while the outgoing transaction still holds it. This is why `TransactionGuard::Write` holds an `Option`: a field cannot be moved out of a type that implements `Drop`, so `take()` is the only way to hand it over. It is `Some` for the transaction's whole life.

## Testing

`a_single_writer_open_holds_the_byte_for_its_lifetime` covers the mode where this changes most, and `a_transaction_outliving_the_database_keeps_the_writer_lock` is new: the byte is still held after the `Database` is dropped under a live transaction, and free once the deferred close has run. Both fail if `share_writer_lock()` takes a lock instead of cloning the open's. The ordering in decision 3 and the shutdown window in decision 1 are not observable from a test — the shared modes reject a caller-supplied backend, and a single-process one has no locks to watch — so they rest on the structure rather than on coverage. The full local check set passes.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1,10 +1,10 @@
 use crate::io;
 use crate::transaction_tracker::{TransactionId, TransactionTracker};
 use crate::tree_store::LocklessBackend;
-#[cfg(feature = "experimental-multiprocess")]
-use crate::tree_store::MultiProcessWriterGuard;
 #[cfg(not(redb_no_std))]
 use crate::tree_store::ReadOnlyBackend;
+#[cfg(feature = "experimental-multiprocess")]
+use crate::tree_store::WriterLock;
 use crate::tree_store::{
     AllocationPolicy, BtreeHeader, InternalTableDefinition, PAGE_SIZE, PageHint, PageNumber,
     PageResolver, ShrinkPolicy, TableTree, TableType, TransactionalMemory,
@@ -367,12 +367,10 @@ pub(crate) enum TransactionGuard {
     Write {
         tracker: Arc<TransactionTracker>,
         transaction_id: TransactionId,
-        // Owned alongside the write slot so the two can be ordered against each other: a
-        // thread waiting on that slot takes the byte on this same file description the moment
-        // it is free, so the byte must be released first and taken after.
-        // None indicates that the writer lock is held at the database level -- i.e. in a single writer mode
+        // The writer lock, held as long as the transaction is. `Option` only so that `Drop`
+        // can move it out
         #[cfg(feature = "experimental-multiprocess")]
-        multi_process_writer: Option<MultiProcessWriterGuard>,
+        writer_lock: Option<Arc<WriterLock>>,
     },
     // Used for internal accesses that happen outside of any tracked transaction,
     // such as opening the database, repairing it, and running integrity checks.
@@ -434,30 +432,18 @@ impl TransactionGuard {
         Ok(Self::new_read(id, tracker, mem))
     }
 
+    /// The caller must already hold the write slot and the lock the transaction writes under.
     pub(crate) fn new_write(
         transaction_id: TransactionId,
         tracker: Arc<TransactionTracker>,
-        #[cfg(feature = "experimental-multiprocess")] mem: &Arc<TransactionalMemory>,
-    ) -> Result<Self> {
-        #[allow(unused_mut)]
-        let mut guard = Self::Write {
+        #[cfg(feature = "experimental-multiprocess")] writer_lock: Arc<WriterLock>,
+    ) -> Self {
+        Self::Write {
             tracker,
             transaction_id,
             #[cfg(feature = "experimental-multiprocess")]
-            multi_process_writer: None,
-        };
-        // After the guard owns the slot, so that a failure here releases it rather than
-        // leaving the tracker with a live write transaction nothing will ever end
-        #[cfg(feature = "experimental-multiprocess")]
-        if let Self::Write {
-            multi_process_writer,
-            ..
-        } = &mut guard
-        {
-            *multi_process_writer = TransactionalMemory::lock_multi_process_writer(mem)?;
+            writer_lock: Some(writer_lock),
         }
-
-        Ok(guard)
     }
 
     pub(crate) fn untracked() -> Self {
@@ -492,15 +478,16 @@ impl Drop for TransactionGuard {
                 tracker,
                 transaction_id,
                 #[cfg(feature = "experimental-multiprocess")]
-                    multi_process_writer: writer,
+                writer_lock,
             } => {
-                // Drop the "writer byte" multi-process file lock, before our in-process lock.
-                // Otherwise, another transaction in our process could re-enter the file lock
-                #[cfg(feature = "experimental-multiprocess")]
-                drop(writer.take());
-                if let Some(mem) = tracker.end_write_transaction(*transaction_id) {
-                    // The Database was dropped while this transaction was live, deferring
-                    // the database close to the end of this transaction
+                let deferred = tracker.end_write_transaction(
+                    *transaction_id,
+                    #[cfg(feature = "experimental-multiprocess")]
+                    writer_lock.take(),
+                );
+                // The Database was dropped while this transaction was live, deferring
+                // the database close to the end of this transaction
+                if let Some(mem) = deferred {
                     close_database(tracker, &mem);
                 }
             }
@@ -1459,28 +1446,48 @@ impl Database {
     /// transaction remains usable and the database closes when the transaction completes.
     pub fn begin_write(&self) -> Result<WriteTransaction, TransactionError> {
         begin_write_with_allocation_policy(
+            self.write_guard()?,
             &self.transaction_tracker,
             &self.mem,
             AllocationPolicy::Default,
         )
+    }
+
+    /// The write slot, and the lock the transaction will write under.
+    fn write_guard(&self) -> Result<TransactionGuard, TransactionError> {
+        // Fail early if there has been an I/O error -- nothing can be committed in that case
+        self.mem.check_io_errors()?;
+        let transaction_id = self.transaction_tracker.start_write_transaction();
+        #[cfg(feature = "experimental-multiprocess")]
+        let writer_lock = match self.mem.lock_writer() {
+            Ok(lock) => lock,
+            Err(err) => {
+                // Nothing owns the slot yet, and this database is live, so no close is deferred
+                let deferred = self
+                    .transaction_tracker
+                    .end_write_transaction(transaction_id, None);
+                assert!(deferred.is_none());
+                return Err(err.into());
+            }
+        };
+
+        Ok(TransactionGuard::new_write(
+            transaction_id,
+            self.transaction_tracker.clone(),
+            #[cfg(feature = "experimental-multiprocess")]
+            writer_lock,
+        ))
     }
 }
 
 // The allocation policy is fixed for the lifetime of the transaction; every page allocation
 // this transaction makes goes through it.
 fn begin_write_with_allocation_policy(
+    guard: TransactionGuard,
     transaction_tracker: &Arc<TransactionTracker>,
     mem: &Arc<TransactionalMemory>,
     allocation_policy: AllocationPolicy,
 ) -> Result<WriteTransaction, TransactionError> {
-    // Fail early if there has been an I/O error -- nothing can be committed in that case
-    mem.check_io_errors()?;
-    let guard = TransactionGuard::new_write(
-        transaction_tracker.start_write_transaction(),
-        transaction_tracker.clone(),
-        #[cfg(feature = "experimental-multiprocess")]
-        mem,
-    )?;
     // Re-checked after acquiring the write slot: the writer this call blocked on can fail its
     // commit, latching an I/O error and discarding the allocator state. The I/O check comes
     // first so a backend failure is not misreported as corruption. Returning drops the guard,
@@ -1512,8 +1519,24 @@ fn ensure_allocator_state_table_and_trim(
     // commit's writes at high page indices (see AllocationPolicy::Lowest)
     // and try_shrink can't reclaim the growth. See
     // https://github.com/cberner/redb/issues/1165
-    let mut tx =
-        begin_write_with_allocation_policy(transaction_tracker, mem, AllocationPolicy::Lowest)?;
+    // Before the lock, which waits on a peer holding the writer byte: a latched I/O failure
+    // means there is nothing to write here anyway
+    mem.check_io_errors()?;
+    // The database is being closed, so nothing can be waiting on the write slot
+    #[cfg(feature = "experimental-multiprocess")]
+    let writer_lock = mem.lock_writer()?;
+    let guard = TransactionGuard::new_write(
+        transaction_tracker.start_write_transaction(),
+        transaction_tracker.clone(),
+        #[cfg(feature = "experimental-multiprocess")]
+        writer_lock,
+    );
+    let mut tx = begin_write_with_allocation_policy(
+        guard,
+        transaction_tracker,
+        mem,
+        AllocationPolicy::Lowest,
+    )?;
     tx.set_quick_repair(true);
     tx.disable_post_commit_free();
     tx.set_shrink_policy(ShrinkPolicy::Maximum);
@@ -2306,7 +2329,7 @@ mod writer_byte_test {
 
     /// This mode settles who writes at open instead, so the byte is held from then until the
     /// database closes. A transaction taking it again would convert that hold and drop what
-    /// remained on release, so it takes nothing.
+    /// remained on release, so it holds the open's instead.
     #[test]
     fn a_single_writer_open_holds_the_byte_for_its_lifetime() {
         let tmpfile = crate::create_tempfile();
@@ -2327,6 +2350,29 @@ mod writer_byte_test {
         assert!(
             probe.try_lock_range(byte_range(WRITER_BYTE)).unwrap(),
             "the byte was still held after the database closed"
+        );
+        probe.unlock_range(byte_range(WRITER_BYTE)).unwrap();
+    }
+
+    /// A transaction that outlives the database keeps the lock, and the deferred close writes
+    /// the allocator state under it
+    #[test]
+    fn a_transaction_outliving_the_database_keeps_the_writer_lock() {
+        let tmpfile = crate::create_tempfile();
+        let db = create(tmpfile.path(), ConcurrencyMode::SingleWriterProcess);
+        let probe = probe(tmpfile.path());
+
+        let write = db.begin_write().unwrap();
+        drop(db);
+        assert!(
+            !probe.try_lock_range(byte_range(WRITER_BYTE)).unwrap(),
+            "dropping the database released the byte from under a live transaction"
+        );
+
+        write.commit().unwrap();
+        assert!(
+            probe.try_lock_range(byte_range(WRITER_BYTE)).unwrap(),
+            "the byte was still held after the deferred close"
         );
         probe.unlock_range(byte_range(WRITER_BYTE)).unwrap();
     }
--- a/src/transaction_tracker.rs
+++ b/src/transaction_tracker.rs
@@ -2,6 +2,8 @@ use crate::sync::{Condvar, Mutex};
 #[cfg(feature = "experimental-multiprocess")]
 use crate::tree_store::HeaderGuard;
 use crate::tree_store::TransactionalMemory;
+#[cfg(feature = "experimental-multiprocess")]
+use crate::tree_store::WriterLock;
 use crate::{Key, Result, Savepoint, TypeName, Value};
 use alloc::collections::BTreeSet;
 use alloc::collections::btree_map::BTreeMap;
@@ -184,11 +186,17 @@ impl TransactionTracker {
     pub(crate) fn end_write_transaction(
         &self,
         id: TransactionId,
+        #[cfg(feature = "experimental-multiprocess")] writer_lock: Option<Arc<WriterLock>>,
     ) -> Option<Arc<TransactionalMemory>> {
         let mut state = self.state.lock().unwrap();
         assert_eq!(state.live_write_transaction.unwrap(), id);
+        // Released before the slot is cleared, under the lock the slot uses: a thread waiting on
+        // it would take the same byte on this description, and this release would free theirs
+        #[cfg(feature = "experimental-multiprocess")]
+        drop(writer_lock);
         state.live_write_transaction = None;
         self.live_write_transaction_available.notify_one();
+
         state.deferred_close.take()
     }
 
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -533,23 +533,51 @@ impl Drop for HeaderGuard<'_> {
     }
 }
 
-/// A hold on the writer byte, released when it drops. Owns its handle, unlike `HeaderGuard`,
-/// since a write transaction outlives the call that takes it.
+/// The lock admitting this process to write, released when the last holder drops it. Holds the
+/// storage rather than the `TransactionalMemory`, which an integrity check takes by
+/// `Arc::get_mut()` and would find shared
 #[cfg(feature = "experimental-multiprocess")]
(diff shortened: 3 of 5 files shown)


src/db.rs
```diff
@@ -367,12 +367,14 @@ pub(crate) enum TransactionGuard {
...
+        // A share of the lock admitting this process to write, which the database holds where
+        // the open took it and the transaction takes itself in multi-writer mode. `Some` for
+        // the transaction's whole life: it is taken as the guard is built and released as it
+        // drops. Owned alongside the write slot so the two can be ordered against each other: a
         // thread waiting on that slot takes the byte on this same file description the moment
         // it is free, so the byte must be released first and taken after.
-        // None indicates that the writer lock is held at the database level -- i.e. in a single writer mode
         #[cfg(feature = "experimental-multiprocess")]
-        multi_process_writer: Option<MultiProcessWriterGuard>,
+        writer_lock: Option<Arc<WriterLock>>,
```
cberner, line 373: Hmm, why is this still an Option?

src/db.rs
```diff
@@ -434,27 +436,33 @@ impl TransactionGuard {
...
     }
 
+    /// `writer_lock` is the lock admitting this process to write, where the caller holds it
+    /// already: the database's where the open took it. Otherwise this takes the writer byte,
+    /// which is what admits a multi-writer process.
     pub(crate) fn new_write(
         transaction_id: TransactionId,
         tracker: Arc<TransactionTracker>,
         #[cfg(feature = "experimental-multiprocess")] mem: &Arc<TransactionalMemory>,
+        #[cfg(feature = "experimental-multiprocess")] writer_lock: Option<&Arc<WriterLock>>,
```
cberner: It seems odd to give the TransactionGuard a None. Maybe we can push this up to the caller, at some point its the Database which knows exactly whether it has a WriterLock to clone or needs to acquire the lock

cberner (commented):
This looks like the right approach, but a) do a pass on the comments and naming to apply my previous feedback and style, b) check every place that still has an Option in this diff to see if it's really appropriate. Ideally it's just at the Database level, and all the lower layers can just have an Arc<WriterLock>

src/db.rs
cberner: Pushed in 86f2421. `new_write()` now takes an `Arc<WriterLock>` and stores it, and the decision moved to `Database::write_guard()`, which takes the write slot and then `share_writer_lock()` — clone the open's, or take the byte. It no longer needs `mem` either.

The slot has to come first: this description's locks on the writer byte are one lock, so a thread that took it while another held the slot would have it released from under it when that one ended. That is why the resolution happens inside `write_guard()` rather than in `begin_write()` before it. `begin_write_with_allocation_policy()` now takes the guard already built, so nothing below the `Database` sees an `Option`.

---

src/db.rs
cberner: This one I could not remove: `Drop` has to hand the lock to `end_write_transaction()`, and Rust forbids moving a field out of a type that implements `Drop` — `mem::replace()` on the whole guard does not help, since destructuring the owned value is the same move (E0509). `Option::take()` is the way to get it out.

Handing it over is what makes the ordering right, so it is worth the `Option`: the tracker now drops the lock under the same state lock that clears the write slot, so a thread waiting on that slot cannot take the byte while this transaction still holds it. A deferred close keeps the lock instead of dropping it, which is what it writes the allocator state under — so the `deferred_writer_lock` field and `take_deferred_writer_lock()` from the first version are gone, and `close_database()` takes an `Arc<WriterLock>`.

The comment on the field now says this. Also in 86f2421: `WriterLock::range` is no longer an `Option` (unlocking a range the backend never took errors, which the drop already ignores, the same way the open ignores it), and a naming and comment pass over the diff.

---

cberner:
86f2421 does both passes. Every `Option` in the diff, and why it is or is not still there:

| Where | Verdict |
|---|---|
| `Database::writer_lock` | Kept — the one level that knows whether the open took a lock. |
| `TransactionalMemory::writer_lock_from_open()` return | Kept — it is what fills that field; `None` is multi-writer mode. |
| `TransactionGuard::Write::writer_lock` | Kept, unavoidably: `Drop` hands the lock to `end_write_transaction()`, and a field cannot be moved out of a type that implements `Drop`. See the thread. |
| `end_write_transaction()`'s parameter | Follows from that `take()`. `None` only on the one path where the slot was taken and the lock could not be. |
| `TransactionGuard::new_write()`'s parameter | Gone — takes `Arc<WriterLock>`, and no longer takes `mem`. |
| `begin_write_with_allocation_policy()`'s parameter | Gone — takes the guard already built, so it does not see the lock at all. |
| `close_database()` / `ensure_allocator_state_table_and_trim()` | Gone — both take `Arc<WriterLock>`. |
| `TransactionTracker`'s `deferred_writer_lock` + `take_deferred_writer_lock()` | Gone — the deferred close carries the transaction's own lock in `DeferredClose`. |
| `WriterLock::range` | Gone — unlocking a range the backend never took errors, which the drop already ignores, as the open does. |

`Database::write_guard()` is where the decision now lives: take the write slot, then clone the open's lock or take the writer byte. The slot has to come first, so the resolution could not move any further up than this.

`a_transaction_outliving_the_database_keeps_the_writer_lock` and `a_single_writer_open_holds_the_byte_for_its_lifetime` both fail if `share_writer_lock()` takes a lock instead of cloning the open's. Full local check set passes.

---

src/db.rs
cberner: Well, can't you clone it instead then?

cberner: No — a clone leaves the field's own `Arc` alive, so the count never reaches zero inside the drop and the unlock lands after it. Field drop glue runs after `Drop::drop` returns, which is after the slot is cleared and the waiter notified, so the waiter takes the byte and this transaction's release then frees it out from under it. That is the window the ordering exists to close. I checked it rather than trusting my reading:

```
body: begin
body: released its clone -- did the file unlock happen?
body: end (slot would be freed here)
drop writer lock          <- after the body, i.e. after the slot is free
after the guard is gone
```

Field order does not work either, for the second half of the problem: a deferred close needs the lock *after* the slot is released, since it begins a transaction of its own. Those two requirements are opposites, and they are mutually exclusive — a deferred close means the database is gone, so nothing can be waiting on the slot. `end_write_transaction()` is the one place that knows which case it is, which is why the lock goes there and it decides under the state lock.

So the `Option` is the move-out mechanism, not a state the code can be in: it is `Some` from construction until that hand-off. If you would still rather not see it, the alternative I can see is for the tracker to own the writer byte for the life of the write slot and hand transactions nothing — the ordering then falls out of one lock — but that puts file locking in `TransactionTracker` and makes #1444's compaction lending harder, so I did not take it on my own.

---

src/db.rs
```diff
@@ -1528,12 +1574,22 @@ fn ensure_allocator_state_table_and_trim(
...
     // No saved allocator state when it needs repair: the next open must rebuild it instead
     // of trusting the saved one
     if !crate::panicking()
         && !mem.needs_repair()
-        && ensure_allocator_state_table_and_trim(transaction_tracker, mem).is_err()
+        && ensure_allocator_state_table_and_trim(
+            transaction_tracker,
+            mem,
+            #[cfg(feature = "experimental-multiprocess")]
+            writer_lock,
```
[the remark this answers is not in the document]
cberner: Confirmed and fixed in 7598be4. `close_database()` moved the lock into the allocator-state transaction, whose guard released it when that transaction ended — before `mem.close()` writes the shutdown header. In the deferred path that was the last owner, so those final writes ran unlocked, which is a regression against master, where the range lives with the file description until it closes.

It now borrows: `close_database()` and `ensure_allocator_state_table_and_trim()` take `Option<&Arc<WriterLock>>`, so the owner stays with the caller — the `Database`'s own field on the direct path, and the `DeferredClose` binding on the deferred one, both of which outlive the call. The `Option` is back on those two signatures, and this is why, along with the sibling finding about waiting on a peer's byte.

Not covered by a test: the window is inside `close_database()`, between the allocator transaction ending and `close()` returning, and there is no seam to observe it from — the shared modes reject a caller-supplied backend, and a single-process one has no locks to watch.

---

src/db.rs
```diff
@@ -1560,7 +1616,21 @@ impl Drop for Database {
             return;
         }
 
-        close_database(&self.transaction_tracker, &self.mem);
+        #[cfg(feature = "experimental-multiprocess")]
+        let Ok(writer_lock) = self.share_writer_lock() else {
```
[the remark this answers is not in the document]
cberner: Confirmed and fixed in 7598be4. `Database::drop()` resolved the share before calling `close_database()`, so in multi-writer mode it waited on a peer's writer byte before reaching the `panicking()` and `needs_repair()` checks that decide nothing will be written. On master the acquisition sits inside `ensure_allocator_state_table_and_trim()`, past those checks, so this was a regression.

It is back there: `ensure_allocator_state_table_and_trim()` takes `Option<&Arc<WriterLock>>` and calls `share_writer_lock()` itself, which is only reached when a write is actually about to happen. `Database::drop()` now just passes `self.writer_lock.as_ref()` and the fallback branch it needed is gone.

---

src/transaction_tracker.rs
```diff
@@ -184,12 +195,27 @@ impl TransactionTracker {
...
+        #[cfg(feature = "experimental-multiprocess")]
+        let deferred = deferred.map(|mem| DeferredClose {
+            mem,
+            writer_lock: writer_lock
+                .expect("a write transaction with a deferred close held the writer lock"),
+        });
+        #[cfg(not(feature = "experimental-multiprocess"))]
+        let deferred = deferred.map(|mem| DeferredClose { mem });
         state.live_write_transaction = None;
         self.live_write_transaction_available.notify_one();
```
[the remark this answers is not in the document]
cberner: This one was already ordered correctly, but only by a detail worth not relying on, so 7598be4 makes it explicit.

The closure passed to `map()` captured `writer_lock` by move, since `Option::expect()` takes `self`. When the deferred close is `None`, `map()` drops that closure — and the lock with it — before returning, so the release happened before the slot was cleared rather than at function scope. Checked rather than reasoned about:

```
before map
  unlock (writer lock dropped)
after map  <- slot would be cleared and the waiter notified here
```

The lock is now `take()`n for a deferred close and `drop()`ped explicitly otherwise, both while the state lock is held and before the slot is cleared, which is what the comment there claims.

---

src/db.rs
```diff
@@ -710,6 +705,11 @@ impl ReadOnlyDatabase {
 pub struct Database {
     mem: Arc<TransactionalMemory>,
     transaction_tracker: Arc<TransactionTracker>,
+    // The lock admitting this process to write, where the open took it: every write
+    // transaction clones this rather than taking one, so no range is locked twice on this
+    // file description. `None` in multi-writer mode, where a transaction takes the byte
+    #[cfg(feature = "experimental-multiprocess")]
+    writer_lock: Option<Arc<WriterLock>>,
```
cberner: Ok, so looking at this PR some more, do you think this is the right place to hold the WriterLock? Maybe it should be on TransactionalMemory instead. That way the deferred close can just get it back from the TM, instead of needing to thread it through more code

cberner (commented):
I think the comments still need some trimming

src/db.rs
cberner: Yes — that is better, and it is in b3c50c0. `TransactionalMemory` holds the lock the open took, so nothing is threaded through the close at all:

- `close_database()` is back to master's signature, and `ensure_allocator_state_table_and_trim()` calls `mem.share_writer_lock()` itself.
- `DeferredClose` is gone; `end_write_transaction()` returns `Option<Arc<TransactionalMemory>>` as it did before.
- `Database::writer_lock` and the free `share_writer_lock()` are gone; `share_writer_lock()` is a method on `TransactionalMemory` and the only way to get a share.

It also fixes the shutdown finding by construction rather than by a borrow: the owner is the memory itself, which outlives `close()` by definition, so there is no window where the last share dies before the shutdown header is written. No `Weak` needed either, since `WriterLock` holds the storage rather than the memory — `check_integrity()`'s `Arc::get_mut()` is untouched.

There is no writer lock for a read-only handle now, which there should not have been: it never writes.

The guard's `Option` is the only one left below the memory, still just the `Drop` move-out. Comments trimmed again in the same commit.

---

src/db.rs
```diff
@@ -367,12 +367,11 @@ pub(crate) enum TransactionGuard {
     Write {
         tracker: Arc<TransactionTracker>,
         transaction_id: TransactionId,
-        // Owned alongside the write slot so the two can be ordered against each other: a
-        // thread waiting on that slot takes the byte on this same file description the moment
-        // it is free, so the byte must be released first and taken after.
-        // None indicates that the writer lock is held at the database level -- i.e. in a single writer mode
+        // A share of the writer lock, handed to end_write_transaction(), which orders its
+        // release against the write slot. `Option` only because a field cannot be moved out of
+        // a type that implements `Drop`
```
cberner: Needs work. Don't describe the functions it gets handed to. See my previous feedback on comments

src/db.rs
```diff
@@ -434,30 +433,19 @@ impl TransactionGuard {
         Ok(Self::new_read(id, tracker, mem))
     }
 
+    /// The caller must already hold the write slot and the writer lock share: see
+    /// `Database::write_guard()`.
```
cberner: Describing it as a "share" is confusing

src/tree_store/page_store/page_manager.rs
```diff
@@ -782,19 +813,20 @@ impl TransactionalMemory {
...
+    /// byte are one lock, and only where a write follows, since taking it waits on the holder.
     #[cfg(feature = "experimental-multiprocess")]
-    pub(crate) fn lock_multi_process_writer(
-        mem: &Arc<Self>,
-    ) -> Result<Option<MultiProcessWriterGuard>> {
-        // The other modes settle who writes when the database is opened: a single-writer holds this byte
-        // for its lifetime, and taking it again here would convert that hold and then drop it.
-        if !matches!(mem.concurrency_mode, ConcurrencyMode::MultiWriterProcess) {
-            return Ok(None);
+    pub(crate) fn share_writer_lock(&self) -> Result<Arc<WriterLock>> {
```
cberner: Share writer lock is a weird name

cberner (commented):
The naming and comments need work. Look at the last few commits to see how Fable wrote them. They were much better

src/tree_store/page_store/page_manager.rs
cberner: Renamed to `lock_writer()` in 89808f1, alongside `lock_header_shared()` and `lock_header_exclusive()` in the same file. "share" is gone from the code entirely — the field is `writer_lock` and the local is too.

Its doc no longer restates what a `WriterLock` is, just what this one gets you and the rule for calling it:

```rust
/// The one the open took, or the writer byte a write transaction takes in multi-writer
/// mode, waiting on the peer that holds it. Taken after the in-process write slot: locks on
/// that byte from this file description are one lock, so the slot is what orders them
```

---

src/db.rs
cberner: Cut to what the field is, in 89808f1:

```rust
        // The writer lock, held as long as the transaction is. `Option` only so that `Drop`
        // can move it out
```

The release ordering it used to describe now appears once, in `end_write_transaction()`, where it is enforced.

---

src/db.rs
cberner: Gone in 89808f1 — the doc is now one line: `The caller must already hold the write slot and the lock the transaction writes under.` `Database::write_guard()`'s doc lost the word too.

Reading the last few merged commits for the pattern was the useful part: state what the thing is, then the invariant that makes it true, and name no call graph. I went back over every comment the diff adds against that, not only the three you marked — the type doc no longer opens the same way as the two fields, and the ordering rule is stated once rather than three times.

---

src/db.rs
```diff
@@ -1512,8 +1519,21 @@ fn ensure_allocator_state_table_and_trim(
     // commit's writes at high page indices (see AllocationPolicy::Lowest)
     // and try_shrink can't reclaim the growth. See
     // https://github.com/cberner/redb/issues/1165
-    let mut tx =
-        begin_write_with_allocation_policy(transaction_tracker, mem, AllocationPolicy::Lowest)?;
+    // The database is being closed, so nothing can be waiting on the write slot
+    #[cfg(feature = "experimental-multiprocess")]
+    let writer_lock = mem.lock_writer()?;
```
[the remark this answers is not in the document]
cberner: Confirmed and fixed in a5a0a34. A failed commit's `AllocatorStateLatch` calls `invalidate_allocator_state()`, not `mark_needs_repair()`, so `close_database()` does enter this helper with the error latched, and the pre-lock check I moved into `write_guard()` no longer covered it. On master `check_io_errors()` ran before the slot and the byte; here it had ended up after both.

```rust
    // Before the lock, which waits on a peer holding the writer byte: a latched I/O failure
    // means there is nothing to write here anyway
    mem.check_io_errors()?;
```

That restores master's ordering: the byte is taken only once the write is known to be possible. `begin_write_with_allocation_policy()` still re-checks afterwards, along with the allocator state, which on master was also behind the byte.

Second time this PR has moved a shutdown-path acquisition in front of a check that decides nothing will be written — worth flagging as the shape to watch here, since neither is observable from a test.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -533,23 +533,50 @@ impl Drop for HeaderGuard<'_> {
...
-        let _ = self.mem.storage.unlock_range(byte_range(WRITER_BYTE));
+        // Errors where the backend has no locks, which the open found the same way
+        let _ = self.storage.unlock_range(self.range.clone());
     }
 }
 
+/// The lock the open took where the mode settles who writes: the whole file where the database
+/// is not shared, and the writer byte where one process writes.
+#[cfg(feature = "experimental-multiprocess")]
+fn writer_lock_from_open(
```
cberner: This needs work too. The rest looks good

cberner: Reworked in e9f1141. It was named like a constructor but took three unrelated things to answer one question, and its doc named only the two cases that produce a lock, not the two that do not.

It is now a classification of the mode, with no storage parameter:

```rust
/// The range the open locked to admit this process as a writer: the whole file where the
/// database is not shared, and the writer byte where one process writes. Nothing in
/// multi-writer mode, where a write transaction takes that byte itself, or for a read-only
/// handle
fn writer_lock_range(read_only: bool, concurrency_mode: ConcurrencyMode) -> Option<Range<u64>>
```

and the guard is built at its one call site, so the early return no longer sits among the arms that yield a range.

---

src/db.rs
```diff
@@ -1512,8 +1519,24 @@ fn ensure_allocator_state_table_and_trim(
...
+    // means there is nothing to write here anyway
+    mem.check_io_errors()?;
+    // The database is being closed, so nothing can be waiting on the write slot
+    #[cfg(feature = "experimental-multiprocess")]
+    let writer_lock = mem.lock_writer()?;
+    let guard = TransactionGuard::new_write(
+        transaction_tracker.start_write_transaction(),
+        transaction_tracker.clone(),
+        #[cfg(feature = "experimental-multiprocess")]
+        writer_lock,
```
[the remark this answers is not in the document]
cberner: The window is real, but it is master's, not this PR's, and holding the lock would not close it. Leaving it for @cberner rather than widening the PR.

On master the same sequence runs: `begin_write_with_allocation_policy()` → `new_write()` → `lock_multi_process_writer()` takes the byte, the transaction holds it, `tx.commit()` ends it, its guard releases the byte, and only then does `close_database()` call `mem.close()`. Nothing on either side holds the byte across the shutdown header write. This PR moved where the byte is released within the guard's drop, not when relative to `close()`.

Holding a clone through `mem.close()` narrows it without fixing it. `flush_shutdown_header()` writes `state.header`, this handle's in-memory copy, and a writable handle in multi-writer mode never picks up a peer's commits — `reload_header()` is gated on `self.read_only`. So a peer that committed at *any* earlier point in this handle's life is overwritten by that write just the same; the byte only excludes a peer committing during the window itself. Same root cause as the ceiling thread on #1444: the missing writer-side reload, which belongs in `begin_write()` after the byte is taken.

If you want the narrowing anyway, it is small — `close_database()` would check `panicking()`/`needs_repair()` first, then take the lock once and hold it across both the allocator write and `mem.close()`, which also keeps the acquisition behind those checks. Say the word and I will add it here; otherwise it goes with the reload slice.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -533,23 +533,42 @@ impl Drop for HeaderGuard<'_> {
...
+        // Errors where the backend has no locks, which the open found the same way
+        let _ = self.storage.unlock_range(self.range.clone());
+    }
+}
+
+/// The range the open locked to admit this process as a writer: the whole file where the
+/// database is not shared, and the writer byte where one process writes. Nothing in multi-writer
+/// mode, where a write transaction takes that byte itself, or for a read-only handle
+#[cfg(feature = "experimental-multiprocess")]
+fn writer_lock_range(read_only: bool, concurrency_mode: ConcurrencyMode) -> Option<Range<u64>> {
```
cberner: Let's switch it back to building the whole lock. And the name should reflect that it's the database lifetime lock. Maybe database_writer_lock_for_mode

cberner: Done in 1b4abc6, with your name:

```rust
/// The writer lock a handle holds for the life of the database: the whole file where the
/// database is not shared, and the writer byte where one process writes. Nothing in
/// multi-writer mode, where a write transaction takes that byte itself, or for a read-only
/// handle
fn database_writer_lock_for_mode(
    storage: &Arc<PagedCachedFile>,
    read_only: bool,
    concurrency_mode: ConcurrencyMode,
) -> Option<Arc<WriterLock>>
```

It builds the whole lock again, so the call site in `TransactionalMemory::new()` is back to one line, and the name now says which lock it is rather than leaving that to the caller.

---

src/db.rs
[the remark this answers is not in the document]
cberner: Done in #1449, the writer-side reload, in both halves: the close's transaction adopts the file's latest commit under the byte, so the shutdown header it writes is the peer's rather than a stale one, and `close_database()` takes the writer lock once and holds it across that commit and `mem.close()`, so no commit lands between them.

---

PR #1446, opened 2026-09-04
Lend a write transaction the locks its caller holds

cberner:
Prep for #1444, which holds the writer byte and the exclusive header lock across the several write transactions a compaction runs. Split out so the plumbing can be checked on its own: every caller here lends nothing, and behaves as before.

## What it does

A write transaction takes the writer lock in `Database::write_guard()` when it begins, and the exclusive header lock when it commits: once for the reclamation scan, and once for each of the two header writes, with a flush between them. A caller that holds either lock across several transactions cannot let a transaction take its own: the header lock's in-process half is a mutex, which its holder cannot take twice, and a second lock on the writer byte from one file description is the same lock, which the transaction's end would release.

`Database::begin_write_with()` takes an `Option<&Arc<WriterLock>>`, and `write_guard()` clones a lent lock rather than calling `lock_writer()`, so the transaction's end drops one holder and leaves the byte held. `WriteTransaction::commit_with()` takes an `Option<&HeaderGuard>`, which `durable_commit()` and `TransactionalMemory::commit()` resolve at each of the three holds through `header_hold()`: the lent guard where there is one, and otherwise a `HeaderHold` of their own that ends with the operation. The leaves take the hold as proof: `write_header()` and `oldest_active_transaction()` take a `&HeaderGuard`, in the style `lock_mp_transaction()` already uses, and the four other header writers (the open's repair, `begin_writable()`, `clear_recovery_required()` and the shutdown header) take one first, ahead of the state lock. `begin_write()` and `commit()` pass `None`, as do the two repair-path commits, so nothing observable changes. `HeaderGuard`'s doc states the rule once, and `lock_header_exclusive()` becomes crate-visible so that a caller can take the hold it lends.

No changelog entry: nothing user-visible changes.

1. **Lend the locks rather than flag the `TransactionalMemory`.** The first cut of #1444 set a `compacting` flag under which every acquisition handed back a guard that locked nothing. Lending threads an `Option` through `commit_with()`, `durable_commit()` and `TransactionalMemory::commit()`, but each hold is a value with an owner and a lifetime, the locking code has no mode, and a transaction that should take a lock of its own cannot silently take none.
2. **Lend the writer lock by `Arc` and the header guard by reference.** The writer lock is already the `Arc<WriterLock>` #1445 gave every transaction, so a lent one is cloned and the byte is released when its last holder drops it. The header guard is the in-process mutex guard itself, which cannot be shared, so it is borrowed for the transaction's lifetime.
3. **Resolve the `Option` at the three holds, not at the entry point.** `commit_with()` could take one hold for the whole commit, but that hold would span the data flush and both header flushes, so every reader, in this process and in others, would wait on the writer's fsyncs to begin. Resolving at each hold keeps master's three short holds for a commit that lends nothing, and only a caller that lends a hold runs a whole commit under one. The leaves then need no `Option` at all.

## Testing

`a_lent_header_hold_outlives_the_commit`: a transaction commits under a lent exclusive hold, which a probe on another description finds still held until it is dropped. A `durable_commit()` that takes its own hold for the scan regardless, or a `TransactionalMemory::commit()` that takes its own for a header write, hangs it on the in-process mutex. `a_lent_writer_byte_outlives_the_transaction`: a transaction lent the writer byte ends without releasing it, which a probe confirms until the lock is dropped; a `write_guard()` that takes the byte instead of cloning the lent lock fails it, since that acquisition and the lent one are one lock, released by the transaction's end. The full local check set passes.

### DIFF
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -36,7 +36,7 @@ use core::convert::TryInto;
 use core::marker::PhantomData;
 use core::mem;
 #[cfg(feature = "experimental-multiprocess")]
-use core::ops::Range;
+use core::ops::{Deref, Range};
 use core::sync::atomic::{AtomicBool, Ordering};
 #[cfg(feature = "logging")]
 use log::warn;
@@ -517,7 +517,10 @@ impl UnpersistedState {
 #[cfg(feature = "experimental-multiprocess")]
 pub(crate) const HEADER_LOCK: Range<u64> = 0..DB_HEADER_SIZE as u64;
 
-/// A hold on the header lock, released when it drops.
+/// A hold on the header lock, released when it drops. Taken ahead of the memory's and the
+/// tracker's state locks, which a reader's registration and the reclamation scan take under it.
+/// Lent to the work under it rather than taken again: the in-process half is a mutex, which its
+/// holder cannot take twice.
 #[cfg(feature = "experimental-multiprocess")]
 pub(crate) struct HeaderGuard<'a> {
     storage: Option<&'a PagedCachedFile>,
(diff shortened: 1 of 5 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -1259,10 +1298,16 @@ impl TransactionalMemory {
     }
 
     pub(crate) fn begin_writable(&self) -> Result {
+        #[cfg(feature = "experimental-multiprocess")]
+        let header_lock = self.lock_header_exclusive()?;
         let mut state = self.state.lock().unwrap();
```
cberner, line 1305: Double check that the ordering of these two locks is always right, with this new refactoring

cberner: Checked every acquisition. The header lock is now taken ahead of both state locks everywhere: a reader's registration and the reclamation scan take the memory's and the tracker's state locks under it, and so do the three writers this changes (`begin_writable()`, `clear_recovery_required()`, the shutdown header). Those three were the only sites that took it the other way round, inside `write_header()` under the state lock, and only in the shared modes. `commit()` drops the state lock before its holds and re-takes it after them, the open path takes the header lock before any state exists, and nothing that holds a state lock calls into an acquisition: `dereference_transaction()` releases a pin without the header lock. Stated once, on `HeaderGuard`, in 3ead8a0.

---

cberner:
lgtm. I'm merge it when CI is green

PR #1448, opened 2026-09-05
Open a multi-writer database under the writer byte

cberner:
Prep for the writer-side reload, on #1443. A `MultiWriterProcess` open that finds no allocator snapshot, which is whenever a live peer's last commit was an ordinary one, rebuilds the allocator and commits holding only the shared writer byte, so a peer's write transaction could commit between the header the open read and the commit it made, and either header then overwrote the other's. Codex raised it on #1443, where the integrity check holds the writer byte and this open was the one writer left that takes none.

## What it does

`TransactionalMemory::new()` takes `WRITER_BYTE` right after the mode bytes, before it reads the header, as the lock the open took: the one `writer_lock` field, which holds the single modes' lock for the life of the database, holds it in multi-writer mode until the handle is writable, when `Database::new()` takes it through `take_open_writer_lock()`. The open holds it until it is over: through the repair and its commit, `begin_writable()`, and the restoration of the file's persistent savepoints, whose transaction is lent the hold as a transaction is lent the open's lock in the other modes; it is released before an error could drop the database, whose close takes the byte itself. A write transaction takes the byte itself once it is released, as before. The single modes are untouched: their lock is held for the life of the database, as it was, and `take_open_writer_lock()` gives `None` for them and for a read-only handle.

The open already waited on a peer's write transaction, at the registration transaction that ends it; it now waits at its start, and the changelog says so, as it does for the check.

1. **The memory takes the byte, the database holds it.** The header read and its write-back are inside `TransactionalMemory::new()`, so the byte has to be taken there; when the open is over is the database's to know, so the hold is a value it takes and drops, rather than state the memory releases on a call, and it is the same field as the other modes' lock rather than a second one, since a handle has one writer lock, whose span depends on the mode. A memory built without a database, as the header lock tests build one, drops the hold at once.
2. **Hold through the whole open, not only the repair.** One hold from before the header read to the end of `Database::new()` is simpler to state than a release and a re-take around the registration transaction, and nothing in between needs the byte free.
3. **Not the reload.** A handle that holds the byte reads a header current as of its open; that it goes stale as peers commit, and that a peer's next commit overwrites this open's repair, is the writer-side reload at `begin_write()` and at the close, which is the next prep.

## Testing

`a_multi_writer_open_holds_the_byte_until_it_is_writable`: with the byte held from another file description, the open does not complete until it is released, and the byte is free once the open returns; never releasing fails the second assertion. `a_multi_writer_open_repairs_under_the_byte`: a live peer's ordinary commit leaves no snapshot, so the open repairs, and the repair callback finds the byte held; not taking it, or dropping the hold before the repair, reports it free. The header lock tests, which open two multi-writer memories directly, release the hold themselves. `restore_persistent_savepoints()` is the transaction the open and the check each ran inline, so that the open can release the hold before it propagates an error. The full local check set passes.

### DIFF
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -572,9 +572,10 @@ impl Drop for WriterLock {
     }
 }
 
-/// The writer lock a handle holds for the life of the database: the whole file where the
-/// database is not shared, and the writer byte where one process writes. Nothing in multi-writer
-/// mode, where a write transaction takes that byte itself, or for a read-only handle
+/// The writer lock a handle holds from the open: for the life of the database, the whole file
+/// where the database is not shared and the writer byte where one process writes; in
+/// multi-writer mode, the writer byte until the handle is writable, which a write transaction
+/// takes itself from then on. Nothing for a read-only handle
 #[cfg(feature = "experimental-multiprocess")]
 fn database_writer_lock_for_mode(
     storage: &Arc<PagedCachedFile>,
@@ -586,8 +587,9 @@ fn database_writer_lock_for_mode(
     }
     let range = match concurrency_mode {
         ConcurrencyMode::SingleProcess => FULL_RANGE,
-        ConcurrencyMode::SingleWriterProcess => byte_range(WRITER_BYTE),
-        ConcurrencyMode::MultiWriterProcess => return None,
+        ConcurrencyMode::SingleWriterProcess | ConcurrencyMode::MultiWriterProcess => {
+            byte_range(WRITER_BYTE)
+        }
     };
 
     Some(Arc::new(WriterLock {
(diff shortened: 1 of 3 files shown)


src/tree_store/page_store/page_manager.rs
```diff
@@ -632,6 +644,10 @@ pub(crate) struct TransactionalMemory {
     // read-only handle
     #[cfg(feature = "experimental-multiprocess")]
     writer_lock: Option<Arc<WriterLock>>,
+    // The writer byte a multi-writer open holds from before the header is read, for the
+    // `Database` to take and hold until the handle is writable: see `take_open_writer_lock()`
+    #[cfg(feature = "experimental-multiprocess")]
+    open_writer_lock: Option<Arc<WriterLock>>,
```
cberner: wtf, we have two WriterLocks? That makes no sense

cberner:
On the two `WriterLock`s: agreed, and gone in 429eddd. A handle has one writer lock, `writer_lock`, whose span depends on the mode. In the single modes it is the open's lock for the life of the database, as before; in multi-writer mode `TransactionalMemory::new()` takes the writer byte into the same field before it reads the header, and `take_open_writer_lock()` hands it to `Database::new()`, which holds it until the open is over. The helper that took the byte is gone too, back to the one place in `lock_writer()`.

The same push takes Codex's finding on the error path: the open releases the hold before it propagates an error from restoring the savepoints, so `db`'s close takes the byte itself rather than a second lock on the same description that its end would release from under the hold. `restore_persistent_savepoints()` is the transaction the open and the check each ran inline, so that the open has one result to release ahead of. The review-thread API is rate-limited for me at the moment, so this is here rather than on the threads; I will resolve Codex's once it answers.

---

cberner:
Closed for now, at cberner's call. The design requires every multi-writer commit to be quick-repair, so the motivating case, an open repairing after a peer's ordinary commit, cannot happen once #1449 enforces that; what remains is the open's header write-back and the repairs after a commit that recorded nothing. To be revisited on those grounds.

---

PR #1449, opened 2026-09-05
Begin a multi-writer transaction from the file's latest commit

cberner:
The writer-side reload itself, on master now that every prep split out of it is merged: #1455, #1456, #1457, #1458, #1460 and #1461. #1443 comes after it. A `MultiWriterProcess` handle never picked up another process's commits: `begin_write()` began from the header the handle had, so two live handles overwrote each other's commits, the compaction scan's ceiling was stale (the Codex thread on #1444), and a close's commit and shutdown header overwrote whatever a peer had committed since (the thread on #1445). The design has every write transaction reload the allocator state from the latest record; #1452 made every commit record one.

## What it does

On the one path #1455 gave every write transaction, `begin_write_with_allocation_policy()` takes the write slot and the writer byte, as #1460 has it, syncs to the file's latest commit in multi-writer mode through `sync_to_latest_commit()`, and then builds the transaction's guard and the transaction. The slot and the byte release themselves on drop, so a start that fails or panics in the sync leaves neither held rather than leaving every later `begin_write()` waiting on the slot; the id is issued past the file's latest commit once the sync is done, through `issue_write_transaction_id()`, and the guard is built from the two with it. `TransactionalMemory::reload_for_write()` reads the header under the header lock, `&HeaderGuard` being the proof, through the read #1457 shares, and where the handle is not on the file's latest commit, by its id, or a panic has latched a repair over the state it holds, syncs to it: the cached pages and the buffered writes are dropped, since that process may have reused any of the pages, and a transaction dropped while a panic unwinds leaves its writes buffered, and the allocator state is dropped for `load_synced_allocator_state()` to load from what the commit recorded. Every commit in multi-writer mode records one, the open's repair's since #1461, so a commit without one is refused as corruption rather than rebuilt from. The file's persistent savepoints are synced through `sync_persistent_savepoints()` from #1456, ahead of anything the transaction frees; it takes the caller's header lock as proof from here, lent by compaction and its own otherwise, as the reload does. A load, or a savepoint sync, that stops part way, on an error or a panic the caller catches, discards the allocator state through the commit path's `AllocatorStateLatch`, which `sync_to_latest_commit()` returns armed and the builder disarms once the savepoints are synced, so the next transaction is refused rather than allocating from a partial state, or freeing from under a savepoint the tracker does not hold yet. A load that succeeds releases the repair latch a transaction dropped in a panic left, since the leak it latched is in the state replaced. That release is why a latched repair syncs where no peer has committed too: a commit made under one records no allocator state, which the next writer to sync to it would refuse, so one panic in one process would otherwise refuse every peer until it reopened.

The close's transaction goes through the same path, so the close commits from the file as a peer last committed it, under the writer byte #1458 has it hold across that commit and the shutdown header. And without the commit, no header: the header the close would write is this handle's, which describes the file only once the commit has synced to it, so a commit that fails, at the sync or after, leaves the file for the next open to recover.

The reload and the compaction hold have to agree on the header lock: compaction lends its exclusive hold to the transactions it begins, as it lends it to their commits, and `header_hold()` gives the reload a hold of its own otherwise. The same lent hold reaches the savepoint sync, which would deadlock under compaction's taking the lock again.

The recovery flag stays this handle's across a sync: a peer's close clears it in the file while this handle is still writing, and the allocator state's load clears it in memory, as it does for the open ahead of `begin_writable()`, so `sync_to_latest_commit()` marks it again for the next commit to write back, for a crash of this handle to be recovered from.

1. **Take the locks, sync, then build the transaction.** The slot and the byte, taken as #1460 has them, release themselves on drop, the sync runs under them, and the guard is built from them with the id once the sync is done, so no id is ever issued unsynced and nothing is left held by a start that fails.
2. **A commit is known by its id.** With every writer reloading under the writer byte, ids only advance, so the file's latest commit is another than the handle's exactly when its id is. The roots need no comparing, and `TransactionHeader` no `PartialEq`. A latched repair syncs whatever the id says, the leak being this handle's own and the recorded state free of it.
3. **No rebuild at the sync.** Every multi-writer commit records the allocator state, the open's repair's with #1461, so a commit without one broke the protocol and is refused rather than repaired from.
4. **The reload runs in multi-writer mode only.** A handle in the other modes is the only writer, so the file's commit is its own; the header read and parse per transaction would be wasted there.
5. **Read transactions on a writable handle still do not reload.** They see the handle's commits and the peer commits its write transactions synced to. Making them reload, as a read-only handle's do, needs the header to be synced ahead of the allocator, and the `begin_read` race a read-only handle guards against; it is a later slice.

## Testing

`a_write_transaction_syncs_to_a_peers_commit`: a peer's row is visible in the transaction, and both handles' commits survive to a reopen; never syncing fails the first assertion, and issuing the id ahead of the sync trips the commit's ordering assertion. `a_write_transaction_syncs_to_a_peers_savepoint`: the peer's persistent savepoint refuses compaction on the handle once a transaction has synced to it, the handle's own savepoint takes a fresh id and both are listed, and once the peer deletes both, compaction proceeds; skipping the sync hands the peer's id out again, and without the forgetting compaction is still refused. `a_close_records_a_peers_commit`: the peer's row survives the handle's close, which used to commit over it. `a_close_writes_no_header_over_a_commit_it_could_not_sync_to`: with the peer's primary slot checksum flipped on disk, the close's commit cannot sync to it, and the slot's bytes read back unchanged after the close; before the fix the close wrote its stale header over them. `a_write_transaction_reads_a_peers_page_over_a_page_a_panic_left_buffered`: a transaction dropped while a panic unwinds leaves its pages buffered, the peer allocates and commits the same pages, and the handle's next transaction reads the peer's value; without the discard it reads its own. `a_sync_refuses_a_commit_without_an_allocator_state`: a peer's commit made with the record skipped is refused by the next `begin_write()` as corruption. `syncing_to_a_peers_commit_releases_the_repair_a_panic_latched`: after a panic-dropped transaction latched a repair, syncing to a peer's commit releases it and the next commit records again. `a_transaction_releases_the_repair_a_panic_latched_without_a_peers_commit`: with no peer commit to sync to, the handle's own next transaction releases it, and a peer syncs to the commit that follows; without the release that commit records nothing and the peer is refused (the Codex thread here). `a_lent_header_lock_covers_syncing_to_a_peers_commit`: a transaction lent compaction's exclusive header lock syncs to a peer's commit, savepoints included, in a thread so that taking the lock again fails rather than hangs. `syncing_to_a_peers_close_keeps_the_recovery_flag`: a peer's close clears the flag in the file, the handle's next commit writes it back, and the handle's close clears it; without the mark the commit leaves it cleared. Not covered: a load or a sync that fails part way, which needs corruption the tree walk reports rather than panics on, or a lock that fails, neither of which the suite can inject under a live handle; nor the writer byte failing at the close, for the same reason; nor a panic in the sync, verified by hand with panics injected into the loader, after which the next `begin_write()` returns, refused until a reopen, where it hung or went ahead on an empty allocator before. The full local check set passes.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1,6 +1,8 @@
 use crate::io;
 use crate::transaction_tracker::{TransactionId, TransactionTracker};
 #[cfg(feature = "experimental-multiprocess")]
+use crate::transactions::AllocatorStateLatch;
+#[cfg(feature = "experimental-multiprocess")]
 use crate::tree_store::HeaderGuard;
 use crate::tree_store::LocklessBackend;
 #[cfg(not(redb_no_std))]
@@ -910,7 +912,13 @@ impl Database {
     // Synchronizes the tracker state for the file's persistent savepoints
     fn sync_persistent_savepoints(&self) -> Result<(), DatabaseError> {
         let txn = self.begin_write().map_err(|e| e.into_storage_error())?;
-        sync_persistent_savepoints(&self.transaction_tracker, &self.mem, &txn)?;
+        sync_persistent_savepoints(
+            &self.transaction_tracker,
+            &self.mem,
+            &txn,
+            #[cfg(feature = "experimental-multiprocess")]
+            None,
+        )?;
         txn.abort()?;
 
         Ok(())
@@ -978,6 +986,19 @@ impl Database {
         // be diagnosed before read references, because every live savepoint also holds a read
         // reference. The tracker covers persistent savepoints created by previous Database
         // instances, because they are re-registered when the database is opened.
+        // In multi-writer mode the tracker's savepoints go stale: a peer may have deleted one
+        // since this handle last synced, and only a write transaction of this handle's own syncs
+        // them again. Refresh before reporting one as a blocker, unless this process already has
+        // a write transaction live, since beginning another would block on it.
+        #[cfg(feature = "experimental-multiprocess")]
+        if self.mem.concurrency_mode() == ConcurrencyMode::MultiWriterProcess
+            && self.transaction_tracker.any_persistent_savepoint_exists()
+            && !self.transaction_tracker.write_transaction_live()
+        {
+            self.begin_write()
+                .map_err(|e| e.into_storage_error())?
+                .abort()?;
+        }
         if self.transaction_tracker.any_persistent_savepoint_exists() {
             return Err(CompactionError::PersistentSavepointExists);
         }
@@ -1012,6 +1033,8 @@ impl Database {
             .begin_write_with(
                 #[cfg(feature = "experimental-multiprocess")]
                 writer_lock.as_ref(),
+                #[cfg(feature = "experimental-multiprocess")]
+                header_lock.as_ref(),
             )
             .map_err(|e| e.into_storage_error())?;
         // Re-check inside the write transaction: a concurrent writer may have created a
@@ -1054,6 +1077,8 @@ impl Database {
                 .begin_write_with(
                     #[cfg(feature = "experimental-multiprocess")]
                     writer_lock.as_ref(),
+                    #[cfg(feature = "experimental-multiprocess")]
+                    header_lock.as_ref(),
                 )
                 .map_err(|e| e.into_storage_error())?;
             txn.skip_allocator_state_record();
@@ -1117,6 +1142,8 @@ impl Database {
                 .begin_write_with(
                     #[cfg(feature = "experimental-multiprocess")]
                     writer_lock,
+                    #[cfg(feature = "experimental-multiprocess")]
+                    header_lock,
                 )
                 .map_err(|e| e.into_storage_error())?;
             if !force_commit && !txn.pending_free_pages()? {
@@ -1407,6 +1434,26 @@ impl Database {
         root.map(|header| BtreeHeader::new(header.root, header.checksum, length))
     }
 
+    /// Loads the allocator state that the file's latest commit recorded. In multi-writer mode
+    /// every commit records one, including the repair an open performs, so a commit that
+    /// recorded none has broken the protocol and is refused as corruption.
+    #[cfg(feature = "experimental-multiprocess")]
+    fn load_synced_allocator_state(mem: &Arc<TransactionalMemory>) -> Result {
+        let Some(tree) = Self::get_allocator_state_table(mem)? else {
+            return Err(StorageError::Corrupted(
+                "The latest commit recorded no allocator state, which every commit in multi-writer mode does".to_string(),
+            ));
+        };
+        mem.load_allocator_state(&tree)?;
+        #[cfg(debug_assertions)]
+        Self::mark_allocated_page_for_debug(mem)?;
+        // This allocator state came from the file, so it holds none of the pages leaked by a
+        // transaction that was dropped while a panic unwound. Commits may record again.
+        mem.clear_needs_repair();
+
+        Ok(())
+    }
+
     fn new(
         file: Box<dyn InternalStorageBackend>,
         allow_initialize: bool,
@@ -1546,18 +1593,23 @@ impl Database {
         self.begin_write_with(
             #[cfg(feature = "experimental-multiprocess")]
             None,
+            #[cfg(feature = "experimental-multiprocess")]
+            None,
         )
     }
 
     fn begin_write_with(
         &self,
         #[cfg(feature = "experimental-multiprocess")] writer_lock: Option<&Arc<WriterLock>>,
+        #[cfg(feature = "experimental-multiprocess")] header_lock: Option<&HeaderGuard<'_>>,
     ) -> Result<WriteTransaction, TransactionError> {
         begin_write_with_allocation_policy(
             &self.transaction_tracker,
             &self.mem,
             #[cfg(feature = "experimental-multiprocess")]
             writer_lock,
+            #[cfg(feature = "experimental-multiprocess")]
+            header_lock,
             AllocationPolicy::Default,
         )
     }
@@ -1565,10 +1617,12 @@ impl Database {
 
 // Syncs the file's persistent savepoints, which another process may have created or
 // deleted since the tracker last saw the file, and continues savepoint ids past the file's.
+// `header_lock` is the header lock the caller already holds, if any.
 fn sync_persistent_savepoints(
     transaction_tracker: &TransactionTracker,
     mem: &TransactionalMemory,
     txn: &WriteTransaction,
+    #[cfg(feature = "experimental-multiprocess")] header_lock: Option<&HeaderGuard<'_>>,
 ) -> Result {
     if let Some(next_id) = txn.next_persistent_savepoint_id()? {
         transaction_tracker.restore_savepoint_counter_state(next_id);
@@ -1589,7 +1643,7 @@ fn sync_persistent_savepoints(
         current.insert(savepoint.get_id(), savepoint.get_transaction_id());
     }
     #[cfg(feature = "experimental-multiprocess")]
-    let hold = mem.header_hold(None)?;
+    let hold = mem.header_hold(header_lock)?;
     transaction_tracker.sync_persistent_savepoints(
         mem,
         &current,
@@ -1598,15 +1652,60 @@ fn sync_persistent_savepoints(
     )
 }
 
-// Takes the write slot and the writer lock, and builds the transaction on them. A caller that
-// already holds the writer lock lends it as `writer_lock`: locking the byte again from this file
-// description would return the same lock, and this transaction's end would then release the
-// caller's. The allocation policy is fixed for the lifetime of the transaction; every page
-// allocation this transaction makes goes through it.
+/// Brings this handle up to the file's latest commit, in multi-writer mode, when it is not
+/// already on it. That is the case when another process has committed since this handle last
+/// read the file, and when a panic latched a repair, which leaves this handle's allocator state
+/// holding pages that the file's latest commit does not.
+///
+/// `writer_lock` proves that the caller holds the writer byte, so that no other process commits
+/// while the header is read and the allocator state loaded from it; `header_lock` is the header
+/// lock the caller already holds, if any.
+///
+/// Returns `None` when the handle was already on the file's latest commit. Otherwise returns a
+/// latch that discards the allocator state unless the caller finishes the sync.
+#[cfg(feature = "experimental-multiprocess")]
+fn sync_to_latest_commit(
+    mem: &Arc<TransactionalMemory>,
+    writer_lock: &WriterLock,
+    header_lock: Option<&HeaderGuard<'_>>,
+) -> Result<Option<AllocatorStateLatch>> {
+    if mem.concurrency_mode() != ConcurrencyMode::MultiWriterProcess {
+        return Ok(None);
+    }
+    let synced = {
+        let hold = mem.header_hold(header_lock)?;
+        mem.reload_for_write(writer_lock, &hold)?
+    };
+    if !synced {
+        return Ok(None);
+    }
+    // Armed before the load, because a load that stops part way leaves an allocator state that
+    // describes neither the commit it came from nor the one it was going to
+    let latch = AllocatorStateLatch::arm(mem.clone());
+    Database::load_synced_allocator_state(mem)?;
+    // Loading the allocator state cleared the recovery flag that a live writer keeps set, and a
+    // peer's close may have cleared it in the file as well. Set it again, so that the next
+    // commit writes it back and a crash of this handle is still recovered from.
+    mem.mark_recovery_required();
+
+    Ok(Some(latch))
+}
+
+// Takes the write slot and the writer byte, syncs to the file's latest commit, and builds the
+// transaction on them.
+//
+// A caller that already holds the writer byte lends its lock as `writer_lock`. Taking the writer
+// byte again from this file description would return that same lock, and the end of this
+// transaction would then release the caller's. A caller that holds the header lock lends it as
+// `header_lock`, because the in-process half of that lock cannot be taken twice.
+//
+// The allocation policy is fixed for the lifetime of the transaction; every page allocation the
+// transaction makes goes through it.
 fn begin_write_with_allocation_policy(
     transaction_tracker: &Arc<TransactionTracker>,
     mem: &Arc<TransactionalMemory>,
     #[cfg(feature = "experimental-multiprocess")] writer_lock: Option<&Arc<WriterLock>>,
+    #[cfg(feature = "experimental-multiprocess")] header_lock: Option<&HeaderGuard<'_>>,
     allocation_policy: AllocationPolicy,
 ) -> Result<WriteTransaction, TransactionError> {
     // Fail early if there has been an I/O error -- nothing can be committed in that case
@@ -1616,6 +1715,8 @@ fn begin_write_with_allocation_policy(
     let slot = WriteSlot::take(transaction_tracker.clone());
     #[cfg(feature = "experimental-multiprocess")]
     let writer_lock = writer_lock.map_or_else(|| mem.lock_writer(), |lent| Ok(lent.clone()))?;
+    #[cfg(feature = "experimental-multiprocess")]
+    let latch = sync_to_latest_commit(mem, &writer_lock, header_lock)?;
     // Re-checked after acquiring the write slot: the writer this call blocked on can fail its
     // commit, latching an I/O error and discarding the allocator state. The I/O check comes
     // first so a backend failure is not misreported as corruption
@@ -1626,19 +1727,37 @@ fn begin_write_with_allocation_policy(
         )
         .into());
     }
+    // Issued once the sync is done, so that the id follows the file's latest commit
+    let transaction_id = transaction_tracker.issue_write_transaction_id(
+        mem.get_last_committed_transaction_id()?,
+        #[cfg(feature = "experimental-multiprocess")]
+        &writer_lock,
+    );
     let guard = TransactionGuard::new_write(
-        transaction_tracker.issue_write_transaction_id(),
+        transaction_id,
         slot,
         #[cfg(feature = "experimental-multiprocess")]
         writer_lock,
     );
-    WriteTransaction::new(
+    let transaction = WriteTransaction::new(
         guard,
         transaction_tracker.clone(),
         mem.clone(),
         allocation_policy,
-    )
-    .map_err(|e| e.into())
+    )?;
+    // The file's persistent savepoints are synced here, before this transaction frees anything,
+    // so that the transactions they point at stay pinned. If the sync fails, the transaction is
+    // aborted and the latch then discards the allocator state.
+    #[cfg(feature = "experimental-multiprocess")]
+    if latch.is_some() {
+        sync_persistent_savepoints(transaction_tracker, mem, &transaction, header_lock)?;
+    }
+    #[cfg(feature = "experimental-multiprocess")]
+    if let Some(latch) = latch {
+        latch.disarm();
+    }
+
+    Ok(transaction)
 }
 
 // Records the allocator state in a commit that also trims the file: the close's last commit,
@@ -1662,6 +1781,8 @@ fn ensure_allocator_state_table_and_trim(
         mem,
         #[cfg(feature = "experimental-multiprocess")]
         writer_lock,
+        #[cfg(feature = "experimental-multiprocess")]
+        header_lock,
         AllocationPolicy::Lowest,
     )
     .map_err(|e| e.into_storage_error())?;
@@ -1700,7 +1821,7 @@ fn close_database(transaction_tracker: &Arc<TransactionTracker>, mem: &Arc<Trans
     };
     #[cfg(feature = "experimental-multiprocess")]
     let writing = writing && writer_lock.is_some();
-    if writing
+    let recorded = writing
         && ensure_allocator_state_table_and_trim(
             transaction_tracker,
             mem,
@@ -1709,11 +1830,16 @@ fn close_database(transaction_tracker: &Arc<TransactionTracker>, mem: &Arc<Trans
             #[cfg(feature = "experimental-multiprocess")]
             None,
         )
-        .is_err()
-    {
+        .is_ok();
+    if writing && !recorded {
         #[cfg(feature = "logging")]
         warn!("Failed to write allocator state table. Repair may be required at restart.");
     }
+    // The shutdown header describes this handle, which matches the file only once the commit
+    // above has synced to the file's latest commit. Without that commit, write no header, and
+    // leave the file for the next open to recover.
+    #[cfg(feature = "experimental-multiprocess")]
+    let writer_lock = writer_lock.filter(|_| recorded);
 
     if mem
         .close(
@@ -2536,6 +2662,82 @@ mod writer_byte_test {
         );
     }
 
+    /// Every commit in multi-writer mode records the allocator state. A sync that finds a
+    /// commit without one refuses it as corruption.
+    #[test]
+    fn a_sync_refuses_a_commit_without_an_allocator_state() {
+        use crate::{StorageError, TransactionError};
+
+        let tmpfile = crate::create_tempfile();
+        let db = create(tmpfile.path(), ConcurrencyMode::MultiWriterProcess);
+        let mut builder = Database::builder();
+        builder.set_concurrency_mode(ConcurrencyMode::MultiWriterProcess);
+        let peer = builder.open(tmpfile.path()).unwrap();
+        let mut write = peer.begin_write().unwrap();
+        write.skip_allocator_state_record();
+        write.open_table(TABLE).unwrap().insert(0, 0).unwrap();
+        write.commit().unwrap();
+
+        assert!(matches!(
+            db.begin_write(),
+            Err(TransactionError::Storage(StorageError::Corrupted(_)))
+        ));
+    }
+
+    /// A transaction dropped while a panic unwinds leaks its pages, which latches a repair.
+    /// Syncing to a peer's commit replaces the allocator state with the file's, which does not
+    /// hold those pages, so the next commit records the state again.
+    #[test]
+    fn syncing_to_a_peers_commit_releases_the_repair_a_panic_latched() {
+        let tmpfile = crate::create_tempfile();
+        let db = create(tmpfile.path(), ConcurrencyMode::MultiWriterProcess);
+        let mut builder = Database::builder();
+        builder.set_concurrency_mode(ConcurrencyMode::MultiWriterProcess);
+        let peer = builder.open(tmpfile.path()).unwrap();
+        let unwound = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let write = db.begin_write().unwrap();
+            write.open_table(TABLE).unwrap().insert(1, 1).unwrap();
+            panic!("unwinding through the transaction");
+        }));
+        assert!(unwound.is_err());
+        assert!(db.mem.needs_repair());
+
+        commit_one(&peer);
+        commit_one(&db);
+        assert!(!db.mem.needs_repair());
+        assert!(
+            Database::get_allocator_state_table(&db.mem)
+                .unwrap()
+                .is_some()
+        );
+    }
+
+    /// A commit made while a repair is latched records no allocator state, and a peer that
+    /// syncs to such a commit refuses it. So the first transaction after the panic releases the
+    /// latch itself, without waiting for a peer to commit.
+    #[test]
+    fn a_transaction_releases_the_repair_a_panic_latched_without_a_peers_commit() {
+        let tmpfile = crate::create_tempfile();
+        let db = create(tmpfile.path(), ConcurrencyMode::MultiWriterProcess);
+        let mut builder = Database::builder();
+        builder.set_concurrency_mode(ConcurrencyMode::MultiWriterProcess);
+        let peer = builder.open(tmpfile.path()).unwrap();
+        let unwound = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let write = db.begin_write().unwrap();
+            write.open_table(TABLE).unwrap().insert(1, 1).unwrap();
+            panic!("unwinding through the transaction");
+        }));
+        assert!(unwound.is_err());
+        assert!(db.mem.needs_repair());
+
+        commit_one(&db);
+        assert!(!db.mem.needs_repair());
+        peer.begin_write()
+            .expect("the peer syncs to a commit that recorded the allocator state")
+            .abort()
+            .unwrap();
+    }
+
     /// A multi-writer compaction ends with a commit recording the allocator state, for the next
     /// writer, in any process, to load rather than rebuild
     #[test]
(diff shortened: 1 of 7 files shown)


src/db.rs
```diff
@@ -1675,18 +1741,26 @@ fn begin_write_with_allocation_policy(
...
         mem.clone(),
         allocation_policy,
-    )
-    .map_err(|e| e.into())
+    )?;
+    // The savepoints the adopted commit's tables hold pin their transactions from here, ahead
+    // of anything this transaction frees
+    #[cfg(feature = "experimental-multiprocess")]
+    if adopted {
+        register_persistent_savepoints(transaction_tracker, mem, &transaction)?;
```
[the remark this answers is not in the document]
cberner: Fixed in a15f291: the lent header hold reaches the registration. `register_persistent_savepoints()` takes the `Option<&HeaderGuard>` its caller holds, and `hold_persistent_savepoints()` takes the guard as proof rather than locking. `a_lent_header_hold_covers_adopting_a_peers_commit` adopts a peer's commit, savepoint included, under a lent exclusive hold in a thread; with the registration taking a hold of its own it deadlocks and the test fails on its timeout.

---

src/db.rs
```diff
@@ -1465,6 +1450,27 @@ impl Database {
...
+    fn load_or_rebuild_allocator_state(mem: &Arc<TransactionalMemory>) -> Result {
+        if let Some(tree) = Self::get_allocator_state_table(mem)? {
+            mem.load_allocator_state(&tree)?;
+            #[cfg(debug_assertions)]
+            Self::mark_allocated_page_for_debug(mem)?;
+        } else {
+            Self::rebuild_allocator_state(mem, &|_| {}).map_err(|err| match err {
+                DatabaseError::Storage(err) => err,
+                _ => unreachable!(),
+            })?;
```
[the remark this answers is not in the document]
cberner: Fixed in a15f291: `adopt_peer_commit()` discards the allocator state when `load_or_rebuild_allocator_state()` fails, as a failed check does, so the next transaction is refused until a reopen repairs the file. Not covered by a test: the suite has no way to inject, under a live handle, corruption the tree walk reports rather than panics on.

---

src/db.rs
```diff
@@ -1605,65 +1611,125 @@ impl Database {
...
+            let deferred = transaction_tracker.end_write_transaction(transaction_id, None);
+            assert!(deferred.is_none());
+            return Err(err.into());
+        }
+    };
+    // In multi-writer mode, another process may have committed since this handle last loaded
+    // the file: its commit is the one this transaction follows
+    #[cfg(feature = "experimental-multiprocess")]
+    let (transaction_id, adopted) =
+        match adopt_peer_commit(transaction_tracker, mem, transaction_id, header_lock) {
```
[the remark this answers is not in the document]
cberner: Fixed in a15f291: `adopt_peer_commit()` returns at once unless the mode is `MultiWriterProcess`, so the other modes read no header per transaction.

---

src/db.rs
```diff
@@ -1652,18 +1757,26 @@ fn begin_write_with_allocation_policy(
...
         mem.clone(),
         allocation_policy,
-    )
-    .map_err(|e| e.into())
+    )?;
+    // The savepoints the adopted commit's tables hold pin their transactions from here, ahead
+    // of anything this transaction frees
+    #[cfg(feature = "experimental-multiprocess")]
+    if adopted {
+        register_persistent_savepoints(transaction_tracker, mem, &transaction, header_lock)?;
```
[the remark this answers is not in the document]
cberner: Reproduced first: without a refusal, this scenario panics in the tracker when the ephemeral savepoint is dropped, on the reference the registration had already released. Fixed in 772d0ef: ephemeral savepoints are refused in `MultiWriterProcess`, with `SavepointError::EphemeralSavepointUnsupported`, as the design lists them unsupported there; `persistent_savepoint()` creates its savepoint through the same path as before. `a_multi_writer_mode_refuses_an_ephemeral_savepoint` covers the refusal.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -1299,6 +1299,39 @@ impl TransactionalMemory {
...
+            .map_err(|err| match err {
+                DatabaseError::Storage(err) => err,
+                other => StorageError::Corrupted(other.to_string()),
+            })?;
+        let (header, _) = unrepaired.finalize(self.storage.raw_file_len()?)?;
+        if *header.primary_slot() == *self.state.lock().unwrap().header.primary_slot() {
+            return Ok(false);
+        }
+
+        self.storage.invalidate_cache_all();
```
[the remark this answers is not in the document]
cberner: The buffer holds nothing of an aborted transaction: `rollback_all()` frees every page the transaction allocated, and `free_helper()` cancels the pending write of each page it frees. A committed transaction's pages are flushed by the commit, since a shared mode refuses `Durability::None`. `a_write_transaction_reads_a_peers_page_over_its_own_aborted_write` (772d0ef) is the scenario: the handle aborts a transaction that wrote a page, the peer allocates the same page and commits, and the handle's next transaction reads the peer's value. It fails with the cancellation removed from `free_helper()`, so it guards this.

---

src/db.rs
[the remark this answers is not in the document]
cberner: Fixed in 772d0ef: a registration that fails part way aborts the transaction and discards the allocator state, as a rebuild that fails part way does, so the next transaction is refused until a reopen holds every savepoint again. Not covered by a test: the suite cannot inject a malformed record or a failing lock under a live handle.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -1299,6 +1299,39 @@ impl TransactionalMemory {
...
+        {
+            let mut state = self.state.lock().unwrap();
+            state.header = header;
```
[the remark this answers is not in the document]
cberner: Reproduced first: after a peer's close, the handle's next commit wrote the flag cleared. The cause is one step past this line: keeping the flag across the adoption is not enough, since `load_allocator_state()` clears it in memory, as it does for the open ahead of `begin_writable()`. Fixed in 84ad072: `adopt_peer_commit()` marks the header a live writer's again after the load, so the commit that follows writes the flag back, as it would have without the adoption, and the close clears it as before. `adopting_a_peers_close_keeps_the_recovery_flag` reads the flag from the file after the peer's close, the commit and the close.

---

src/transactions.rs
```diff
@@ -975,6 +977,11 @@ impl WriteTransaction {
             poisoned: AtomicBool::new(false),
             durability: InternalDurability::Immediate,
             two_phase_commit: false,
+            // A multi-writer commit records the allocator state, for the next writer, in any
+            // process, to load rather than rebuild from the trees
+            #[cfg(feature = "experimental-multiprocess")]
+            quick_repair: mem.concurrency_mode() == ConcurrencyMode::MultiWriterProcess,
```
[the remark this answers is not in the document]
cberner: Fixed in 8387edb: `set_quick_repair(false)` has no effect in `MultiWriterProcess`, as `set_two_phase_commit(false)` has none in the shared modes, and its doc says so; compaction's commits opt out through a crate-private `skip_allocator_state_record()` instead. `a_multi_writer_commit_records_the_allocator_state_regardless` commits with quick-repair turned off and finds the record; with the setter honouring the call, the record is missing.

---

src/tree_store/page_store/page_manager.rs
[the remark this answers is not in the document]
cberner: Reproduced first, for the caught-panic case: the transaction's pages stayed buffered, the peer allocated and committed the same pages, and the handle's next transaction read its own. Fixed in 5d280e1: `reload_for_write()` discards the write buffer before installing the peer's state, as `clear_cache_and_reload()` does, since a buffered write can only belong to the state being replaced. `a_write_transaction_reads_a_peers_page_over_a_page_a_panic_left_buffered` is that scenario. A commit that fails on an I/O error latches it, so no transaction begins on the handle after that and nothing is adopted; the discard covers a failure of any other kind.

---

src/db.rs
```diff
@@ -1466,6 +1469,27 @@ impl Database {
         root.map(|header| BtreeHeader::new(header.root, header.checksum, length))
     }
 
+    /// Loads the allocator state the file's latest commit recorded, or rebuilds it from the
+    /// trees where that commit recorded none: a repair's, or one made while the state needed
+    /// repair
+    #[cfg(feature = "experimental-multiprocess")]
+    fn load_or_rebuild_allocator_state(mem: &Arc<TransactionalMemory>) -> Result {
+        if let Some(tree) = Self::get_allocator_state_table(mem)? {
+            mem.load_allocator_state(&tree)?;
```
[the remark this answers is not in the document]
cberner: Reproduced first: after a transaction dropped in a panic, the adopting handle's next commit recorded no allocator state. Fixed in a73b558: `load_or_rebuild_allocator_state()` releases the latch after a load as after a rebuild, since the leak it latched is in the state the load replaced. `adopting_a_peers_commit_releases_the_repair_a_panic_latched` is the scenario, and its next commit records again.

---

src/db.rs
```diff
@@ -1733,14 +1846,23 @@ fn ensure_allocator_state_table_and_trim(
...
-    if !crate::panicking()
-        && !mem.needs_repair()
+    // of trusting the saved one. Nor after a latched I/O failure, decided ahead of the lock,
+    // which waits on a peer holding the writer byte
+    let writing = !crate::panicking() && !mem.needs_repair() && mem.check_io_errors().is_ok();
+    // One hold across the allocator state's commit and the shutdown header: a commit another
+    // process made between them would be overwritten by the header
+    #[cfg(feature = "experimental-multiprocess")]
+    let writer_lock = if writing {
+        mem.lock_writer().ok()
```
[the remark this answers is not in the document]
cberner: Confirmed by reading: `begin_write_with_allocation_policy()` takes a lock of its own where it is lent none, and the transaction's end releases it. Fixed in a73b558: without the byte the close writes neither. `close_database()` counts a failed `lock_writer()` as not writing, and `TransactionalMemory::close()` takes the lock it writes the shutdown header under, or writes none, so the next open recovers from the file as the last commit left it. Not covered by a test: the suite cannot make the byte's lock fail.

---

src/transaction_tracker.rs
```diff
@@ -302,31 +321,50 @@ impl TransactionTracker {
...
+                continue;
+            }
+            state.reference_transaction(
+                mem,
+                transaction,
+                #[cfg(feature = "experimental-multiprocess")]
+                header,
+            )?;
+            if let Some(replaced) = state.valid_savepoints.insert(id, transaction) {
+                state.dereference_transaction(mem, replaced);
```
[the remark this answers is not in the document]
cberner: Real on this head, and fixed in the next PR of the stack, #1454, which was split out of this one deliberately: `restore_savepoint()` checks a persistent savepoint's handle against the file's record, its transaction and its user root, and refuses one the file no longer matches as invalid, while a handle read from the file restores as before. `a_savepoint_handle_a_peer_replaced_under_its_id_is_invalid` there is this scenario, a file copied in that carries a savepoint under the handle's savepoint and transaction ids with another root; it fails on this head. Left open for @cberner to decide whether to fold #1454 into this PR or merge them in order.

---

src/db.rs
```diff
@@ -1540,65 +1564,148 @@ impl Database {
...
+            let deferred = transaction_tracker.end_write_transaction(transaction_id, None);
+            assert!(deferred.is_none());
+            return Err(err.into());
+        }
+    };
+    // In multi-writer mode, another process may have committed since this handle last loaded
+    // the file: its commit is the one this transaction follows
+    #[cfg(feature = "experimental-multiprocess")]
+    let (transaction_id, adopted) =
+        match adopt_peer_commit(transaction_tracker, mem, transaction_id, header_lock) {
```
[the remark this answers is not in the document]
cberner: Reproduced first, with a panic injected into the allocator loader: the next `begin_write()` on another clone never returned. Fixed in bc966c4: the transaction's guard is built as soon as the writer lock is taken, before the adoption, so a failure or a panic in it drops the guard and releases the slot; `adopt_peer_commit()` takes the guard and renumbers it, through `TransactionGuard::follow()`, when it moves the transaction past the peer's commit. With the injection the next `begin_write()` now returns at once, refused until a reopen since the allocator state was dropped mid-load, as after a load that fails. Not covered by a test: the suite cannot inject a panic into the adoption.

---

src/db.rs
```diff
@@ -1583,15 +1633,55 @@ fn register_persistent_savepoints(
...
+        let hold = mem.header_hold(header_lock)?;
+        mem.reload_for_write(&hold)?
+    };
+    if !adopted {
+        return Ok(false);
+    }
+    if let Err(err) = Database::load_or_rebuild_allocator_state(mem) {
+        // Rebuilt part way, the allocator state describes neither the file nor anything else,
+        // and holding one must continue to mean it describes the file
+        mem.invalidate_allocator_state();
```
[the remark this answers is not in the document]
cberner: Reproduced first, with a panic injected into the adoption's rebuild after the fresh allocator state was installed: the next `begin_write()` went ahead on that empty state. Fixed in e133ecd: the commit path's `AllocatorStateLatch` is armed around the load or rebuild in `adopt_peer_commit()` and around the savepoint registration in the builder, and disarmed only once each completes, so an error or a panic the caller catches discards the allocator state either way; the next `begin_write()` is then refused until a reopen, as after a load that fails. With the injection the next `begin_write()` now returns that refusal. Not covered by a test: the suite cannot inject a panic into the rebuild.

---

src/db.rs
```diff
@@ -1694,7 +1826,13 @@ fn close_database(transaction_tracker: &Arc<TransactionTracker>, mem: &Arc<Trans
         warn!("Failed to write allocator state table. Repair may be required at restart.");
     }
 
-    if mem.close().is_err() {
+    if mem
+        .close(
+            #[cfg(feature = "experimental-multiprocess")]
+            writer_lock.as_ref(),
+        )
```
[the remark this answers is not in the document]
cberner: Reproduced first: with the peer's primary slot checksum flipped on disk, so that the close's commit could not adopt it, the close wrote this handle's stale header over it, and a reopen would have lost the peer's row. Fixed in 8b2431f: `close_database()` hands the writer lock to `close()` only when the allocator state commit succeeded, so without the commit no shutdown header is written, and the next open recovers from the file as the last commit left it. `a_close_writes_no_header_over_a_commit_it_could_not_adopt` is that scenario, reading the slot's bytes back after the close; they changed before the fix.

---

src/transaction_tracker.rs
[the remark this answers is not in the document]
cberner: Closing this out: the collision needs a savepoint replaced under its id, which a persisted counter that never reuses an id makes impossible within one file's history. It takes a file replaced from a copy under a live handle, and externally modified files are not something redb guards against; the behaviour after one is the user's. #1454, which guarded this, is closed for the same reason.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -1304,6 +1304,40 @@ impl TransactionalMemory {
         Ok(current)
     }
 
+    /// Adopts the commit another process made since this handle last loaded the file, ahead of
+    /// a write transaction: the header, with the cached pages dropped, since that process may
+    /// have reused any of them, and the allocator state dropped for the caller to load or
+    /// rebuild. Under the header lock, which `header_lock` proves held, and the writer byte,
+    /// which the caller holds too, so the commit adopted is the one the transaction follows.
+    /// Returns whether there was one
```
cberner: This is quite long. Either it needs to be trimmed, or we need to simplify the function so that this amount of explaining is unecessary

cberner (commented):
This needs a serious pass on the comments and/or simplifying to make that many comments unnecessary. I'll look at it again when that is done.

src/tree_store/page_store/page_manager.rs
cberner: Trimmed to three lines in 3f25d8e, and the rest of the PR's comments with it: each is one or two lines now, and the adoption's two latches are one, returned by `adopt_peer_commit()` and disarmed once the savepoints are synced.

---

cberner:
The comment pass is in 3f25d8e. Every comment the PR adds is one or two lines now, and the code lost the piece that needed the most explaining: the adoption's two latches are one, returned by `adopt_peer_commit()` and disarmed once the savepoints are synced, so the builder no longer arms and aborts by hand. #1443 had the same pass.

---

src/db.rs
```diff
@@ -1583,15 +1633,50 @@ fn sync_persistent_savepoints(
     )
 }
 
+/// In multi-writer mode, adopts the commit another process made since this handle last loaded
+/// the file, if there is one, and renumbers the write transaction `guard` holds past it. Under
+/// the writer byte, and the header lock where the caller lends it as `header_lock`. Returns the
+/// latch that discards the allocator state unless the caller completes the adoption
+#[cfg(feature = "experimental-multiprocess")]
+fn adopt_peer_commit(
```
cberner: Let's use sync rather than adopt as the terminology

src/transaction_tracker.rs
```diff
@@ -294,6 +294,25 @@ impl TransactionTracker {
         state.next_transaction_id = id;
     }
 
+    // Renumbers the live write transaction `id` past `last_committed`, a commit another process
+    // made after the id was issued
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn follow_committed_transaction(
```
cberner: Follow implies a leader and followers. Use another term

src/transactions.rs
```diff
@@ -895,17 +895,17 @@ impl SavepointTransactionState {
 // Discards the in-memory allocator state on drop, unless disarmed. An incomplete commit may
 // have returned pages to the allocator that the durable roots still reference through the
 // freed tables; such an allocator state must never be used, or persisted by a clean
-// shutdown, again.
-struct AllocatorStateLatch {
+// shutdown, again. Nor may one that a load, a rebuild or a savepoint sync left part way.
```
cberner: This is hard to follow with all the negatives

cberner (commented):
I skimmed it. Address the comments then I'll look again

src/db.rs
cberner: Renamed throughout in bc10699: `sync_to_peer_commit()`, and the comments, tests and messages say sync rather than adopt, in #1443 as well.

---

src/transaction_tracker.rs
cberner: Now `renumber_past_commit()`, and the guard's `renumber()`, in bc10699.

---

src/transactions.rs
cberner: Rewritten in bc10699 to say what the latch is armed around and what dropping it armed does, without the negatives.

---

cberner:
All three addressed in bc10699: sync is the terminology, in the function names, comments, tests and messages here and in #1443; the renumbering is `renumber_past_commit()`; and the latch's comment is rewritten in the positive. Same code otherwise.

---

src/transaction_tracker.rs
```diff
@@ -294,6 +294,25 @@ impl TransactionTracker {
         state.next_transaction_id = id;
     }
 
+    // Renumbers the live write transaction `id` past `last_committed`, a commit another process
+    // made after the id was issued
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn renumber_past_commit(
```
cberner: Hang on, what? How is this possible? Is this another thing that only happens if the user messes with the file directly?

cberner: No, this is the ordinary multi-writer case. The write slot issues the transaction's id from this process's counter, in `start_write_transaction()`, before the writer byte is taken and before the sync. The counter continues from the last commit this process knows: its own, or the last peer commit it synced to. So any peer commit since then is one the counter has not seen, whether the peer made it while this handle sat idle between transactions or while this transaction waited for the byte.

Two processes open at commit 5. B commits 6. A begins a write: its slot issues 6, A takes the byte, syncs to B's commit 6, and its transaction becomes 7. Left at 6, its commit would land under the same id as B's, which recovery orders the commit slots by.

The id is issued at the slot rather than after the sync because the slot has to be taken first, before the byte, and the slot's owner is known by its id: the deferred close at `Database::drop()` and the tracker's own assertions reason about it. Reserving the slot without an id until the byte is held is what the staging branch did. I'll make the comment say this rather than "after the id was issued".

---

cberner: That seems wrong to issue an id that is not synced. I think we should issue it after the sync

cberner: Agreed, changing it: the slot is taken without an id, the sync moves the counter past the file's latest commit, and the id is issued from it after that. The renumbering goes away.

---

cberner: Done in 72b9cc4. The tracker's write slot is `Free`, `Taken` or `Live(id)`: `take_write_slot()` takes it without an id, and `issue_write_transaction_id(last_committed)` issues one past the file's latest commit once the sync is done, so the id is never unsynced. The guard holds the slot and the byte only; `WriteTransaction::new()` takes the id. `renumber_past_commit()` and the guard's `renumber()` are gone, a net cut of sixteen lines.

---

src/db.rs
```diff
@@ -369,7 +371,6 @@ pub(crate) enum TransactionGuard {
     },
     Write {
         tracker: Arc<TransactionTracker>,
-        transaction_id: TransactionId,
```
cberner, line 392: I'm sorry, but what? This doesn't seem right at all. Starting a write txn should take the locks, sync, and then create the txn guard and other txn objects

cberner: Reordered in cec84fd to that: the slot and the byte are taken, the sync runs, and then the guard is built with its id and the transaction from it. The slot is a small hold, `WriteSlotHold`, that releases it on drop and runs any deferred close, so a start that fails part way leaves nothing held; the guard's `Write` variant carries the id, the byte and that hold, dropped in that order. `WriteTransaction::new()` reads the id from the guard as before.

---

src/db.rs
```diff
@@ -358,6 +360,27 @@ impl CacheStats {
     }
 }
 
+// The write slot, taken ahead of the writer byte and held until the transaction ends. Releases
+// it on drop, and performs the close a `Database::drop()` deferred to that end
+pub(crate) struct WriteSlotHold {
```
cberner: struct names should be nouns, so "hold" is not good to have in a name. Do a cleanup pass on the naming in this PR

src/db.rs
```diff
@@ -1402,6 +1421,26 @@ impl Database {
...
+    fn load_or_rebuild_allocator_state(mem: &Arc<TransactionalMemory>) -> Result {
+        if let Some(tree) = Self::get_allocator_state_table(mem)? {
+            mem.load_allocator_state(&tree)?;
+            #[cfg(debug_assertions)]
+            Self::mark_allocated_page_for_debug(mem)?;
+        } else {
+            Self::rebuild_allocator_state(mem, &|_| {}).map_err(|err| match err {
+                DatabaseError::Storage(err) => err,
+                _ => unreachable!(),
+            })?;
```
cberner: How is it possible to reach this? Quick repair is always used, so it seems like this is an error path from the sync peer function. If so, we should just raise an error, because the protocol has been violated

cberner: It is reachable today without a violation: the open's repair commit records no allocator state, and a create's first commit is a repair's, so a peer syncing to a file another process had just created or repaired found no snapshot and rebuilt. `a_write_transaction_rebuilds_the_allocator_a_peers_repair_left_unsaved` was that case.

Changing it as you say: the multi-writer open ends a repair with a commit that records the allocator state, as the check's repair does since #1443 and the close does, so no commit in a multi-writer file lacks a snapshot, and this branch becomes a `Corrupted` error. A unit test commits from a peer with the record skipped and has the sync refuse it.

---

src/db.rs
```diff
@@ -1467,6 +1506,14 @@ impl Database {
 
         // Restore the tracker state for any persistent savepoints
         db.sync_persistent_savepoints()?;
+        // In multi-writer mode a repair ends with a commit recording the allocator state, as the
+        // check's does, so that every commit a peer syncs to carries one
+        #[cfg(feature = "experimental-multiprocess")]
+        if repaired && concurrency_mode == ConcurrencyMode::MultiWriterProcess {
+            ensure_allocator_state_table_and_trim(&db.transaction_tracker, &db.mem, None, None)?;
+        }
```
cberner: This looks like a standalone fix. If so, split it out into a prep PR.

src/db.rs
```diff
@@ -358,6 +360,27 @@ impl CacheStats {
     }
 }
 
+// The write slot, taken ahead of the writer byte and held until the transaction ends. Releases
+// it on drop, and performs the close a `Database::drop()` deferred to that end
+pub(crate) struct WriteSlot {
```
cberner, line 365: Ok ya, I like this WriteSlot and WriteSlotState refactoring, but let's move it into a prep PR

cberner: Moved into #1460, which this PR is restacked on.

---

src/db.rs
cberner: Split out into #1461, on #1460; this PR is restacked on both.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -1304,6 +1304,35 @@ impl TransactionalMemory {
...
+    /// dropping the cached pages, the buffered writes and the allocator state with the old one;
+    /// the caller reloads the allocator. Under the header lock and the writer byte
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn reload_for_write(&self, header_lock: &HeaderGuard<'_>) -> Result<bool> {
+        let unrepaired = Self::read_file_header(&self.storage, self.page_size, header_lock)
+            .map_err(DatabaseError::into_storage_error_or_corrupted)?;
+        let (header, _) = unrepaired.finalize(self.storage.raw_file_len()?)?;
+        // Ids only advance, every writer reloading under the byte, so a commit is known by its id
+        if header.primary_slot().transaction_id == self.get_last_durable_transaction_id()? {
+            return Ok(false);
```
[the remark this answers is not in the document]
cberner: Reproduced first, then fixed in 78d26c8. A transaction dropped in a panic latches the repair, the id is unchanged with no peer commit, so the sync was skipped, and the commit that followed deleted the allocator state table without recreating it. Every peer syncing to that commit was refused as corruption and stayed refused until it reopened, from one panic in one process.

The sync now runs on a latched repair whatever the id says: the leak is this handle's own, and the state its last commit recorded is free of it, so the reload discards the leak and the buffered writes exactly as a peer's commit does, and the load clears the latch for the next commit to record again. `sync_to_peer_commit()` is `sync_to_latest_commit()` accordingly. `a_transaction_releases_the_repair_a_panic_latched_without_a_peers_commit` is the scenario, asserting the peer's `begin_write()` succeeds; it fails on the previous head.

---

src/tree_store/page_store/page_manager.rs
```diff
@@ -1304,6 +1304,38 @@ impl TransactionalMemory {
...
+        // Ids only advance, every writer reloading under the byte, so a commit is known by its
+        // id. A latched repair is this handle's own leak, which the commit's state is free of
+        if header.primary_slot().transaction_id == self.get_last_durable_transaction_id()?
+            && !self.needs_repair()
+        {
+            return Ok(false);
+        }
+
+        // That process may have reused any page cached or buffered here
+        self.storage.discard_write_buffer();
```
cberner, line 1339: This seems wrong. It could lose data, no? We should instead assert that the write buffer is empty

src/tree_store/page_store/page_manager.rs
```diff
@@ -1304,6 +1304,38 @@ impl TransactionalMemory {
...
+        self.storage.invalidate_cache_all();
+        {
+            let mut state = self.state.lock().unwrap();
+            state.header = header;
+            state.read_from_secondary = false;
+            state.allocators = None;
+        }
+        // Mirrors the allocator state that was just dropped, so it is repopulated with it
+        #[cfg(debug_assertions)]
+        self.allocated_pages.lock().unwrap().clear();
```
cberner, line 1350: Similarly, shouldn't we assert that this is empty?

src/tree_store/page_store/page_manager.rs
```diff
@@ -1304,6 +1304,38 @@ impl TransactionalMemory {
...
+    /// Syncs to the file's latest commit where this handle is not on it, dropping the cached
+    /// pages, the buffered writes and the allocator state it holds; the caller reloads the
+    /// allocator. Under the header lock and the writer byte
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn reload_for_write(&self, header_lock: &HeaderGuard<'_>) -> Result<bool> {
+        let unrepaired = Self::read_file_header(&self.storage, self.page_size, header_lock)
+            .map_err(DatabaseError::into_storage_error_or_corrupted)?;
+        let (header, _) = unrepaired.finalize(self.storage.raw_file_len()?)?;
+        // Ids only advance, every writer reloading under the byte, so a commit is known by its
+        // id. A latched repair is this handle's own leak, which the commit's state is free of
```
cberner: "under the byte" is not a good way to phrase it. Never short "writer byte" to "byte" that is very confusing. Also this really needs work, it is very confusing to read

src/tree_store/page_store/page_manager.rs
```diff
@@ -1304,6 +1304,38 @@ impl TransactionalMemory {
         Ok(current)
     }
 
+    /// Syncs to the file's latest commit where this handle is not on it, dropping the cached
+    /// pages, the buffered writes and the allocator state it holds; the caller reloads the
+    /// allocator. Under the header lock and the writer byte
```
cberner: needs work

src/db.rs
```diff
@@ -1407,6 +1421,24 @@ impl Database {
...
+    fn load_synced_allocator_state(mem: &Arc<TransactionalMemory>) -> Result {
+        let Some(tree) = Self::get_allocator_state_table(mem)? else {
+            return Err(StorageError::Corrupted(
+                "The latest commit recorded no allocator state, which every commit in multi-writer mode does".to_string(),
+            ));
+        };
+        mem.load_allocator_state(&tree)?;
+        #[cfg(debug_assertions)]
+        Self::mark_allocated_page_for_debug(mem)?;
+        // The state is the file's now, free of the leak a panic-dropped transaction latched
```
cberner: word this more clearly

src/db.rs
```diff
@@ -1598,15 +1637,46 @@ fn sync_persistent_savepoints(
...
 
+/// In multi-writer mode, syncs to the file's latest commit where this handle is not on it:
+/// another process committed, or a panic latched a repair over the state this one holds. Under
+/// the writer byte, and the header lock where the caller lends it as `header_lock`. Returns the
+/// latch that discards the allocator state unless the caller completes the sync
+#[cfg(feature = "experimental-multiprocess")]
+fn sync_to_latest_commit(
+    mem: &Arc<TransactionalMemory>,
+    header_lock: Option<&HeaderGuard<'_>>,
+) -> Result<Option<AllocatorStateLatch>> {
```
cberner, line 1671: I think this should also take proof of a WriterLock. Otherwise, it's unclear whether the individual operations like reload_for_write() and load_synced_allocator_state() could be reading torn data from different commits if another process raced an committed (it looks like this is not possible because the caller has the writer_lock, but we should take a borrow of that to prove it)

src/db.rs
```diff
@@ -1626,19 +1698,33 @@ fn begin_write_with_allocation_policy(
         )
         .into());
     }
+    // Issued after the sync, past the file's latest commit
+    let transaction_id =
+        transaction_tracker.issue_write_transaction_id(mem.get_last_committed_transaction_id()?);
```
cberner: Seems like issue_write_transaction_id() should probably take proof of the WriterLock too

src/db.rs
```diff
@@ -1709,11 +1797,14 @@ fn close_database(transaction_tracker: &Arc<TransactionTracker>, mem: &Arc<Trans
...
             None,
         )
-        .is_err()
-    {
+        .is_ok();
+    if writing && !recorded {
         #[cfg(feature = "logging")]
         warn!("Failed to write allocator state table. Repair may be required at restart.");
     }
+    // No commit, no header: this handle's describes the file only once its commit has synced to it
```
cberner: reword this. it's poorly written

cberner (commented):
Look at the style of the last few commits that Fable made, that is how I want comments written. This PR looks like it's the right direction, but you need to do a polish and clarity pass on all the comments. Also I had some other comments

PR #1452, opened 2026-09-05
Record the allocator state on every multi-writer commit

cberner:
The first of the writer-side reload's pieces, on master; #1453, #1449, #1454 and #1443 follow it, each on the one before. The design has every multi-writer commit record the allocator state, for the next write transaction, in any process, to load rather than rebuild from the trees; a commit recorded it only where `set_quick_repair(true)` had been called, as in the other modes.

## What it does

A `MultiWriterProcess` transaction's quick-repair is on, and `set_quick_repair(false)` has no effect there, as `set_two_phase_commit(false)` has none in the shared modes; its doc says so. Compaction's own commits opt out through a crate-private `skip_allocator_state_record()`, since the record is a table each commit frees and writes anew, which the next round would move again and find pending, so neither its drain nor its page moves would end. Compaction still ends with the close's commit, which records (#1450).

1. **Enforce in the setter rather than at commit.** The field's value is then always what the commit does, and the compaction opt-out is one crate-private call rather than a second flag.
2. **Every commit, not a default.** A default an application can turn off is not the design's requirement, and the changelog's "every commit" would not be true.

## Testing

`a_multi_writer_commit_records_the_allocator_state_regardless`: a commit told to skip quick-repair records the allocator state; with the setter honouring the call, the record is missing. The full local check set passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,8 +37,8 @@
   `Durability::None` is refused, and the database may be opened read-only while another process has
   it open for writing; each new read transaction then sees that process's durable commits, and
   `Database::compact()` treats a read transaction in another process as a transaction in progress.
-  In `MultiWriterProcess`, `Database::compact()` ends with a commit recording the allocator state,
-  for the next writer to load.
+  In `MultiWriterProcess`, every commit records the allocator state, for the next writer to load;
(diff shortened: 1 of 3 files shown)


cberner:
Lgtm. I'll merge it when ci is done

PR #1453, opened 2026-09-05
Refuse ephemeral savepoints in multi-writer mode

cberner:
The second of the writer-side reload's pieces, on master now that #1452 is merged; #1449, #1454 and #1443 follow it. The design lists ephemeral savepoints as unsupported in multi-writer mode, and the API did not refuse them. An ephemeral savepoint's id is known to this process alone: a persistent savepoint another process creates can take it, and a handle that adopts that process's commit, as the reload will have every write transaction do, then holds the persistent savepoint under the id of its own ephemeral one, unpinning that snapshot and panicking when the ephemeral savepoint is dropped.

## What it does

`ephemeral_savepoint()` refuses in `MultiWriterProcess`, with a new `SavepointError::EphemeralSavepointUnsupported`, beside `SetDurabilityError::NonDurableCommitUnsupported`; its doc says so. `persistent_savepoint()` creates its savepoint through the same path as before, now `create_savepoint()`.

1. **Refuse, rather than keep ephemeral ids apart from the peers'.** The design withholds the support an id space of their own would be, and the refusal is the smaller change.
2. **A new error variant.** No existing `SavepointError` says "unsupported in this mode"; the durability refusal got a variant of its own for the same reason.

## Testing

`a_multi_writer_mode_refuses_an_ephemeral_savepoint`: refused, while a persistent savepoint is still created. `an_ephemeral_savepoint_locks_the_snapshot_it_references` runs in `SingleWriterProcess` now, the shared mode that supports one. The full local check set passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
   `Database::compact()` treats a read transaction in another process as a transaction in progress.
   In `MultiWriterProcess`, every commit records the allocator state, for the next writer to load;
   compaction's own commits are the exception, and `Database::compact()` ends with one that does.
+  Ephemeral savepoints are refused there, with `SavepointError::EphemeralSavepointUnsupported`.
 
 ### redb-derive (unreleased)
 * Fix `#[derive(Value)]` and `#[derive(Key)]` failing to compile on structs whose lifetimes are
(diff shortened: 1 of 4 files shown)


cberner:
Lgtm. I'll merge it soon

PR #1455, opened 2026-09-05
Build every write transaction on one path

cberner:
The first of two preps split out of the writer-side reload (#1449), on master. A write transaction was built in two places: `begin_write()` took the write slot and the writer lock through `write_guard()` and handed the guard to `begin_write_with_allocation_policy()`, while the allocator state commit the close and compaction make built a guard of its own, taking the lock before the slot. The reload has to adopt a peer's commit between the lock and the transaction, on every path a write transaction is built, so there should be one.

## What it does

`begin_write_with_allocation_policy()` takes the slot and the lock itself, lent the writer lock where its caller holds one, since a second lock on the byte from this file description is the same lock and the transaction's end would release the caller's; `begin_write()` and the close's and compaction's allocator state commit both begin their transactions there, and `write_guard()` is gone. The allocator rebuild and its debug marking take a shared reference to the memory rather than a `&mut` that only asserted exclusivity: no transaction is live where they run, or the caller holds the write slot, and a read transaction never touches the allocator.

1. **Slot, then lock, on both paths.** The close's helper took the lock first; the builder takes the slot first, as `begin_write()` always did, since the slot is what orders this process's writers and nothing waits on it at the close or in compaction.
2. **No behaviour change.** The I/O check, the slot, the lock and the re-check after the slot happen as before, in the same order, on both paths.

## Testing

The existing suite, including the lent-lock tests in `active_transaction_test` and the close tests in `writer_byte_test`, which cover both paths. The full local check set passes.

---

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -914,7 +914,7 @@ impl Database {
             Ok(true) => {
                 let live_allocator_hash = self.mem.allocator_hash();
                 let live_roots = [self.mem.get_data_root(), self.mem.get_system_root()];
-                match Self::rebuild_allocator_state(&mut self.mem, &|_| {}) {
+                match Self::rebuild_allocator_state(&self.mem, &|_| {}) {
                     // Only a durable root can be rewritten from here, so a live root whose table
                     // count was recomputed must be rolled back and repaired by the reload below
                     Ok(roots) if roots != live_roots => Ok(None),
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1328,8 +1326,10 @@ impl Database {
     // The returned roots carry table counts recounted from the trees that were walked. These
     // counts are stored in the commit slot rather than in a page, so no page checksum covers
     // them; recounting here is what lets the rest of the codebase trust them.
+    // The allocator state is the caller's to replace: no transaction is live, or it holds the
+    // write slot, since a read transaction never touches the allocator
```
cberner: This comment is confusing

cberner (commented):
A few of the comments need work. Otherwise this lgtm

src/db.rs
cberner: Reworded in 68fabaa: "Callers must ensure nothing else uses the allocator meanwhile: either no transaction is live, or the caller holds the write slot (read transactions never allocate or free)." The builder's comment above `begin_write_with_allocation_policy()` is reworded in the same push, since it had the same density: it now says the function takes the slot and the lock and builds the transaction on them, and that a caller already holding the writer lock lends it because locking the byte again from this file description would return the same lock.

---

PR #1457, opened 2026-09-05
Read the file's header in one place

cberner:
The first of two more preps split out of the writer-side reload (#1449), on master. The header was read from the file and parsed in five places, each a copy of the same lines: the open's two reads, the check's reload, the shared reader's reload and the corruption probe; the writer's reload in #1449 would have added a sixth.

## What it does

`TransactionalMemory::read_file_header()` reads and parses the header under the header lock its caller proves held with a `&HeaderGuard`, as a writer publishes the header under that lock, and the five readers call it: the open's two take the lock around the call as they did, the check's reload and the corruption probe take it through `lock_header_shared()`, and the shared reader's reload passes the one it is given. `DatabaseError::into_storage_error_or_corrupted()` is the reload's mapping of a parse error into the storage error its path returns, for the writer's reload in #1449 to share; a header read under a live handle can need no upgrade, since the open checked the version.

1. **Parse only.** The readers diverge right after the parse: the open reconciles the layout with the file's length, the shared reader consults neither, the probe looks at one checksum. So the shared function stops at the parsed header.
2. **The lock is the caller's.** The open holds it before there is a `TransactionalMemory`, and a reload is handed it by the transaction, so the function takes proof rather than the lock, as the header writes do.

## Testing

A refactor with no behaviour change: the two callers that took the shared lock through the associated function now take it through `lock_header_shared()`, which is that call. The existing open, reload, check and corruption tests cover every caller. The full local check set passes.

---

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -251,6 +251,16 @@ impl From<io::Error> for DatabaseError {
     }
 }
 
+impl DatabaseError {
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn into_storage_error_or_corrupted(self) -> StorageError {
+        match self {
+            DatabaseError::Storage(storage) => storage,
+            other => StorageError::Corrupted(other.to_string()),
+        }
+    }
+}
+
 impl From<StorageError> for DatabaseError {
     fn from(err: StorageError) -> DatabaseError {
         DatabaseError::Storage(err)
--- a/src/tree_store/page_store/page_manager.rs
+++ b/src/tree_store/page_store/page_manager.rs
@@ -762,6 +762,15 @@ impl TransactionalMemory {
         })
     }
 
+    fn read_file_header(
+        storage: &PagedCachedFile,
+        page_size: u32,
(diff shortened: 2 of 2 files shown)


src/error.rs
```diff
@@ -251,6 +251,18 @@ impl From<io::Error> for DatabaseError {
     }
 }
 
+impl DatabaseError {
+    // The storage error this is, or a corruption where it is of another kind: a header read
+    // under a live handle can need no upgrade, since the open checked the version
```
cberner: remove

src/tree_store/page_store/page_manager.rs
```diff
@@ -762,6 +762,18 @@ impl TransactionalMemory {
         })
     }
 
+    /// The header as the file has it, parsed and nothing more. Read under the header lock, which
+    /// `_header_lock` proves held: a writer publishes a header under the same lock, so the bytes
+    /// read are one header's
```
cberner: remove

cberner:
lgtm, except the comments that need to be removed

PR #1458, opened 2026-09-05
Hold the writer byte across the close's commit and header

cberner:
The second of two more preps split out of the writer-side reload (#1449), on master now that #1457 is merged. The close's commit took the writer byte through its transaction and released it at the commit's end, and the shutdown header was then written under the header lock alone, so a commit another process made between the two was overwritten by the header: half of the thread on #1445, the other half being the close's commit itself beginning from a stale header, which #1449 fixes.

## What it does

`close_database()` takes the writer byte once, ahead of its commit, lends it to the transaction as compaction lends its own, and keeps it for `TransactionalMemory::close()`, which writes the shutdown header under the lock it is lent, or writes none. Whether there is anything to write is decided ahead of the byte, from the panic, repair and I/O checks the two writes made separately, so a close with nothing to write does not wait on a peer's write transaction, as it did not before; and where the byte cannot be taken, neither is written, since the header would overwrite whatever a peer commits meanwhile.

1. **One hold, taken by the close.** Locks on the byte from one file description are one lock, so the transaction cannot take its own inside the close's; the close takes it and lends it, as compaction does for its transactions.
2. **No byte, no header.** The header claims a clean shutdown of the file as this handle has it; written outside the byte it could land over another process's commit, so the close without the byte leaves the flag set for the next open to recover.

## Testing

`the_shutdown_header_is_written_under_the_writer_byte_or_not_at_all`: the memory's close writes the header only when lent the byte, seen through a shared reader, which is refused an unclean file no live writer holds. `a_close_leaves_a_clean_file`: the database's close lends it. The window between the commit and the header cannot be hit deterministically without a hook in the backend, which multi-writer mode refuses, so the tests cover the contract's two halves and the code holds the byte across both by construction. The full local check set passes.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1677,14 +1677,27 @@ fn ensure_allocator_state_table_and_trim(
 // write-transaction slot.
 fn close_database(transaction_tracker: &Arc<TransactionTracker>, mem: &Arc<TransactionalMemory>) {
     // No saved allocator state when it needs repair: the next open must rebuild it instead
-    // of trusting the saved one
-    if !crate::panicking()
-        && !mem.needs_repair()
+    // of trusting the saved one. Nor after a latched I/O failure, decided ahead of the lock,
+    // which waits on a peer holding the writer byte
+    let writing = !crate::panicking() && !mem.needs_repair() && mem.check_io_errors().is_ok();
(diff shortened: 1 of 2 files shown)


cberner:
lgtm. I'll merge, as the stack goes in

PR #1461, opened 2026-09-06
Record the allocator state after a multi-writer repair

cberner:
A prep for #1449, split out of it at review, on #1460. In multi-writer mode every commit records the allocator state, since #1452, for the next write transaction in any process to load rather than rebuild; the open's repair ended without one.

## What it does

`Database::new()` ends a repair in `MultiWriterProcess` mode with `ensure_allocator_state_table_and_trim()`, the recording commit the close and compaction make, once the tracker holds the file's persistent savepoints. A create is a repair, a new file having no state to load, so a created database is recorded from its first commit too. Before this, a peer opening the file after a repair, or after a create, ahead of the handle's first commit, found no recorded state and repaired again; with #1449, a sync to that commit finds none where every other commit records one.

1. **A second commit, rather than a recording repair.** The repair's own commit is `TransactionalMemory::commit()`, made ahead of the tracker; the recording commit is a write transaction, as compaction's is, and begins once the tracker holds the file's savepoints, as every write transaction does.
2. **Multi-writer mode only.** In the other modes nothing reads the file between the repair and the close, which records, and a second record would leave the file one record larger, the pages of the one it replaces being freed only by the next commit.

## Testing

`a_multi_writer_repair_records_the_allocator_state`: a create in multi-writer mode, whose first commit is a repair's, leaves a recorded state behind it; it fails without the commit. The full local check set passes.

---

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1432,12 +1432,13 @@ impl Database {
         let mut mem = Arc::new(mem);
         // If the last transaction used 2-phase commit and updated the allocator state table, then
         // we can just load the allocator state from there. Otherwise, we need a full repair
-        if let Some(tree) = Self::get_allocator_state_table(&mem)? {
+        let repaired = if let Some(tree) = Self::get_allocator_state_table(&mem)? {
             #[cfg(feature = "logging")]
             debug!("Found valid allocator state, full repair not needed");
             mem.load_allocator_state(&tree)?;
             #[cfg(debug_assertions)]
             Self::mark_allocated_page_for_debug(&mem)?;
+            false
         } else {
             #[cfg(feature = "logging")]
             warn!("Database {:?} not shutdown cleanly. Repairing", &file_path);
@@ -1457,7 +1458,8 @@ impl Database {
                 #[cfg(feature = "experimental-multiprocess")]
                 None,
             )?;
-        }
+            true
+        };
 
         mem.begin_writable()?;
         // Past recovery, so a reader finding this byte held knows the flag means a live writer
@@ -1472,6 +1474,14 @@ impl Database {
 
         // Restore the tracker state for any persistent savepoints
         db.sync_persistent_savepoints()?;
+        // In multi-writer mode a repair ends with a commit recording the allocator state, as
+        // compaction does, so that the next open, in any process, loads it
+        #[cfg(feature = "experimental-multiprocess")]
+        if repaired && concurrency_mode == ConcurrencyMode::MultiWriterProcess {
+            ensure_allocator_state_table_and_trim(&db.transaction_tracker, &db.mem, None, None)?;
+        }
+        #[cfg(not(feature = "experimental-multiprocess"))]
+        let _ = repaired;
 
         Ok(db)
     }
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1472,6 +1474,14 @@ impl Database {
 
         // Restore the tracker state for any persistent savepoints
         db.sync_persistent_savepoints()?;
+        // In multi-writer mode a repair ends with a commit recording the allocator state, as
+        // compaction does, so that the next open, in any process, loads it
+        #[cfg(feature = "experimental-multiprocess")]
+        if repaired && concurrency_mode == ConcurrencyMode::MultiWriterProcess {
+            ensure_allocator_state_table_and_trim(&db.transaction_tracker, &db.mem, None, None)?;
```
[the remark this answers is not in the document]
cberner: Real, but it is the open protocol's race rather than this commit's, and this commit is not where it is worst. A multi-writer open holds no writer byte at all: `database_writer_lock_for_mode()` returns `None` for the mode, so two handles opening an unrecorded file already repair concurrently and land their repairs' `mem.commit()` over each other, with no byte held by either, before either reaches this call. The recording commit is the first thing in that sequence that takes the byte.

What is left of the gap here, a commit built on roots the handle read before a peer committed, is what #1449 closes: the transaction this call begins syncs to the file's latest commit under the byte it takes, so it commits from the file as the peer left it. Holding one byte across the whole open, which is what the repairs need, is the multi-writer open protocol, still to come.

---

cberner, line 1481: Do we need to hold a lock across these to make sure other processes don't start a repair or race too?

cberner: Yes, and nothing does today: `database_writer_lock_for_mode()` returns `None` for the mode, so the header read in `TransactionalMemory::new()`, the repair decision and the repair's own `mem.commit()` all run with no byte held. Two processes opening an unrecorded file both repair and commit over each other, and this recording commit is the first thing in the open that takes the byte at all.

Taking the byte across these lines would not be enough on its own, though. The header they decide from was read before the byte, so a peer can have repaired and recorded in between, and this process would repair from a stale header and commit over it. Under the byte the open has to re-read the header and re-check for the allocator state table, and repair only where it is still missing. That re-read is `reload_for_write()` from #1449, which is why I would make the open protocol the slice after it rather than fold it in here.

The design question inside it is whether every multi-writer open takes the byte, or only one that finds no recorded state. Always makes an open wait on a peer's live write transaction, which it does not do today; only-on-repair keeps that, at the cost of the re-check above. I would take it where the load finds nothing, then re-read and re-check under it. The create race on an empty file is the same protocol's other half, and `docs/design.md` says nothing about the open yet. Happy to write it up as the next PR after #1449.

---

cberner: Yes the recheck sounds good. Do that stacked on this PR. No need to restack others on it until it merges

cberner: Done, stacked on this one as #1462, and #1449 left where it is.

One correction to what I said above: every multi-writer open already takes the byte, for the transaction that syncs the file's savepoints, so taking it on the repair path adds no waiting that was not already there. Taking it for every open would cost nothing extra either, and I kept it to the repair path only so the byte's span covers the work that writes; say if you would rather have it over the whole open.

---

PR #1463, opened 2026-09-07
Give up the writer byte before the write slot

cberner:
Prep, split out of #1413. Independent of the other two (#1413's savepoint change and the open's header hold); any order.

## What it does

The write slot's state answers two different questions, and `Live` was a poor proxy for the second.

Whether a write transaction **occupies the slot** is what makes a second one block, and what `Database::compact()` asks before beginning one of its own. Whether **this process holds the writer byte** is the other. `Live` was wrong for it at both ends:

* A transaction takes the slot before acquiring the byte, so a slot that is merely `Taken` may be blocked on a peer that is committing.
* `TransactionGuard::Write` released the byte before giving up the slot -- deliberately, so that a thread waiting on the slot takes the same byte on this file description -- which left the slot reading `Live` for a byte already released.

The byte becomes a value, `LiveWriter`, which owns the lock and downgrades the slot to `Taken` from its `Drop`. `Drop` runs before a value's own fields are dropped, so the downgrade is ordered ahead of the release by this type rather than by the order of the fields around it, and the release order the guard depends on is unchanged. Whether the slot is occupied is untouched: `Taken` still blocks a second transaction, and `write_transaction_live()` still reports it, which is what `compact()` reads.

## Why now, and why nothing changes

Nothing on master reads `Live` for byte ownership yet -- `compact()`, the only other consumer, asks the occupancy question -- so this is behaviour-preserving here by construction. It is prep for the writer-side reload in #1413, whose gate skips a read's header reload while this process holds the byte; against a slot that is `Live` at either end of the byte's life, that gate reloads too little.

I left it as its own change rather than folding it into that one because the invariant belongs to the write path, and because the two questions being distinct is worth stating on its own.

## Testing

`a_write_transaction_gives_up_the_writer_byte_before_the_slot`: while a transaction is live both questions answer yes; after the byte is given up the slot is still occupied but no longer claims the byte; after the transaction ends neither holds. It fails with the downgrade removed.

The ordering against the release itself is not tested: the two drops have no interposition point, so the test drives the state machine and the type keeps the order. That is the reason for `LiveWriter` owning the lock rather than sitting beside it as a sibling field -- a sibling ordering is one a later edit can silently reverse.

`cargo fmt --check`, `cargo clippy --all --all-targets --all-features` under `--deny warnings`, `cargo clippy --all --all-targets` with default features, and `cargo test --all --all-features` all pass locally.

https://claude.ai/code/session_01C9BSFJKvzsERy8jjeMxwjD

---
_Generated by [Claude Code](https://claude.ai/code/session_01C9BSFJKvzsERy8jjeMxwjD)_

### DIFF
--- a/src/transaction_tracker.rs
+++ b/src/transaction_tracker.rs
@@ -85,8 +85,16 @@ impl Key for SavepointId {
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 enum WriteSlotState {
     Free,
-    Taken,
+    // The slot is taken and the writer byte is being acquired, which waits on any peer that
+    // holds it. There is no transaction id yet: it is issued from the file's latest commit,
+    // which is only known once the byte is held
+    Initializing,
+    // A live transaction guarantees that this process holds the writer byte
     Live(TransactionId),
+    // The transaction is releasing its writer-lock reference while still holding the slot.
+    // Another owner of the lock may keep the writer byte held after the transaction ends.
+    #[cfg(feature = "experimental-multiprocess")]
+    Finalizing,
 }
 
 struct State {
@@ -239,7 +247,7 @@ impl TransactionTracker {
         while state.write_slot != WriteSlotState::Free {
             state = self.live_write_transaction_available.wait(state).unwrap();
         }
-        state.write_slot = WriteSlotState::Taken;
+        state.write_slot = WriteSlotState::Initializing;
     }
 
     // Issues the slot's transaction an id that follows `last_committed`, the id of the file's
@@ -251,7 +259,7 @@ impl TransactionTracker {
         #[cfg(feature = "experimental-multiprocess")] _writer_lock: &WriterLock,
     ) -> TransactionId {
         let mut state = self.state.lock().unwrap();
-        assert_eq!(state.write_slot, WriteSlotState::Taken);
+        assert_eq!(state.write_slot, WriteSlotState::Initializing);
         state.next_transaction_id = state.next_transaction_id.max(last_committed);
         let transaction_id = state.next_transaction_id.increment();
         #[cfg(feature = "logging")]
@@ -259,6 +267,22 @@ impl TransactionTracker {
         state.write_slot = WriteSlotState::Live(transaction_id);
 
         transaction_id
+    }
+
+    // Leave `Live` before releasing the transaction's writer-lock reference
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn begin_finalizing(&self) {
+        let mut state = self.state.lock().unwrap();
+        assert!(matches!(state.write_slot, WriteSlotState::Live(_)));
+        state.write_slot = WriteSlotState::Finalizing;
+    }
+
+    #[cfg(all(test, feature = "experimental-multiprocess"))]
+    pub(crate) fn holds_writer_byte(&self) -> bool {
+        matches!(
+            self.state.lock().unwrap().write_slot,
+            WriteSlotState::Live(_)
+        )
     }
 
     // Whether a write transaction holds the write slot in this process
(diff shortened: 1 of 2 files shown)


src/transaction_tracker.rs
```diff
@@ -206,6 +206,26 @@ impl TransactionTracker {
...
+    // ends. `Live` then means this process holds the byte and `Taken` that a write transaction
+    // has the slot without it, which is what makes another one block; both are occupied, which
+    // is the separate question `write_transaction_live()` answers
+    #[cfg(feature = "experimental-multiprocess")]
+    pub(crate) fn release_writer_byte(&self) {
+        let mut state = self.state.lock().unwrap();
+        if matches!(state.write_slot, WriteSlotState::Live(_)) {
+            state.write_slot = WriteSlotState::Taken;
+        }
+    }
```
cberner: Release is a weird name. It sounds like it is releasing the lock. Also, I'm not sure that toggling back into Taken is a good idea for clarity. Do we need to add another state? Maybe Initializing, Live, Finalizing, Free

cberner: Both taken, in b4695af. Your four states, which turned out to buy more than clarity.

`WriteSlotState` is now `Free`, `Initializing`, `Live(TransactionId)`, `Finalizing`, and the method is `begin_finalizing()` -- it moves the slot into `Finalizing`, and releasing the lock is `LiveWriter`'s own field dropping afterwards, which the old name did read like.

With four states each transition has exactly one caller and can assert what it expects, so the sequence is checked rather than described:

| | asserts | leaves |
|---|---|---|
| `take_write_slot()` | waits for `Free` | `Initializing` |
| `issue_write_transaction_id()` | `Initializing` | `Live(id)` |
| `begin_finalizing()` | `Live(_)` | `Finalizing` |
| `end_write_transaction()` | not `Free`, not `Live` | `Free` |

That last row is what the toggle could not have. `end_write_transaction()` refusing a slot still reading `Live` means every write transaction in the suite checks that the byte was given up first: with `begin_finalizing()` removed from `LiveWriter::drop`, the suite fails rather than only the one test written for it. Which also let the test drop the artificial mid-drop poke it needed before -- it now just asserts the two questions during and after a transaction, and the invariant carries the ordering.

`Finalizing` is feature-gated, since without `experimental-multiprocess` there is no byte to give up and the variant would be dead; the assert has a plain not-`Free` form there.

Clippy under `--deny warnings` with all features and with default, and `cargo test --all --all-features`, are green.

---

src/transaction_tracker.rs
```diff
@@ -206,6 +215,22 @@ impl TransactionTracker {
...
+        let mut state = self.state.lock().unwrap();
+        assert!(matches!(state.write_slot, WriteSlotState::Live(_)));
+        state.write_slot = WriteSlotState::Finalizing;
+    }
+
+    #[cfg(all(test, feature = "experimental-multiprocess"))]
+    pub(crate) fn holds_writer_byte(&self) -> bool {
+        matches!(
+            self.state.lock().unwrap().write_slot,
+            WriteSlotState::Live(_)
```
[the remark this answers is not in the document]
cberner: Both gaps are real, and the change you propose would turn the first one into the bug this predicate exists to prevent.

Between `mem.lock_writer()` returning and the id being issued, this process holds the byte but is *not* yet on the file's latest commit -- `sync_to_latest_commit()` is what puts it there, and it runs in exactly that gap. A read beginning there must read the file: a peer may have committed before we took the byte, and this handle's own state is still behind it. That is the finding on #1413 (`transaction_tracker.rs:432` there, the `Taken` gate), and moving the transition to follow acquisition reopens it -- a read would skip the reload, pin this handle's stale id and follow stale roots into pages a peer's scan has already freed. The gate wants "this handle is on the file's latest commit", and holding the byte is only equivalent to that once the sync has run, which is where `Live(id)` starts. That is also why `Live` carries the id: it is issued from the file's latest commit, which is only readable once the byte is held.

The lent-lock gap is not reachable. Every caller that lends a lock across an inner transaction's end is `&mut self` -- `compact()` at `db.rs:1067` and `check_integrity()` at `db.rs:823` -- so no `begin_read()` can run on that handle while one is in progress; the remaining lender is `Database::new()`, before the handle is shared. If it were reachable it would be a no-op anyway: the lender holds the byte, so no peer commits, and the file's latest is this handle's own last inner commit.

On this PR the predicate has one consumer, and it is the test at `db.rs:3621`. The gate that reads it is #1413's, and the argument above is against that branch's code, not this one's.

---

cberner:
lgtm

PR #1466, opened 2026-09-07
Recover multi-writer handles after interrupted compaction

cberner-ai:
If a process exits during compaction or repair before recording its final allocator snapshot, a surviving `MultiWriterProcess` handle currently returns `Corrupted` from `begin_write()` and requires reopening. This change lets that handle verify the committed trees, rebuild the allocator, and resume writing.

Recovery runs under the local write slot and writer byte when no valid allocator snapshot is available. It retains pages referenced by the freed-page tables, preserves reader and savepoint snapshots, and rejects damaged roots without leaving a usable partial allocator. The next normal commit records a fresh snapshot. The changelog describes the recovery behavior.

Regression coverage includes a subprocess exiting after compaction drains pending frees, a surviving writer committing while a peer retains its read snapshot, savepoint restoration after an aborted recovery transaction, and rejection of invalid checksums and table counts. Both successful-recovery regressions failed before the fix.

Validation:

- `just test` (formatting, Clippy, dependency checks, and tests with all features).
- Original process-crash reproduction: the surviving writer now starts a transaction successfully.
- Existing fuzz harness: bounded `just fuzz` run using `nightly-2025-07-26`, `-max_len=10000`, and no sanitizer; approximately 46,000 executions without a failure, stopped at the time limit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@
   and a write transaction begins from the file as another process last committed it, as do the
   close and `Database::check_integrity()`, which waits for a write transaction in another process
   to end.
+* In `MultiWriterProcess`, existing handles recover automatically after another process exits
+  during compaction or repair. They can resume writing without reopening the database.
 
 ### redb-derive (unreleased)
 * Fix `#[derive(Value)]` and `#[derive(Key)]` failing to compile on structs whose lifetimes are
--- a/src/db.rs
+++ b/src/db.rs
@@ -896,8 +896,8 @@ impl Database {
         } else {
             Some(own_allocator_hash)
         };
-        // Every multi-writer commit records the allocator state, so one that did not is a commit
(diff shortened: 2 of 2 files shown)


docs/design.md
```diff
@@ -518,6 +518,10 @@ also take an exclusive lock on the "writer byte" during a write transaction.
 In addition to the cache invalidation and synchronization protocols, all write transactions must use
 quick-repair, and ephemeral savepoints are not supported. At the beginning of a write transaction
 the allocator state must be reloaded from the latest quick-repair snapshot in the database file.
+If compaction or repair is interrupted before recording its final snapshot, the next writer verifies
+the committed trees and rebuilds the allocator state, including pages retained by the freed-page
+tables. Its next commit records a new snapshot. This recovery holds the writer byte and the local
+write slot, so already-open handles can continue without reopening the file.
```
cberner: Remove

PR #1467, opened 2026-09-07
Fix file locking on 32-bit GNU/Linux

cberner-ai:
Opening even an empty database can panic on 32-bit GNU/Linux because the locking protocol uses offsets near `2^62`, while `libc::off_t` is only 32 bits on targets such as i686. Both lock offsets and lengths can fail the checked conversion.

Use `flock64` and the appropriate large-file `fcntl` syscall for GNU/Linux acquisitions, queries, and unlocks, including blocking locks. Calling through `libc::syscall` avoids glibc's narrow `flock` conversion without requiring the `fcntl64` symbol introduced in glibc 2.28.

Add regressions for high offsets remaining distinct from low offsets, exact boundaries of large ranges, and blocking locks. CI now runs the locking tests with default and all features, plus the multi-process integration suite, on i686.

Validation:

- Before the fix, all 12 existing default-feature locking tests panicked on `i686-unknown-linux-gnu` with `TryFromIntError`.
- `just test` passed, including formatting, Clippy, dependency checks, and all-feature tests.
- i686 GNU/Linux: 14 default-feature locking tests, 12 all-feature locking tests, and all 19 multi-process integration tests passed. Clippy passed for all targets and features.
- i686 musl: all 12 all-feature locking tests passed.
- All-feature builds checked successfully for `armv7-unknown-linux-gnueabihf` and `x86_64-unknown-linux-gnux32`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,6 +163,16 @@ jobs:
       - name: Clippy (default features)
         run: cargo clippy --all --all-targets -- -Dwarnings
 
+      - name: Run 32-bit locking tests
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends --yes gcc-multilib
+          rustup target add i686-unknown-linux-gnu
+          cargo test -p $(cargo pkgid) --target i686-unknown-linux-gnu --lib range_lock_tests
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -63,6 +63,7 @@
   all built-in types are unaffected.
 
 ## 4.3.0 - 2026-XX-XX
+* Fix opening a database panicking on 32-bit GNU/Linux targets when acquiring or querying file locks.
```
cberner: Remove. This regression was never released it is just on master

PR #1468, opened 2026-09-08
Avoid deadlock after a failed multi-writer open

cberner-ai:
A malformed persistent savepoint can make a multi-writer open fail while it still holds the opening writer lock. The partially constructed `Database` then runs its normal close, which tries to take that lock again and hangs on Windows. On Linux, the same cleanup can publish shutdown commits with an incomplete savepoint tracker.

Construct `Database` after savepoint tracking and the final repair allocator commit succeed. Savepoint loading now operates directly on the memory and tracker, allowing failed initialization to release its resources through the existing backend cleanup.

The regression creates a valid allocator snapshot with a malformed savepoint, then checks that opening returns the original corruption error within a timeout, releases every lock while a duplicate file handle remains alive, and leaves the committed transaction ID unchanged. It runs on Windows, Linux, and macOS under `experimental-multiprocess`.

Validation:

- The regression failed before the fix on Linux: cleanup advanced the committed transaction ID from 7 to 10. It passes with the fix.
- `just test` passed, including formatting, Clippy, dependency checks, and all-feature tests.
- All 132 default-feature unit tests passed.
- A bounded `just fuzz` run completed over 60,000 executions without a failure before being stopped at its two-minute limit.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -945,7 +945,9 @@ impl Database {
 
         // The tracker holds the persistent savepoints the reloaded tables hold, as the open's does
         if peer_committed {
-            self.sync_persistent_savepoints(
+            Self::sync_persistent_savepoints(
+                &self.transaction_tracker,
+                &self.mem,
                 #[cfg(feature = "experimental-multiprocess")]
                 Some(writer_lock),
             )?;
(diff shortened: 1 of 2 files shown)


cberner:
lgtm

PR #1470, opened 2026-09-08
Distinguish invalidated allocator state from missing state

cberner:
The absence of allocator state currently covers both a snapshot waiting to be loaded and state invalidated by an interrupted operation. Write initialization and integrity checks need to distinguish these cases: invalidated state requires reopening, while a header reload leaves its caller responsible for loading the new allocator snapshot.

Replace `Option<Allocators>` with `AllocatorState::{None, Loaded(Allocators), Invalidated}`. Header reloads leave `None`; failed operations mark the state `Invalidated`; loading or rebuilding installs `Loaded`. Write initialization and integrity checks now reject `Invalidated` specifically. Existing reload paths already load the snapshot before reaching these checks, so this prepares for #1413 without changing current behavior.

Allocator access diagnostics distinguish missing state from invalidated state. The state-transition test covers initialization, loading, invalidation, and rebuilding; the existing mutex-poisoning test separately checks cleanup during unwinding.

Validation: `just test` passes, including the dependency audit, formatting, Clippy, and all-feature tests.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -825,7 +825,7 @@ impl Database {
         self.mem.check_io_errors()?;
         // Once the allocator state has been discarded (by a failed commit or integrity check),
         // the database must be reopened to rebuild it; this check requires one to compare against
-        if !self.mem.allocator_state_loaded() {
+        if self.mem.allocator_state_invalidated() {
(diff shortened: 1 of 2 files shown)


cberner:
lgtm

PR #1471, opened 2026-09-08
Capture read snapshots under the header lock

cberner:
Read registration pins a transaction ID before `ReadTransaction::new()` reads the data root. `ReadOnlyDatabase` uses a separate `begin_read` mutex to prevent another read from reloading the header between those steps.

Capture the ID and root together under the memory-state mutex, with the header lock held through capture and registration. Return the root with its transaction guard and use it to construct reads in both `Database` and `ReadOnlyDatabase`. This removes the extra `ReadOnlyDatabase` mutex and releases the header lock before constructing the table tree. Savepoint registration continues to use the same guard allocation path.

Regression tests cover commits between registration and construction, including non-durable commits, and a read-only header reload before an earlier read is constructed in both multiprocess modes.

Validation: `just test`; the local snapshot regression with default features; `cargo check --no-default-features`; a bounded `just fuzz` smoke run without a sanitizer.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -450,13 +450,14 @@ impl TransactionGuard {
         }
     }
 
+    // Returns a read guard and the data root its registration protects.
     pub(crate) fn allocate_read(
         tracker: Arc<TransactionTracker>,
         mem: &Arc<TransactionalMemory>,
-    ) -> Result<Self> {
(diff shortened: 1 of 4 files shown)


cberner:
lgtm

PR #1472, opened 2026-09-09
Read shared read-only snapshots directly from the file

cberner:
Shared read-only transactions currently refresh their snapshots by replacing `TransactionalMemory`'s cached header. Read the validated primary ID/root directly from the file instead, leaving the in-memory header and allocators untouched.

`PagedCachedFile` stores the read cache's last observed transaction ID and handles comparison and invalidation in `update_transaction_id()`. `TransactionalMemory` calls it after reading the file snapshot, while still holding the header lock. Observing a different commit invalidates cached pages; repeated reads of the same commit retain them. `HeaderGuard` remains a plain lock guard. This changes only the existing shared read-only path; writable handles continue to use their in-memory snapshots.

Tests cover both multiprocess modes, including cache reuse, invalidation after a peer commit, unchanged in-memory ID/root, and retention of older snapshots.

Validation: `just test`, `cargo check --no-default-features`, and a bounded `just fuzz` smoke run without a sanitizer.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -3548,7 +3548,7 @@ mod active_transaction_test {
     }
 
     #[test]
-    fn read_only_snapshot_survives_reload_before_construction() {
+    fn read_only_snapshot_survives_new_read_before_construction() {
         use super::{ReadTransaction, TransactionGuard};
         use crate::ReadableTable;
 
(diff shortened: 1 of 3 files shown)


cberner:
lgtm

PR #1474, opened 2026-09-09
Report a failed release of the active transaction byte

cberner:
The byte a read transaction holds is released as its last reference goes away. A failure was dropped on the floor:

```rust
// Failing to release only leaves a peer reclaiming less than it could
if self.active_transaction_lock_references(id) == 0 {
    let _ = mem.unlock_mp_transaction(id);
}
```

That comment understates it. The byte stays held for the life of the handle, and it is how every other process learns how far back this one is still reading. So a failed release stops the **whole cohort** from reclaiming past that transaction, for as long as the handle is open -- not one peer, and not briefly.

The callers are drop paths (`TransactionGuard::drop`, savepoint teardown), with nothing to return an error to and no way to undo the failure. So it stays best-effort, but it is now reported instead of discarded:

> Failed to release a finished read transaction: {failure}. Until this database is closed, no process sharing it can reclaim space from data overwritten or deleted after this point, so the file may grow

`persistent_savepoint()` releases the byte on the same condition, so both releases go through one `release_active_transaction_lock()` rather than the message being mirrored in two places.

## Notes

- The `error` import is gated on `all(logging, experimental-multiprocess)`, and the binding is consumed by a `let _` when `logging` is off, so neither becomes an unused-item warning under `RUSTFLAGS=--deny warnings`.
- The binding is named `failure` rather than `error`, so it does not read like the `error!` macro beside it, and so it is not an underscore-prefixed binding used in a format string (`clippy::used_underscore_binding`, denied via `clippy::pedantic` in `lib.rs`).
- No `CHANGELOG.md` entry: no behaviour change, only a log line on a path that should never be taken.

## Testing

- `cargo clippy --tests` under `--all-features` and under `experimental-multiprocess` alone, since the `logging`-off build is the one that needs the `let _`
- `cargo test --all --all-features` -- 25 test binaries, 0 failures
- `cargo check` under default, `experimental-api-5`, `experimental-multiprocess`, `logging`, and `logging,experimental-multiprocess`, since the change is gated on two features independently

### DIFF
--- a/src/transaction_tracker.rs
+++ b/src/transaction_tracker.rs
@@ -14,6 +14,8 @@ use core::mem;
 use core::mem::size_of;
 #[cfg(feature = "logging")]
 use log::debug;
+#[cfg(all(feature = "logging", feature = "experimental-multiprocess"))]
+use log::error;
 
 #[derive(Copy, Clone, Hash, Ord, PartialOrd, Eq, PartialEq, Debug)]
 pub(crate) struct TransactionId(u64);
@@ -166,13 +168,26 @@ impl State {
         id: TransactionId,
     ) {
         self.decrement_reference_count(id);
-        // Failing to release only leaves a peer reclaiming less than it could
         #[cfg(feature = "experimental-multiprocess")]
         if self.active_transaction_lock_references(id) == 0 {
-            let _ = mem.unlock_mp_transaction(id);
+            Self::release_active_transaction_lock(mem, id);
         }
         #[cfg(not(feature = "experimental-multiprocess"))]
         let _ = mem;
+    }
+
+    #[cfg(feature = "experimental-multiprocess")]
+    fn release_active_transaction_lock(mem: &TransactionalMemory, id: TransactionId) {
+        if let Err(failure) = mem.unlock_mp_transaction(id) {
+            #[cfg(feature = "logging")]
+            error!(
+                "Failed to release a finished read transaction: {failure}. Until this database is \
+                 closed, no process sharing it can reclaim space from data overwritten or deleted \
+                 after this point, so the file may grow"
+            );
+            #[cfg(not(feature = "logging"))]
+            let _ = failure;
+        }
     }
 
     // A persistent savepoint's reference, which takes no byte. See `persistent_savepoint_references`
@@ -200,7 +215,7 @@ impl State {
         *self.persistent_savepoint_references.entry(id).or_insert(0) += 1;
         #[cfg(feature = "experimental-multiprocess")]
         if self.active_transaction_lock_references(id) == 0 {
-            let _ = mem.unlock_mp_transaction(id);
+            Self::release_active_transaction_lock(mem, id);
         }
         #[cfg(not(feature = "experimental-multiprocess"))]
         let _ = mem;


src/transaction_tracker.rs
```diff
@@ -166,10 +168,16 @@ impl State {
...
+        // A byte left held pins `id` for the life of this handle, bounding what every process in
+        // the cohort may reclaim. Nothing here can undo that, so it is reported rather than dropped
         #[cfg(feature = "experimental-multiprocess")]
-        if self.active_transaction_lock_references(id) == 0 {
-            let _ = mem.unlock_mp_transaction(id);
+        if self.active_transaction_lock_references(id) == 0
+            && let Err(error) = mem.unlock_mp_transaction(id)
+        {
+            #[cfg(feature = "logging")]
+            warn!("Failed to release the active transaction byte for {id:?}: {error}");
```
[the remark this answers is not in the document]
cberner: Confirmed and fixed. `convert_reference_to_persistent_savepoint()` releases the byte on the same condition and discarded the error the same way, so warning on only one of the two would have left the savepoint path silent.

Rather than mirror the warning, both releases now go through a single `release_active_transaction_lock()`, so there is one place the byte is released and one place a failure can be dropped.

Pushed in b237917.

---

src/transaction_tracker.rs
```diff
@@ -166,15 +168,27 @@ impl State {
...
     }
 
+    // Releases the byte as the last read of `id` goes away. A byte left held pins `id` for the
+    // life of this handle, bounding what every process in the cohort may reclaim, and nothing
+    // here can undo that, so the failure is reported rather than dropped
+    #[cfg(feature = "experimental-multiprocess")]
+    fn release_active_transaction_lock(mem: &TransactionalMemory, id: TransactionId) {
+        if let Err(error) = mem.unlock_mp_transaction(id) {
+            #[cfg(feature = "logging")]
+            warn!("Failed to release the active transaction byte for {id:?}: {error}");
```
cberner: This should be an error, and the message should be more user friendly. It should tell them the consequence, not internal details about the protocol bytes

cberner: Done, in 2bc8ca4. Now `error!`, and the message says what the user sees rather than how the protocol works:

> Failed to release a finished read transaction: {failure}. Until this database is closed, no process sharing it can reclaim space from data overwritten or deleted after this point, so the file may grow

No transaction bytes, ids, or lock ranges. The underlying I/O error stays, since that is the part a user can act on.

The binding is named `failure` rather than `error` so it does not read like the `error!` macro beside it.

---

src/transaction_tracker.rs
```diff
@@ -166,15 +168,31 @@ impl State {
...
-            let _ = mem.unlock_mp_transaction(id);
+            Self::release_active_transaction_lock(mem, id);
         }
         #[cfg(not(feature = "experimental-multiprocess"))]
         let _ = mem;
     }
 
+    // Releases the byte as the last read of `id` goes away. A byte left held pins `id` for the
+    // life of this handle, bounding what every process in the cohort may reclaim, and nothing
+    // here can undo that, so the failure is reported rather than dropped
```
cberner: remove

src/transaction_tracker.rs
```diff
@@ -166,15 +168,31 @@ impl State {
...
+    #[cfg(feature = "experimental-multiprocess")]
+    fn release_active_transaction_lock(mem: &TransactionalMemory, id: TransactionId) {
+        if let Err(failure) = mem.unlock_mp_transaction(id) {
+            #[cfg(feature = "logging")]
+            error!(
+                "Failed to release a finished read transaction: {failure}. Until this database is \
+                 closed, no process sharing it can reclaim space from data overwritten or deleted \
+                 after this point, so the file may grow"
```
cberner, line 186: "is closed" implies that dropping Database will close it. But really the process needs to exit, right?

cberner: I traced it, and I think dropping the `Database` is enough -- the process does not need to exit.

`close_database()` always calls `mem.close()`, which reaches `FileBackend::close()` -> `unlock_range(FULL_RANGE)` -> `release_all_locks()`, and that drains every range in `locked_ranges` and unlocks each one. The byte this message is about is still in that set, because a failed unlock deliberately keeps it there:

```rust
// Forgotten only once it is really released, so that a failure leaves the range for
// close() to retry rather than leaving it held with nothing left to release it
```

So closing the database retries exactly this release. If the retry also fails, the fd closing when the last `Arc<PagedCachedFile>` goes releases it regardless, since these are open-file-description locks.

That makes "closed" both accurate and the earlier of the two bounds, which is why I would rather not tell users to restart a process when dropping the handle is enough. Say the word if you read the path differently and I will reword -- you know this code better than I do, and I would rather the message be wrong-but-conservative than wrong-but-reassuring.

---

PR #1477, opened 2026-09-09
Support locking in custom backends

cberner-ai:
Passing a `FileBackend` to `Builder::create_with_backend()` discarded its locking behavior. A second writer could open a database already in use, while multi-process writer modes failed with `Unsupported`. All builder paths now use the supplied backend's locking methods.

Expose six object-safe locking methods and `close()` on `StorageBackend`, replacing the private backend trait and lockless wrapper. Lock methods accept explicit `Bound<u64>` endpoints and return `BackendError::Io` or `BackendError::Unsupported`. Defaults preserve existing implementations, including lockless single-process opens in `no_std` builds. The API documents ownership, non-overlapping acquisitions, exact-range unlocks, query behavior, and cleanup at close.

The core selects the concurrency protocol's ranges for every backend. If a partial range is unsupported, it releases any acquired ranges before trying `Unbounded, Unbounded`. Immutable opens query the shared writer byte only when their own locks leave it free; a whole-storage fallback needs no query. Whole-storage-only backends need the two nonblocking acquisition methods, unlock, and close.

`FileBackend` locks the requested ranges literally and tracks acquisitions for cleanup, including when a caller retains a cloned file handle. A small Linux compatibility shim also attempts the legacy whole-file lock for single-process opens in 4.x; this is best effort and disabled under `experimental-api-5`.

Validation:

- `just test`: dependency checks, formatting, Clippy, and the full all-feature test suite passed.
- Default-feature locking regressions and Clippy passed, including exclusion of legacy writers after one of several immutable readers closes.
- i686 locking and public range tests passed.
- The standalone `no_std` runtime regression, a whole-file-only backend runtime probe with a retained file handle, and multiprocess Rustdoc with warnings denied passed.
- A bounded fuzz run completed 29,192 executions in 61 seconds without a failure.


src/tree_store/page_store/file_backend/range_lock.rs
```diff
@@ -232,34 +270,40 @@ mod windows_imp {
...
-            let len = range.end - range.start;
+        fn unlock_range(&self, range: impl RangeBounds<u64>) -> io::Result<()> {
+            let (start, end) = checked_range(range)?;
+            let len = end
+                .unwrap_or(u64::MAX)
+                .checked_sub(start)
+                .filter(|len| *len > 0)
+                .ok_or_else(|| {
+                    io::Error::new(io::ErrorKind::InvalidInput, "unrepresentable lock range")
+                })?;
```
cberner: Should this be part of checked_range()?

src/tree_store/page_store/page_manager.rs
```diff
@@ -710,35 +714,55 @@ impl TransactionalMemory {
...
+            (Bound::Excluded(SHARED_WRITER_BYTE), Bound::Unbounded),
+            (
+                Bound::Unbounded,
+                if read_only {
+                    Bound::Excluded(SHARED_WRITER_BYTE)
+                } else {
+                    Bound::Included(SHARED_WRITER_BYTE)
+                },
+            ),
+        ];
```
cberner: This nesting is very hard to read. Add some intermediate variables

PR #1480, opened 2026-09-11
Keep aborted savepoint handles invalid across writer processes

cberner-ai:
A `Savepoint` handle can outlive the aborted transaction that created its persistent savepoint. In `MultiWriterProcess`, a peer can reuse that persistent ID, making the stale handle valid again and allowing it to restore an obsolete snapshot that loses committed data.

Assign each savepoint registration a monotonically increasing local ID and validate it on restore. Unchanged registrations retain their identity across peer commits; replacements receive a fresh identity and the correct snapshot reference. Transaction drop retires unfinished registrations, including after a panic or failed commit. Public savepoint IDs and the file format are unchanged.

Regressions cover explicit abort, ordinary drop, and panic unwinding, with replacements at both the same and a newer snapshot. They also verify that valid handles survive repeated lookups and peer commits, and that deletion releases their references. The API and design docs describe the behavior.

Validation passed:

- `just test` (including all 25 multiprocess integration tests).
- 60-second transaction/savepoint fuzz smoke test: 27,091 runs without a failure.
- `no_std` compile with warnings denied.
- Both original audit reproducers now reject stale handles with `InvalidSavepoint`.

### DIFF
--- a/src/transaction_tracker.rs
+++ b/src/transaction_tracker.rs
@@ -45,6 +45,15 @@ impl TransactionId {
 #[derive(Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Debug)]
 pub(crate) struct SavepointId(pub u64);
 
+// Unique within one tracker, including across aborted transactions and peer ID reuse.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub(crate) struct LocalSavepointId(u64);
+
+struct SavepointRegistration {
+    local_id: LocalSavepointId,
+    transaction_id: TransactionId,
+}
+
 impl SavepointId {
     pub(crate) fn next(self) -> SavepointId {
         SavepointId(self.0 + 1)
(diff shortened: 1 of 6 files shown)


src/transactions.rs
```diff
@@ -1147,6 +1149,9 @@ impl WriteTransaction {
     /// This savepoint will exist until it is deleted with
     /// [`delete_persistent_savepoint()`](Self::delete_persistent_savepoint).
     ///
+    /// If this transaction is aborted, the savepoint and any handles obtained for it become
+    /// invalid, even if another writer later reuses its id.
```
cberner: Remove the "even if another writer..." part.

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@
   to end.
 * In `MultiWriterProcess`, existing handles recover automatically after another process exits
   during compaction or repair. They can resume writing without reopening the database.
+* Fix aborted persistent savepoint handles becoming valid again when another writer reuses
+  their ids in `MultiWriterProcess`. Restoring such a handle could lose committed data.
```
cberner: remove. This regression was never released

src/transaction_tracker.rs
```diff
@@ -127,6 +137,27 @@ struct State {
...
+            self.valid_savepoints
+                .insert(
+                    id,
+                    SavepointRegistration {
+                        local_id,
+                        transaction_id,
+                    },
+                )
+                .is_none()
+        );
```
cberner: Add an intermediate variable. The formatting here is really bad

src/transactions.rs
cberner: Introduce some intermediates if possible. This if condition is hard to read now

PR #1490, opened 2026-09-17
Add Value::NICHE and implement Value and Key for the NonZero integer types

### LINKED ISSUE #873: Implement `Key` for `NonZero…`

cberner:
Fixes #873
Fixes #1408
Fixes #1409.

`Value` gains an optional `NICHE`: a byte string that `as_bytes()` never returns. `Option<T>` encodes `None` as `T`'s niche in place of its tag byte, so `Option<T>` costs no more space than `T`. The `NonZero` integer types implement `Value` and `Key`, encoded as the primitives they wrap, with zero as their niche: `Option<NonZeroU32>` is 4 bytes rather than 5. No specialization is needed, since the varying behavior lives on redb's own trait.

Related: https://github.com/cberner/redb/pull/1408 and https://github.com/cberner/redb/pull/1409 explore the same approach.

## Design

- `const NICHE: Option<&'static [u8]> = None` on `Value`. A fixed width type's niche must be exactly its width; a variable width type's may be any length, including empty. Types that declare none keep the tag encoding byte for byte, so existing tables are unaffected.
- `Option<T>` with a niche: `None` is the niche, `Some(x)` is `x`'s own encoding. `None` still sorts below every `Some`, negative signed values included, and is still `min_encoded_key()`; two `Some` values separate as `T`'s encodings do.
- `(T,)` forwards `T`'s niche, since it is encoded exactly as `T`.
- `#[derive(Value)]` on a struct with exactly one field forwards the field's niche too, since it is encoded exactly as `(T,)`. The generated code refers to `Value::NICHE`, so such a struct can only be derived against a redb that has it: deriving it against redb 3.0 through `#[redb(crate = "...")]` no longer compiles, and the compatibility test's one-field inner struct gained a second field to keep covering 3.0.
- `NonZeroU8` to `NonZeroU128` and `NonZeroI8` to `NonZeroI128` delegate to the wrapped primitive's `Value` and `Key` impls. Each has its own type name, so a table of the primitive, which may hold zeros, does not open as the `NonZero` type. `NonZeroUsize` and `NonZeroIsize` are left out, as `usize` and `isize` are.
- Debug assertions in `Option<T>` check the niche rules: a fixed width niche is that wide, no value encodes to the niche, and `T::separator()` never returns it.

## Not done here

`Option<T>`'s type name does not change when `T` gains a niche. That is fine for the new `NonZero` types, but a user adding `NICHE` to a type that already has `Option<T>` tables would have that data misread rather than rejected. The changelog warns about it. A type-name change for niche-encoded `Option` is probably part of the migration story for `Option<&str>` and `Option<bool>`, which this PR deliberately leaves alone.

## Commits

One per step, each building and passing its tests on its own:

1. Add `Value::NICHE`, with the changelog subheadings
2. Forward the element's niche through `(T,)`
3. Encode `Option<T>` without a tag when `T` declares a niche
4. Implement `Value` and `Key` for the `NonZero` integer types
5. Forward the field's niche through single-field derived structs

## Tests

- Unit tests: a fixed width test type whose niche is the bytewise largest encoding, a variable width test type tried with empty, 1-byte and 3-byte niches (including a `separator()` that would naturally produce the niche), all ten `NonZero` types with ordering checked against Rust's own `Ord` for `Option<NonZero*>`, nested `Option<Option<_>>`, tuple forwarding, and a niche-encoded `None` substituted as a collapsed tuple tail.
- Integration tests: `NonZero` and `Option<NonZero>` tables, a 2049-entry `Option<NonZeroI32>` table asserting 12 stored bytes per entry with a multi-level tree, reopened from disk and integrity-checked, and type identity in both key and value positions.
- Derive tests: one-field tuple and named structs forward the niche, nesting forwards it again, multi-field and unit structs declare none, `Option` of a one-field struct round-trips through a table with `None` first, and the shadowed-name and no-prelude modules gained one-field structs so the generated constant compiles there too.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets --all-features` with `--deny warnings`, `cargo test --all-features` (which covers the derive crates), `cargo doc`, `cargo deny check licenses advisories`, the `thumbv7em-none-eabihf` no_std check, the `wasm32-unknown-unknown` check of redb and redb-derive, and the new tests under `wasm32-wasip1-threads` with the CI-pinned nightly all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MzoHC3wdiF7ZEBM9XucstN)_


src/types.rs
```diff
@@ -160,6 +160,21 @@ pub trait Value: Debug {
...
+    /// every encoding, so `Option<Self>` costs no more space than `Self`. Declaring one is
+    /// optional, and the default is `None`.
+    ///
+    /// A fixed width type's niche must be exactly [`fixed_width()`](Self::fixed_width) bytes
+    /// long. A variable width type's niche may be any length, including empty.
+    ///
+    /// Implementations must ensure that `as_bytes()` never returns the niche, for any value:
+    /// `Option<Self>` would decode that value as `None`. Adding a niche to a type that already
+    /// has tables, or changing it, changes the encoding of `Option<Self>` without changing its
+    /// `type_name()`, so existing `Option<Self>` data would be misread rather than rejected.
```
cberner: Don't include so much discussion of the Option type. You can give Option as a short example of a container type that may use the NICHE for an optimization, but this documentation should focus only on the requirements of NICHE: that as_bytes() never returns it

cberner (commented):
Also add another commit which adds NICHE forwarding for single field structs since they share the same encoding as (T,)

cberner:
Both review items are in the pushed series (head `3a8fccf`):

- The `NICHE` doc now states only the requirement: `as_bytes()` never returns it, plus the width rule for fixed width types, with `Option` as a one-line example. Amended into the first commit, since this repo does not merge fixup commits.
- New fifth commit: `#[derive(Value)]` on a struct with exactly one field forwards the field's `NICHE`, since it is encoded exactly as `(T,)`. One trade-off to be aware of: the generated impl refers to `Value::NICHE`, so a one-field struct can no longer be derived against redb 3.0 through `#[redb(crate = "...")]`. The derive docs and changelog say so, and the compatibility test's one-field `Old3Inner` gained a second field to keep covering 3.0. If you would rather keep 3.0 for one-field structs, an opt-out on the `#[redb(...)]` attribute would do it; say the word and I will add it.

---

cberner (commented):
Remove the references to version "4.4" in the redb-derive comments

cberner:
Done in `44647ae`: the redb-derive comments, the compatibility test's module doc, and the derive changelog bullet no longer name a version; they say a one-field struct needs a redb that has `Value::NICHE`. Amended into the fifth commit.

---

cberner:
On the Codex P1 ("give niche-encoded options a distinct type identity"): this is the item listed under "Not done here" above, and the `Value::NICHE` changelog entry warns about it. It is deliberately left for the migration work planned for `Option<&str>` and `Option<bool>`, where the spelling of the new type identity gets decided once for all of them.

Two precisions on the exposure. It cannot affect the types this PR adds: `Option<NonZero*>` has never been stored with a tag. And it cannot affect a fixed width type that adds a niche later: the stored table definition records the fixed width, `open_table()` compares it, and `Option<T>` is one byte narrower with the niche, so such a table fails with `TypeDefinitionChanged` rather than being misread. What remains is a user's variable width type that starts declaring `NICHE` while it already has `Option<T>` tables, which the docs and changelog call out.

Not changing the type identity in this PR unless @cberner wants it here.

---

PR #1491, opened 2026-09-20
Name Option of a niche type after the niche

cberner:
`Option<T>` now has a type name of its own when `T` declares a `Value::NICHE`, so that the type name follows the encoding: `Option` of a type with a niche and `Option` of the same type without one are not equivalent types. A table that stores `Option` of a type in the tag byte encoding, written before the type declared a niche, keeps its own type name and fails to open as the new type with `TableTypeMismatch`, instead of being misread. That makes adding a niche to a type safe by construction, which the 4.4.0 changelog entry previously warned against; such a table can be read through a copy of the type that declares no niche.

`NICHE` is unreleased, so no stored table changes its name: `Option<NonZero*>` and `Option` of a user type with a niche get the new name from the start. The niche forwarding through `(T,)` and single-field derived structs, and the composites that embed the inner name (`Vec`, arrays, tuples, derived structs), pick it up automatically, which the new unit test pins down.

This is the prerequisite for giving `&str` and `String` a niche behind a feature flag, where existing `Option<&str>` tables have to keep their identity; that change follows separately.


src/types.rs
```diff
@@ -418,8 +422,14 @@ impl<T: Value> Value for Option<T> {
...
     fn type_name() -> TypeName {
         let inner = T::type_name();
-        TypeName::internal(&format!("Option<{}>", inner.name()))
-            .into_composite(inner.is_user_defined())
+        // The niche changes the encoding, so the name records it: a table written before the
+        // type declared one keeps its own name, and is not misread
+        let name = if T::NICHE.is_some() {
+            format!("niche::Option<{}>", inner.name())
+        } else {
+            format!("Option<{}>", inner.name())
```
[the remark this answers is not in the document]
cberner: No, I don't think that's necessary

src/types.rs
```diff
@@ -172,7 +172,11 @@ pub trait Value: Debug {
     /// no niche.
     ///
     /// Container types may use it as an encoding that no value of `Self` occupies: `Option<Self>`,
-    /// for example, encodes `None` as it, in place of a tag byte.
+    /// for example, encodes `None` as it, in place of a tag byte. Declaring a niche on a type
+    /// changes the encoding of `Option` of it, and its type name with it, so the `Option` with
+    /// the niche and the one without are not equivalent types: a table that stores `Option` of
+    /// the type from before it declared the niche fails to open with a type mismatch, instead of
+    /// being misread, and can be read through a copy of the type that declares no niche.
```
cberner: Too verbose

src/types.rs
```diff
@@ -418,8 +422,14 @@ impl<T: Value> Value for Option<T> {
 
     fn type_name() -> TypeName {
         let inner = T::type_name();
-        TypeName::internal(&format!("Option<{}>", inner.name()))
-            .into_composite(inner.is_user_defined())
+        // The niche changes the encoding, so the name records it: a table written before the
+        // type declared one keeps its own name, and is not misread
+        let name = if T::NICHE.is_some() {
```
cberner, line 422: remove

cberner: Removed the code comment above this line in fe3d82b, and shortened the `NICHE` doc to one sentence. If you meant the `let name` binding itself, say so and I will inline it.

---

PR #1492, opened 2026-09-21
Add the Legacy type and give &str a niche, behind the experimental-niches feature flag

cberner:
Two commits, behind the new `experimental-niches` feature flag, which requires `experimental-api-5`:

**Add `Legacy<T>`, a type without its niche.** A blanket impl that delegates everything to `T` and leaves `NICHE` at its default. Declaring a niche changes the encoding of `Option<T>` and its type name with it, so a table of `Option<T>` written before the niche no longer opens as `Option<T>`: it opens as `Option<Legacy<T>>`, however deeply nested (`Vec<Option<Legacy<T>>>`, `Option<(Legacy<T>,)>`), and yields plain `Option<T>`, so a table definition migrates by changing the type alone. This works for any type that declares a niche, user-defined ones included. A struct deriving `Value` cannot have `Legacy<T>` as a field, since the derive reads each field back as its own type; redb-derive is not stable, so that is left for later.

**Give `&str` a niche.** `&str` declares `0xff` as its `NICHE`, so `Option<&str>` is stored as the string itself, with the niche standing for `None`, in place of a tag byte. `Option<&str>` with the flag is therefore not the same type as `Option<&str>` without it: a table written without the flag fails to open with `TableTypeMismatch` instead of being misread, and opens as `Option<Legacy<&str>>`; the `Legacy` doc has a doctest showing the copy-across migration. The file format is unchanged: no version bump and no catalog rewrite.

Tests: a unit test on `Legacy<NonZeroU32>` pins that `Option<Legacy<T>>` has the tag byte encoding and the un-niched name. `tests/niche_tests.rs` writes a redb 2.6 fixture (file format v3) holding `Option<&str>` as values and as keys, checks that the new definitions are rejected with `TableTypeMismatch`, and reads every entry back through the legacy definitions. Unit tests in `src/types.rs` pin the `&str` niche encoding and the type names.

`String` will follow in a separate PR.


src/legacy_types.rs
```diff
@@ -0,0 +1,99 @@
...
+///
+/// On its own, `LegacyStr` is `&str`: a table of it is a table of `&str`.
+#[derive(Debug)]
+pub struct LegacyStr;
+
+impl Value for LegacyStr {
+    type SelfType<'a>
+        = &'a str
+    where
+        Self: 'a;
```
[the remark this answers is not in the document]
cberner: Confirmed, and fixed in feda1b6: `LegacyStr` is now a newtype over `&str` (`LegacyStr<'a>(pub &'a str)`, derefs to `str`), with `SelfType` being itself, so it works as a field of a derived struct. `Option<LegacyStr>` keeps the old name and encoding, so a struct of the same name deriving `Value` with the field as `Option<LegacyStr>` opens the old table. Added a test against a redb 2.6 fixture of such a table, read through the derived legacy struct.

---

src/legacy_types.rs
```diff
@@ -0,0 +1,110 @@
...
+/// `&str` as it was before it declared a niche: reads a table of `Option<&str>` written without
+/// the `experimental-niches` feature.
+///
+/// Such a table stores `Option<&str>` with a tag byte, under the type name `Option<&str>` had
+/// then. With the feature, `Option<&str>` is encoded with the niche and has another type name to
+/// match, so the table no longer opens as `Option<&str>`. `Option<LegacyStr>` has the old
+/// encoding and the old name, however deeply it is nested: `(u64, Option<LegacyStr>)` opens a
+/// table of `(u64, Option<&str>)`, and a struct deriving `Value` with a field of
+/// `Option<LegacyStr>` opens a table of the struct of that name whose field is `Option<&str>`.
+/// The table is migrated by copying it into one of `Option<&str>`:
```
cberner: Refer to it as before version 5.0, since this will be stabilized in 5.0, rather than "before it declared a niche". Also I like that there's an example, but tighten up the rest of this commentary

src/legacy_types.rs
```diff
@@ -0,0 +1,110 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct LegacyStr<'a>(pub &'a str);
+
+impl Deref for LegacyStr<'_> {
+    type Target = str;
+
+    fn deref(&self) -> &str {
+        self.0
+    }
+}
```
cberner: Hmm, do we need this?

src/legacy_types.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl Value for LegacyStr<'_> {
+    type SelfType<'a>
+        = LegacyStr<'a>
```
cberner: Can't we declare this `&str`? I know it goes against the docs of `Value`, but I think that's what the previous Legacy wrapper did. No?

src/legacy_types.rs
cberner: Reworded as before 5.0, and cut to two short paragraphs around the example.

---

src/legacy_types.rs
cberner: No. Removed; the example uses `.map(|s| s.0)` instead.

---

src/legacy_types.rs
cberner: We can, and the first version of this PR did (the old `Legacy<(T,)>` used the bare tuple as its `SelfType`, the same way). The one thing it costs is redb-derive: the derive builds the struct from each field's `SelfType`, so with `SelfType = &str` a struct deriving `Value` cannot have an `Option<LegacyStr>` field, and its old table can only be opened through a hand-written `Value` impl that reproduces the derive's encoding. That was Codex's finding above, and the newtype is the fix for it. `Legacy<(T,)>` never came up against this, since it was only ever a table's top-level type.

So it is a trade: `&str` reads nicer for plain tables, the newtype keeps derived structs migratable with a one-line field change. I have left the newtype in. Say the word and I will switch it to `&str`, which also drops the wrapping and the derive test.

---

cberner: Switched to `&str`: `LegacyStr` is a unit struct again, a table of `Option<LegacyStr>` yields `Option<&str>`, and the newtype, its wrapping, and the derive test are gone. The doc notes in one line that a struct deriving `Value` cannot have it as a field.

---

src/legacy_types.rs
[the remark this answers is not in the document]
cberner: Reverted, on the maintainer's call: `SelfType` is `&str` again, so a table definition migrates by changing the type alone. A struct deriving `Value` cannot use `LegacyStr` as a field for now; redb-derive is not stable, and the doc says so.

---

src/legacy_types.rs
```diff
@@ -0,0 +1,94 @@
...
+/// }
+/// txn.delete_table(OLD)?;
+/// txn.commit()?;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// It wraps the string it reads, and outside an `Option` it is plain `&str`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct LegacyStr<'a>(pub &'a str);
```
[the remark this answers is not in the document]
cberner: Outdated by caee473: `LegacyStr` is a unit struct there, with `SelfType = &str`, so a table of `Option<LegacyStr>` yields `Option<&str>` and there is nothing to dereference.

---

cberner (commented):
Split it into two commits: one that adds the Legacy type, and another that adds the NICHE to &str

PR #1494, opened 2026-09-21
Give bool a niche, under the experimental-niches feature

cberner:
Under `experimental-niches`, `bool` declares `0xff` as its `NICHE`, which neither of its values encodes to, so `Option<bool>` is stored in one byte, with the niche standing for `None`, instead of a tag byte plus the value. As with `&str`, `Option<bool>` with the flag is not the same type as `Option<bool>` without it: a table written without the flag fails to open with `TableTypeMismatch` instead of being misread, and opens as `Option<Legacy<bool>>` for the copy-across migration. The file format is unchanged.

Tests: a unit test pins the `Option<bool>` width, encodings, round trip, and type name with and without the feature, and the redb 2.6 fixture in `tests/niche_tests.rs` gains `Option<bool>` value and key tables, rejected under the new definitions and read back through the legacy ones.

Follows [cberner/redb#1493](https://github.com/cberner/redb/pull/1493) (`String`), rebased onto it. The `char` PR is stacked on this one.


cberner:
lgtm. I'll merge it once the conflicts are resolved

PR #1497, opened 2026-09-21
Drop the lifetime parameter from KeyRange

cberner:
## Summary

`KeyRange<'a, K>` carried the lifetime of the borrowed key inside the range. It was inherited from the `RangeBounds` signature it replaced, where the method declared `'a` for its `KR: Borrow<K::SelfType<'a>>` bound, and it was hoisted into the trait so the impls could state the same bound.

Nothing observes that lifetime: `key_bounds()` encodes the bounds into owned `Vec<u8>` before the tree is touched, and the iterators borrow the table, not the range. So the trait is now `KeyRange<K>`, each impl quantifies the lifetime it needs for its `Borrow` bound itself (Rust allows an impl lifetime that appears only in where-clauses), and the range taking methods drop the `<'a>` they declared to feed it:

```rust
// before
fn range<'a>(&self, range: impl KeyRange<'a, K>) -> Result<Range<'_, K, V>>;
// after
fn range(&self, range: impl KeyRange<K>) -> Result<Range<'_, K, V>>;
```

The practical upside is for generic code over the range taking methods: `impl KeyRange<'_, K>` is rejected as an anonymous lifetime in `impl Trait`, so callers had to declare a lifetime just to pass a range through. Now they write `impl KeyRange<K>`.

The ranges the methods accept, and how they infer, are unchanged: `..`, `2..8`, `&2..&8`, `(Bound::Excluded(2), Bound::Unbounded)`, `&range`, and `key.as_str()..` borrowing from a local `String` are all still covered by `range_signature_inference`. Everything is behind `experimental-api-5`; implementors of `ReadableTable` / `ReadableMultimapTable` under the flag update their `range()` signature, as the delegating tables in the tests do. The existing 5.0.0 changelog entry already describes `KeyRange` without a lifetime, so no changelog change is needed.

## Validation

- `cargo fmt --check`, and `cargo clippy --all-targets -D warnings` with all features and with default features
- `cargo test` for the redb crate with all features and with default features, doc tests included
- `cargo doc` (the only warning is the pre-existing chrono `FixedOffset` table one)
- `cargo check` for `redb-bench` and for the fuzz crate under `--cfg fuzzing`

https://claude.ai/code/session_016q9Z3ZhrvK51x5TtuV8Cqo


CHANGELOG.md
```diff
@@ -102,6 +102,9 @@
 * Under the `experimental_cursor` feature flag, `CursorMut::insert_before()` and
   `CursorMut::insert_after()` now return `CursorError`, whose `UnorderedKey` variant replaces
   `StorageError::UnorderedKey`. `Error::UnorderedKey` is unchanged.
+* Under the `experimental-api-5` feature flag, `KeyRange` no longer has a lifetime parameter:
+  `impl KeyRange<'a, K>` in a signature becomes `impl KeyRange<K>`, and the range taking methods
+  drop the lifetime they declared for it. The ranges they accept are unchanged.
```
cberner: remove. API-5 changes should only be documented in the 5.0.0 release notes. They are experimental until then.

cberner: Removed in 1d62407; the PR is now the four source files only. The existing 5.0.0 entry already describes `KeyRange` without a lifetime, so nothing was added there.

---

src/key_range.rs
```diff
@@ -14,7 +14,7 @@ use core::ops::{Bound, Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, Ran
 /// range type rather than of the method, so `..` carries no type to infer and needs no annotation.
 ///
 /// This trait is sealed and cannot be implemented outside of redb.
-pub trait KeyRange<'a, K: Key + 'a>: Sealed {
+pub trait KeyRange<K: Key>: Sealed {
```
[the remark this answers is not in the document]
cberner: Stays as is: the maintainer's review on this PR removed the 4.4.0 note because `experimental-api-5` changes are only documented in the 5.0.0 release notes, and the existing 5.0.0 `KeyRange` entry already describes the trait as it will ship, without a lifetime parameter. Migrating from a stable 4.x release never sees the old form.

---

